PR #132, opened 2018-10-15
Switch to using a custom resource for AMI lookups

ebroder:
The CloudFormation template previously had a hard-coded set of AMIs to
use, but that gets stale quickly. Instead, use a custom resource backed
by a Lambda function to lookup the latest AMIs whenever the
CloudFormation template gets updated.

(This won't, e.g., upgrade instances if they're replaced between
CloudFormation runs, but it's still better than a fixed AMI.)

I haven't been able to test this in prod yet, because the stack is currently stuck due to a bug in an earlier version. I think it'll eventually reset after enough time.

### DIFF
--- a/cloudformation/jolly-roger.yaml
+++ b/cloudformation/jolly-roger.yaml
@@ -104,12 +104,6 @@ Parameters:
     Default: ''
     NoEcho: true
 
-Mappings:
-  UbuntuAmis:
-    us-east-1:
-      trusty: ami-5ac2cd4d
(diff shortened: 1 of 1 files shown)


ebroder:
OK, I think this works now (I just applied it in production, and it seemed to do the right thing).

There was a little wonkiness with the interactions between using async/await and using the `cfn-response` library. Hopefully the comment + code make clear what's going on, but I can add some elaboration if needed.

r? @zarvox

PR #160, opened 2018-11-03
Eliminate the private docker registry

ebroder:
Now that the Docker repo is public, it's no longer needed.


ebroder:
Rolled this change to prod to test; went fine

PR #163, opened 2018-11-19
Take easy package upgrades

ebroder:
None of these have breaking changes that affect us. All of the remaining outdated packages are due to the BS4 upgrade or the react-router 4 upgrade:

```
evan@bester:~/src/jolly-roger (evan/upgrades)$ meteor npm outdated
Package                 Current  Wanted  Latest  Location
bootstrap                 3.3.7   3.3.7   4.1.3  jolly-roger
react-router              3.2.1   3.2.1   4.3.1  jolly-roger
react-router-bootstrap   0.23.3  0.23.3  0.24.4  jolly-roger
```


package.json
```diff
@@ -12,36 +12,35 @@
     "lint": "eslint --ext js,jsx ."
   },
   "dependencies": {
-    "@babel/runtime": "^7.1.2",
-    "bcrypt": "^1.0.1",
+    "@babel/runtime": "^7.1.5",
+    "bcrypt": "^3.0.2",
```
ebroder: Main change in the version bump is switching from `$2a$` to `$2b$` hashes, but the new version can read old hashes.

package.json
```diff
@@ -12,36 +12,35 @@
...
     "bootstrap": "^3.3.6",
     "classnames": "^2.2.5",
     "element-resize-detector": "^1.1.14",
     "express": "^4.14.0",
-    "googleapis": "^34.0.0",
-    "history": "^2.0.0",
```
ebroder: We weren't actually using the history library directly; it's also pulled in by react-router, which is how we use it.

package.json
```diff
@@ -12,36 +12,35 @@
...
-    "libhoney": "^1.0.0-beta.2",
+    "googleapis": "^35.0.0",
+    "libhoney": "^1.2.1",
     "logfmt": "^1.2.0",
-    "marked": "^0.3.5",
+    "marked": "^0.5.1",
```
ebroder: I validated that our monkeypatches still work.

PR #166, opened 2018-12-31
Switch from emoji and Glyphicons to Font Awesome

ebroder:
The Font Awesome React component renders icons using SVG, which means
that we don't have to worry about font paths anymore. Also the icon set
is pretty great.

(This supersedes #165)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -267,6 +267,36 @@
       "resolved": "https://registry.npmjs.org/@emotion/utils/-/utils-0.8.2.tgz",
       "integrity": "sha512-rLu3wcBWH4P5q1CGoSSH/i9hrXs7SlbRLkoq9IGuoPYNGQvDJ3pt/wmOM+XgYjIDRMVIdkUWt0RsfzF50JfnCw=="
     },
+    "@fortawesome/fontawesome-common-types": {
+      "version": "0.2.12",
+      "resolved": "https://registry.npmjs.org/@fortawesome/fontawesome-common-types/-/fontawesome-common-types-0.2.12.tgz",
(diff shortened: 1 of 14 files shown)


package-lock.json
```diff
@@ -447,442 +530,36 @@
...
     "batch-processor": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/batch-processor/-/batch-processor-1.0.0.tgz",
       "integrity": "sha1-dclcMrdI4IUNEMKxaPa9vpiRrOg="
     },
     "bcrypt": {
-      "version": "3.0.2",
-      "resolved": "https://registry.npmjs.org/bcrypt/-/bcrypt-3.0.2.tgz",
-      "integrity": "sha512-kE1IaaRchCgdrmzQX/eBQKcsuL4jRHZ+O11sMvEUrI/HgFTQYAGvxlj9z7kb3zfFuwljQ5y8/NrbnXtgx5oJLg==",
+      "version": "1.0.3",
```
zarvox: I'm a bit weirded out by the bcrypt version downgrade here; any ideas what's going on?

ebroder: The downgrade happened in 2cbfd20b153fbc12e0373baab13ffa61d5c8643a; I think it just didn't get picked up by the lock file then for some reason.

ebroder:
I don't totally understand what happened with the lock file - I just ran `meteor npm i --save ...`. I think this may be previous dependency changes that didn't actually get captured in the lock file for some reason? (e.g. a handful of them seem to be libraries that bcrypt 1 depends on, but bcrypt 3 bundles into its own distribution)

It looks like this shrinks the browser bundle size by about 0.1M or so, probably because we only build in the icons we actually use.

PR #168, opened 2019-01-02
Replace "(puzzle)" link with puzzle piece icon

ebroder:
There was a suggestion from a few years ago that this might make the view a bit more compact. I'm a little worried about learning curve (I added a title tooltip to try and tame that a bit, not sure it'll actually be enough), but it does give us some pixels back.

Here's two screenshots to compare:

screenshot from 2019-01-01 22-03-51
screenshot from 2019-01-01 22-09-06

Thoughts?

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -46,7 +46,7 @@
   }
 
   .puzzle-link {
-    flex: 1 1 50px;
+    flex: 1 1 1em;
     display: inline-block;
     vertical-align: top;
     text-align: center;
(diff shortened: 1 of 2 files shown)


zarvox:
I weakly prefer the text over the icon, but I'm also the sort of person who prefers text labels for all links and buttons.

If y'all feel there's a reasonable benefit to having the icon, I won't object to merging this.  I agree the tooltip helps reduce the potential negative impact to learning curve.  We can always revert if we discover people are struggling to find the puzzle link?

grantaelliott:
Personally, I prefer this, but I know this is the sort of thing @zarvox and I philosophically disagree on.  I would argue that the icon approach frees up space and is less distracting than text.  I don't have strong feelings about it though.

PR #169, opened 2019-01-02
Remove obsolete TODOs

zarvox:
Some TODO lines are vestigial -- either we've implemented the thing named or it's no longer relevant to current designs.

### DIFF
--- a/imports/server/hooks/slack-hooks.js
+++ b/imports/server/hooks/slack-hooks.js
@@ -27,7 +27,6 @@ class SlackHooks extends Hookset {
   }
 
   onPuzzleNoLongerSolved(puzzleId) { // eslint-disable-line no-unused-vars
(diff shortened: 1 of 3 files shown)


imports/server/hooks/slack-hooks.js
```diff
@@ -27,7 +27,6 @@ class SlackHooks extends Hookset {
   }
 
   onPuzzleNoLongerSolved(puzzleId) { // eslint-disable-line no-unused-vars
-    // TODO: unarchive Slack channel
```
ebroder: lulz we were going to do channel-per-puzzle? oh to be young and naive

zarvox: yeah in hindsight I'm pretty glad we wound up doing our own chat thing instead of leaning too heavily on Slack

PR #171, opened 2019-01-02
Add gravatars to profile page

zarvox:
We already have email addresses from profiles, so to get profile pictures from gravatar we basically just have to add an MD5 implementation (~4kB).  While many people have Google accounts which they link to their jolly-roger accounts, we don't explicitly require them, and a Google integration might involve serverside pieces, whereas the gravatar integration could be entirely clientside, so I favored using gravatars over Google profile pictures for a first-pass profile image.  If we decide we want to, we could always build such a Google integration in as well for people who link Google accounts.

I don't actually feel strongly about if this is useful or not, but I found an old TODO and figured this was low-hanging fruit so I threw it together.

### DIFF
--- a/imports/client/components/Gravatar.jsx
+++ b/imports/client/components/Gravatar.jsx
@@ -1,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import md5 from 'blueimp-md5';
+
+function gravatarHash(emailAddress) {
+  return md5(emailAddress.trim().toLowerCase());
+}
+
+class Gravatar extends React.PureComponent {
+  static propTypes = {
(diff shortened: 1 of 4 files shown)


zarvox:
Yeah, the `OthersProfilePage` could probably benefit from us applying any design effort whatsoever to it some time.

I could believe that we might someday decide that we should use avatars in the chat UI, but there's considerations like "are gravatars used by enough of our team to be a good choice here?" and "do we think that the reduction in information density is worth the cost of imagery?" and "do we want to have to fetch profile primary email addresses for everyone in a hunt on every puzzle page" so I'm intentionally sidestepping all that for now.  But those are certainly possibilities if we like the idea of rethinking the chat UI.

PR #196, opened 2019-01-13
Fix breadcrumbs and guess queue announcement whitespace

zarvox:
I give up on trying to rely on mount order to determine breadcrumb sorting.  Let's just hardcode the sort orders everywhere we register breadcrumbs, since we don't even reuse page layouts at different places in the routes hierarchy.  (We might want to do that for the user profile route, but we don't right now, and it probably doesn't matter much.)

There was a breadcrumb that actually linked to the wrong page, but it never got used as a link so nobody ever noticed, until it was sorted out of order and then I clicked the link and it went the wrong place and I noticed.

Also, the guess-queue announcement used to not have breaking whitespace between a number of spans which looked bad.  I added spaces so they don't look ridiculous and can line-break.

### DIFF
--- a/imports/client/components/AllProfileListPage.jsx
+++ b/imports/client/components/AllProfileListPage.jsx
@@ -30,6 +30,7 @@ class AllProfileListPage extends React.Component {
         itemKey="users"
         to="/users"
         label="Users"
+        depth={0}
       >
         {body}
       </this.context.navAggregator.NavItem>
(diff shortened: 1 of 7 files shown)


zarvox:
Okay, since you took care of the breadcrumbs thing in #199 , I'll close this and make a new PR with the whitespace in guess queue notifications

PR #197, opened 2019-01-13
Redesign view filters / top UI on puzzle list page

aldeka:
Search bar is moved inline with the other view filters. View filters all
become toggle buttons instead of the nav pills + checkbox thing previously.

Add puzzle button is now on its own row.

On small screens, the announcements/guesses/hunters links move to the top.

Before, mobile:
screen shot 2019-01-12 at 22 28 49

After, mobile:
screen shot 2019-01-12 at 22 28 34

Before, desktop:
screen shot 2019-01-12 at 22 29 00

After, desktop:
screen shot 2019-01-12 at 22 29 15

### DIFF
--- a/client/stylesheets/puzzlelist.scss
+++ b/client/stylesheets/puzzlelist.scss
@@ -2,8 +2,44 @@
   display: flex;
   flex-direction: row;
   align-items: center;
-  justify-content: space-between;
   flex-wrap: wrap;
+
+  ul {
+    margin-bottom: 0;
+  }
+}
+
+.puzzle-view-controller {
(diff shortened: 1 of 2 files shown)


zarvox:
Thanks for the tweaks!  I think this is definitely better than what we had there before.

Mind fixing the ESLint failures?

aldeka:
ESLint errors fix'd!

grantaelliott:
@aldeka In Safari the list of links does not render to the right of the filter controls and instead wraps to below them.  At first glance, it appears the problem may be the text input expanding as much as possible.

PR #202, opened 2019-01-13
Make sure list of tag IDs is unique

### LINKED ISSUE #180: Work harder to prevent setting duplicate tags
It seems like there are a few cases where it's possible to end up with a puzzle that has multiple copies of the same tag (tag ID, not tag name) assigned to it. This makes the UI a bit wonky.

I think the main way to fix this is to make sure that we always use `$addToSet` and `$pull` when modifying tags. Today, that happens with `addTagToPuzzle` and `removeTagFromPuzzle`, but not (e.g.) when calling `updatePuzzle`. I think I'd propose 2 steps to fix this:

- Make `addTagToPuzzle` and `removeTagFromPuzzle` both take a list of tags. For `addTagsToPuzzle`, we'll need to use the [`$each` modifier](https://docs.mongodb.com/manual/reference/operator/update/addToSet/#each-modifier) to make it behave the way we want. For `removeTagFromPuzzle`, there's [an example](https://docs.mongodb.com/manual/reference/operator/update/pull/#remove-all-items-that-equal-a-specified-value) for removing multiple values at the same time. We probably want to rename the methods to `addTagsToPuzzle` and `removeTagsFromPuzzle`.
- Make `updatePuzzle` use `addTagsToPuzzle` and `removeTagsFromPuzzle` instead of setting the tags array directly.

ebroder:
This tries to ensure that the list of tags doesn't end up with
duplicates, preferably on the server side, but in the worst case
cleaning them up client-side.

Fixes #180.

I ended up doing this instead of consolidating around `addTag[s]ToPuzzle` and `removeTag[s]FromPuzzle` because you'd still have to compute what you were adding/removing and that seemed annoying.

### DIFF
--- a/imports/server/fixture.js
+++ b/imports/server/fixture.js
@@ -42,10 +42,12 @@ Meteor.methods({
       }, {
         $set: {
           hunt: huntId,
-          tags: puzzle.tags,
           title: puzzle.title,
           url: puzzle.url,
           answer: puzzle.answer,
+        },
+        $addToSet: {
+          tags: { $each: puzzle.tags },
         },
       });
     });
--- a/imports/lib/models/puzzles.js
+++ b/imports/lib/models/puzzles.js
@@ -1,9 +1,14 @@
+import { _ } from 'meteor/underscore';
 import { huntsMatchingCurrentUser } from '../../model-helpers.js';
 import Base from './base.js';
 import PuzzlesSchema from '../schemas/puzzles.js';
 import ActiveOperatorRole from '../active-operator-role.js';
 
-const Puzzles = new Base('puzzles');
+const Puzzles = new Base('puzzles', {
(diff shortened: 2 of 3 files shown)


imports/server/fixture.js
```diff
@@ -42,11 +42,13 @@ Meteor.methods({
       }, {
         $set: {
           hunt: huntId,
-          tags: puzzle.tags,
           title: puzzle.title,
           url: puzzle.url,
           answer: puzzle.answer,
         },
+        $addToSet: {
```
zarvox: Not that this is wrong, but FWIW I consider it a bug in the fixture if any of them have duplicate tag ids

ebroder: Yeah, I agree, but I felt like being exhaustive.

PR #205, opened 2019-01-14
Make it possible to link directly to guess submission page

### LINKED ISSUE #181: Make puzzle link in guess queue go directly to guess page
Today, when operators click on the puzzle page link from the guess queue or notifications, it takes them to the puzzle page, so they have to find the answer submission link and click that. Not terribly expensive, but unnecessary. Ideally, we could send operators directly to the answer submission page.

This is slightly challenging in that we may not know the URL structure for the hunt website in advance of hunt starting, so we probably don't want to hard-code any transformation logic here. However, we can probably have a hunt-level setting that says how to transform puzzle URLs.

Two ideas for how to do this:
- Provide a regex + a replacement string (basically the two arguments to [`String.prototype.replace`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace)). This gives us pretty good flexibility, since we can handle most URL structures. We'd also definitely need to think through how bad of a stored-XSS this introduces.
- Provide a simple templating language with variable substitution for things like `$HOST` and `$PATH`. This would have been sufficient for our site last year (e.g. we'd just set the template to `$HOST/submit/$PATH` or whatever), and I think it would work for past sites as well (maybe with the exception of Huntception), but if we need to replace a specific path component rather than add something at the beginning or the end, it might not be enough.

I think I lean towards the latter approach (especially because I don't want to think about whether user-provided input to `String.prototype.replace` is a stored-self-XSS).

ebroder:
I decided to use Mustache because it felt better than coming up with my own variable substitution code. I think that this does accomplish what we need, although it's not the most user-friendly thing imaginable. Maybe that's OK.

Minor note: It looks like the [API in question](https://developer.mozilla.org/en-US/docs/Web/API/URL) may not be supported in IE11, so we may want to grab a polyfill (a la https://www.npmjs.com/package/url-polyfill)

Fixes #181.

### DIFF
--- a/imports/client/components/GuessQueuePage.jsx
+++ b/imports/client/components/GuessQueuePage.jsx
@@ -8,10 +8,13 @@ import { withTracker } from 'meteor/react-meteor-data';
 import { withBreadcrumb } from '@ebroder/react-breadcrumbs-context';
 import subsCache from '../subsCache.js';
 import GuessesSchema from '../../lib/schemas/guess.js';
+import HuntsSchema from '../../lib/schemas/hunts.js';
 import PuzzlesSchema from '../../lib/schemas/puzzles.js';
 import Guesses from '../../lib/models/guess.js';
+import Hunts from '../../lib/models/hunts.js';
 import Profiles from '../../lib/models/profiles.js';
 import Puzzles from '../../lib/models/puzzles.js';
+import { guessURL } from '../../model-helpers.js';
 
 /* eslint-disable max-len */
 
@@ -47,6 +50,7 @@ const daysOfWeek = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Fri
 class GuessBlock extends React.Component {
   static propTypes = {
     canEdit: PropTypes.bool.isRequired,
+    hunt: PropTypes.shape(HuntsSchema.asReactPropTypes()).isRequired,
     guess: PropTypes.shape(GuessesSchema.asReactPropTypes()).isRequired,
     createdByDisplayName: PropTypes.string.isRequired,
     puzzle: PropTypes.shape(PuzzlesSchema.asReactPropTypes()).isRequired,
@@ -96,7 +100,7 @@ class GuessBlock extends React.Component {
           </div>
           <div>
             {'Puzzle: '}
-            <a href={this.props.puzzle.url} target="_blank" rel="noopener noreferrer">{this.props.puzzle.title}</a>
+            <a href={guessURL(this.props.hunt, this.props.puzzle)} target="_blank" rel="noopener noreferrer">{this.props.puzzle.title}</a>
             {' ('}
             <Link to={`/hunts/${this.props.puzzle.hunt}/puzzles/${this.props.puzzle._id}`}>discussion</Link>
             )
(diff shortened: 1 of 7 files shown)


imports/lib/models/puzzles.js
```diff
@@ -1,14 +1,33 @@
...
+const Puzzles = new class extends Base {
+  constructor() {
+    super('puzzles', {
+      transform(doc) {
+        return _.extend({}, doc, { tags: _.uniq(doc.tags) });
+      },
+    });
+  }
+
+  guessURL(hunt, puzzle) {
```
ebroder: I definitely don't love this as a place to stash this logic, but I'm not really sure where else to put it.

imports/client/components/GuessQueuePage.jsx
```diff
@@ -157,14 +162,18 @@ const crumb = withBreadcrumb(({ params }) => {
...
   const guessesHandle = subsCache.subscribe('mongo.guesses', {
     hunt: params.huntId,
   });
   const puzzlesHandle = subsCache.subscribe('mongo.puzzles', {
     hunt: params.huntId,
   });
   const displayNamesHandle = Profiles.subscribeDisplayNames(subsCache);
-  const ready = guessesHandle.ready() && puzzlesHandle.ready() && displayNamesHandle.ready();
+  const ready = huntHandle.ready() && guessesHandle.ready() && puzzlesHandle.ready() && displayNamesHandle.ready();
+  const hunt = ready ? Hunts.findOne({ _id: params.huntId }) : null;
```
zarvox: prefer `undefined` to `null`

imports/lib/models/puzzles.js
zarvox: I'd just as soon put it in its own file under `/imports/lib/` - maybe `guess-helpers.js` or something?  Or given that the only place it gets used is client-side, maybe even `/imports/client/guess-helpers.js`.

imports/lib/models/puzzles.js
```diff
@@ -1,14 +1,33 @@
...
 
-const Puzzles = new Base('puzzles', {
-  transform(doc) {
-    return _.extend({}, doc, { tags: _.uniq(doc.tags) });
-  },
-});
+// Disable escaping - we're passing this in as an attribute, not rendering it as
+// HTML, so we don't need to escape. Pretty not great that this is only
+// available as a global setting.
+Mustache.escape = s => s;
```
zarvox: Instead of quietly disabling escaping everywhere, can we just make the operator responsible for using triple-braces or the `&variable` syntax instead of double-braces to indicate that they don't want things to be escaped?  This seems like the sort of thing that will happen once per hunt and which you or I can be responsible for getting right

imports/client/components/HuntListPage.jsx
```diff
@@ -210,6 +218,45 @@ class HuntModalForm extends React.Component {
...
+              {' '}
+              <code>hostname</code>
+              {' '}
+              or
+              {' '}
+              <code>pathname</code>
+              {'. '}
+              As an example, setting this to
+              {' '}
+              <code>{'{{origin}}/submit{{pathname}}'}</code>
```
zarvox: based on my other comment about not disabling autoescaping, make this template have triple-`{{{` for the fields

imports/lib/models/puzzles.js
ebroder: Since we already have a `model-helpers.js`, I moved it there, although I think the idiomatic thing to do is probably to split each of those into their own file. /shrug problem for some other time.

ebroder:
Alright, PTAL?

PR #206, opened 2019-01-15
Switch to public breadcrumbs package

ebroder:


ebroder:
(PR got merged, so this has the `updateCrumb` fix)

PR #207, opened 2019-01-15
Introduce TypeScript

ebroder:
I'm generally in favor of trying to migrate code to TypeScript, and this sets up the infrastructure to do it.

I started with a simple file (and I basically stubbed the types on `guessURL` because I don't have schema types yet), but there's some really cool stuff going on in `huntsMatchingCurrentUser`:
- We can write a function that's generic to any model with a `hunt` property
- There's some neat type inference going on around line 32 - TypeScript can tell that each of those represents one of the 3 unioned types of a query term (check out the types on the Mongo APIs!)

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/945fd1a38258b551241eba913701b313c28b7884/types/meteor/mongo.d.ts#L61-L67

Larger conversions will be hard. In particular, I want to have a good answer for model schemas. In particular, I don't love that database query results are typed, but there's no actual load-time validation of schema - that seems nice and assumption-breaking. I want to play with wiring something like https://github.com/gcanti/io-ts into simple-schema. Not sure if that can be made to work.


ebroder:
Oh yeah. This should be reviewable commit-by-commit. First commit is basic find-replace

imports/client/components/AccountForm.tsx
```diff
@@ -1,20 +1,50 @@
...
+  errorMessage: string;
+  successMessage: string;
+  email: string;
+  password: string;
+  displayName: string;
+  phoneNumber: string;
+}
+
+class AccountForm extends React.Component<AccountFormProps, AccountFormState> {
   static propTypes = {
```
ebroder: I'm not sure what the right strategy is for proptypes with TS. Arguably if the whole codebase was converted we could just drop them, I guess?

imports/client/components/AccountForm.tsx
```diff
@@ -272,18 +302,18 @@ class AccountForm extends React.Component {
...
-              {format === 'login' || format === 'requestPwReset' ? emailInput : null}
-              {format === 'login' || format === 'enroll' || format === 'resetPwd' ? pwInput : null}
-              {format === 'login' ? pwResetOptionComponent : null}
-              {format === 'enroll' ? enrollmentFields : null}
+              {this.state.submitState === AccountFormSubmitState.FAILED ? <Alert bsStyle="danger">{this.state.errorMessage}</Alert> : null}
+              {this.state.submitState === AccountFormSubmitState.SUCCESS && this.state.successMessage ? <Alert bsStyle="success">{this.state.successMessage}</Alert> : null}
+              {format === AccountFormFormat.LOGIN || format === AccountFormFormat.REQUEST_PW_RESET ? emailInput : null}
+              {format === AccountFormFormat.LOGIN || format === AccountFormFormat.ENROLL || format === AccountFormFormat.RESET_PWD ? pwInput : null}
+              {format === AccountFormFormat.LOGIN ? pwResetOptionComponent : null}
+              {format === AccountFormFormat.ENROLL ? enrollmentFields : null}
```
ebroder: FWIW, I was a little surprised that it wasn't a type error to compare a string against an enum. I think it's allowed because the enum is assignable to the string (c.f. [the language spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#4.19.3))

server/main.js
```diff
@@ -1,35 +1,35 @@
...
+import '../imports/server/ansible';
+import '../imports/server/api-init';
+import '../imports/server/api_keys';
+import '../imports/server/chat';
+import '../imports/server/feature_flags';
+import '../imports/server/fixture';
+import '../imports/server/git_revision';
+import '../imports/server/guesses';
+import '../imports/server/hunts';
 import '../imports/server/migrations-run.js'; // runs migrations
```
zarvox: did you also want to change this line?

server/main.js
```diff
@@ -1,35 +1,35 @@
...
+import '../imports/server/puzzle';
+import '../imports/server/server-render';
+import '../imports/server/setup';
+import '../imports/server/slack-methods';
+import '../imports/server/subscribers';
+import '../imports/server/operator';
+import '../imports/server/users';
 
 // Load facades for convenient shell access.
 import Schemas from '../imports/lib/schemas/facade.js'; // eslint-disable-line no-unused-vars
```
zarvox: similarly these with comments?

server/main.js
ebroder: Yes - my find-replace was kind of coarse, I guess :joy:

imports/client/components/AccountForm.tsx
```diff
@@ -1,20 +1,50 @@
 import { Meteor } from 'meteor/meteor';
 import { Accounts } from 'meteor/accounts-base';
-import PropTypes from 'prop-types';
-import React from 'react';
-import Alert from 'react-bootstrap/lib/Alert';
+import * as PropTypes from 'prop-types';
+import * as React from 'react';
```
zarvox: Is this going to muck with our ability to do lazy-loading (which I previously recall you mentioning required using just default imports?)

ebroder: Hmm. I don't think so. This mostly comes into play because TypeScript draws a [distinction](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) between a "default export" and an "exports object" that I don't entirely understand. The docs assert that they're incompatible, which I don't totally see, but I also don't have an especially deep understanding of how the different modules systems work.

In any case, I think we have several options here:
- It's not clear that deferring imports of React itself is especially valuable, since we won't be able to _do_ anything until it loads anyway. (And I think we can avoid needing this `*` import for our own modules)
- I think it's possible we can tune this behavior with the `allowSyntheticDefaultImports` setting in `tsconfig.json`

PR #211, opened 2019-01-15
Puzzle list control CSS fixes.

grantaelliott:
Fixes two glitches in puzzle list controls CSS:
- On Safari, the text field always started on the next line instead of flowing alongside the buttons.
- The text field did not expand to fill available space when the window was narrower than 768px but still wide enough for the buttons and text field on the same line.

### DIFF
--- a/client/stylesheets/puzzlelist.scss
+++ b/client/stylesheets/puzzlelist.scss
@@ -24,6 +24,10 @@
       padding-left: 12px;
       margin-bottom: 8px;
     }
+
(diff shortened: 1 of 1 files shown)


grantaelliott:
@ebroder They're pretty straightforward.  Fixing the Safari-specific problem requires setting width to something non-zero.  I just left it at the old 250px figure, which is arbitrary and somewhat inelegant.  Open to suggestions if someone has a nicer way of fixing that.

Issue #192, opened 2019-01-12
Puzzle page on tablets


grantaelliott:
On iPad (and I suspect all sufficiently wide non-cursor interfaces), the following bugs affect puzzle pages:

- The embedded document cannot be edited.
- The embedded document does not update in real time.
- The right pane can be too wide, producing horizontal scrolling of the entire right pane, including the title section. (Open “Things to Bring to Hunt” on iPad to reproduce.) This may be exclusive to embedded spreadsheets.  If the content overflows the last column, the document can be scrolled horizontally, as it should, but the entire right pane can also be scrolled somewhat.  If the content does not overflow, the document cannot be scrolled - only the entire right pane.  This behavior also seems to be affected by the need for vertical scrolling.
- The embedded document cannot be scrolled vertically. (Open “Jolly Roger Walkthrough” on iPad to reproduce.) This may be exclusive to embedded text docs.

These do not affect phone views, because at reduced width the Google document is suppressed in favor of a button.

The first (and maybe the second) point may simply be a property of embedding Google documents now.  It may be that the only solution is to make the “Open worksheet” button available.

grantaelliott:
Closing this for now in light of #215.  The first two points do not appear to be addressable at this time, but a workaround has been added.  The latter two have been fixed.

grantaelliott:
Reopening this because it turns out documents still can't be scrolled vertically when in view mode on tablets (though sheets can).  This is way too minor in practice to deal with this morning, but it can probably be fixed.

grantaelliott:
These issues are largely resolved in iPadOS 13. While quirks such as locked vertical scrolling through view-mode docs will likely persist on other tablets and iPads running iOS 12 and earlier, I think it still makes sense to close this issue as
- I suspect relatively modern iPads account for the overwhelming majority of our tablet user base.
- On affected devices, the external link fallback solves the functional problem. (And on iPads, a split view between the Sheets app and the narrow JR layout reproduces the intended interface nicely.)
- The underlying issues are ultimately out of our control and any workarounds are likely to be fragile.

PR #215, opened 2019-01-17
Improved functionality on tablets.

grantaelliott:
Addresses #192.

On tablets and phones, Google loads a static html view in browsers, instead of the live editable version and I have found no way to force it do otherwise.  (On phones this makes no difference as there isn't space for the document anyway.)  As a workaround, this offers a link to open the document in the native app.  The link is not shown on devices with fine pointer control and is always shown on windows too small to load the document (as it always was).  On iPads, it's then possible to keep Safari open in narrow mode side-by-side with the native Sheets app, replicating the full JR experience.

(The static view unfortunately also renders slightly incorrectly, but this appears to simply be a bug in Google Sheets: The tab switcher is absolutely positioned as through the title bar were present, but that seems to be automatically hidden when embedded.)

Unrelatedly, iOS has unusual rules for sizing iframes.  This created scrolling problems on tablets for the static view that Google loads.  This is worked around with some odd CSS sizing and a scrolling=no attribute.

Finally, because the addition of this link made the puzzle metadata pane even more crowded, it's been reorganized somewhat and cleaned up with more consistent sizing and layout.  It is difficult to adjust the height of the react-select used to add tags, so the appearance is disrupted somewhat when that element is visible, and I would appreciate an assist with that if time permits.

Tested in narrow, wide (tablet), and wide (desktop) modes.  Open to suggestions, of course.

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -379,8 +379,14 @@ table.puzzle-list {
 }
 
 .puzzle-document > .gdrive-embed {
-  width: 100%;
-  height: 100%;
+  /* Workaround for unusual sizing behavior of iframes in iOS Safari */
+  /* Width and height need to be specified in px then adjusted by min and max */
+  width: 0px;
+  height: 0px;
+  min-width: 100%;
+  max-width: 100%;
+  min-height: 100%;
+  max-height: 100%;
   position: absolute;
   top: 0;
   right: 0;
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -140,10 +140,12 @@ table.puzzle-list {
 }
 
 .tag {
-  display: inline-block;
-  margin: 2px;
-  padding: 2px;
-  border-radius: 2px;
+  display: inline-flex;
+  align-items: center;
+  line-height: 24px;
+  margin: 2px 4px 2px 0;
+  padding: 0 6px;
+  border-radius: 4px;
   background-color: #ddd;
   color: #000;
   &.tag-administrivia {
@@ -179,7 +181,7 @@ table.puzzle-list {
   display: inline-block;
   vertical-align: middle;
   text-align: center;
-  margin: 0 0 0 4px;
+  margin: 0 0 0 6px;
 }
 
 /*Guessing and History Modal */
@@ -310,10 +312,6 @@ table.puzzle-list {
   justify-content: space-between;
 }
 
-.puzzle-metadata-title-set {
-  flex:1;
-}
-
 .puzzle-metadata-title {
   font-weight: bold;
 }
@@ -322,30 +320,25 @@ table.puzzle-list {
   cursor: pointer;
 }
 
-.puzzle-metadata-answer, .puzzle-metadata .tag {
+.puzzle-metadata-answer {
   display: flex;
   align-items: center;
   line-height: 24px;
   margin: 2px 0;
   padding: 0 6px;
   border-radius: 4px;
-}
-
-.puzzle-metadata-answer {
   background-color: #00ff00;
   color: #000000;
 }
 
 .puzzle-metadata .tag-list {
   display: flex;
+  flex-grow: 1;
   justify-content: flex-start;
   align-items: flex-start;
   align-content: flex-start;
   flex-wrap: wrap;
-}
-
-.puzzle-metadata .tag {
-  margin-right: 4px;
+  margin-left: 4px;
 }
 
 .puzzle-metadata-row .btn {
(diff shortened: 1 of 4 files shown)


imports/client/components/PuzzlePage.jsx
```diff
@@ -684,12 +684,10 @@ class PuzzlePageMetadata extends React.Component {
     const isAdministrivia = _.findWhere(tags, { name: 'administrivia' });
     const answerComponent = this.props.puzzle.answer ? (
       <span className="puzzle-metadata-answer">
-        Solved:
-        {' '}
         <span className="answer">{this.props.puzzle.answer}</span>
       </span>
     ) : null;
-    const hideViewCount = this.props.puzzle.answer || this.props.subcountersDisabled;
+    const hideViewCount = this.props.subcountersDisabled;
```
zarvox: I was a little surprised by this; generally we disable showing sub counters for solved puzzles at least on the puzzle list page, but I concede that it's weird to immediately hide them on the puzzle page, so I think I support this behavior change

imports/client/components/PuzzlePage.jsx
```diff
@@ -701,58 +699,54 @@ class PuzzlePageMetadata extends React.Component {
...
+                count={this.props.viewCount}
+                name={`puzzle:${this.props.puzzle._id}`}
               />
-            </div>
+            )}
           </div>
+          {this.props.puzzle.answer && answerComponent}
+        </div>
+        <div className={classnames('puzzle-metadata-row', this.props.isDesktop && 'puzzle-metadata-tag-editor-row')}>
+          <TagList
```
zarvox: Can I convince you to add the "Tags:" label back here?  Without it, on a puzzle that doesn't have any tags, it's not super clear what the tags are or what the + button is doing.

grantaelliott:
Behavioral changes to the metadata pane:
- Worksheet link is now styled as a link and is visible on tablets, even in wide mode.  (Still visible on narrow mode for all devices.)
- Puzzle link is moved from top right to bottom left, putting it in a row with the other actionable elements: the worksheet link and the guess button (still in the bottom right)
- Solved puzzle answer is moved over into the top right corner instead of inline with title text.
- Visitor count remains visible after solving.
- Removed unnecessary labels "Tags" and "Solved"
The new layout is three lines instead of two, but each occupies less vertical space because of cleanup in margins and padding and wraps later.  With no wrapping, the pane increased from 76px to 89px.

imports/client/components/PuzzlePage.jsx
grantaelliott: That's a good point.  If your concern is mostly the empty tag set case, I'd say we should just add a "(No Tags)" message.  But I'm also not particularly opposed to adding "Tags" back.  Your call.

client/stylesheets/puzzle.scss
```diff
@@ -297,66 +297,74 @@ table.puzzle-list {
...
-  text-align: right;
-  order: 2;
+.puzzle-metadata .view-count {
+  cursor: pointer;
 }
 
-.puzzle-metadata-tags {
-  display: inline-block;
-  height: 24px;
+.puzzle-metadata-answer, .puzzle-metadata .tag {
```
zarvox: Any reason you're overriding the styles for tags here?  I definitely agree with the adjustments to make the answer field larger and rounded, but applying the same to `.puzzle-metadata .tag` makes the tags look inconsistent with tags everywhere else in the app.

I'd suggest:

1. make this block not apply to `.puzzle-metadata .tag`
2. decide if you want to change the tag styles everywhere, and if you do, update `.tag` accordingly (~line 142) if so so we only have one set of styles for that type of widget

grantaelliott: I changed it in this context to make line heights and alignment consistent in the metadata pane.  I was reluctant to change tags globally, for fear of breaking something somewhere else.  It looks like tags only show up on the landing page and the related puzzles pane, so it probably wouldn't be too bad to make sure everything is consistent.

imports/client/components/PuzzlePage.jsx
zarvox: I think I like adding `Tags:` back.  It gives us a way to tell people what they should call the colored blobs.  Hunts and puzzles are hinted at by the nav breadcrumbs, guesses are hinted at by the button for showing previous guesses, but tags don't have much visibility for what to call that concept other than that label.

client/stylesheets/puzzle.scss
zarvox: Cool, happy to have rounded tags everywhere.  I think I originally made them the way they are to minimize pixels given to padding and to maximize information density, but I don't think we're quite that strapped for space.  I'd say try making this new style be the one we use everywhere then?

grantaelliott: Sounds good to me.

grantaelliott: Good call on this.  Turned out to be pretty easy to do and the consistency is undeniably nice.

PR #216, opened 2019-01-17
Make more user-provided content line-breakable

### LINKED ISSUE #185: Make sure user-provided text has `word-break: break-word`
We should double-check anywhere that we take user-provided output without a length limit and render it, and make sure those elements have `word-break: break-word` set in their CSS so that users can't break the layout with overflows.

This definitely includes chat messages and submitted answers. We may want to include operator-managed text like announcements and puzzle titles.

zarvox:
In particular, break display names and answer queue submissions in the
notifications and the guess queue page.

Fixes #185.

Before:

screenshot_20190117_143326

screenshot_20190117_143342

After:

screenshot_20190117_143202

screenshot_20190117_143249

### DIFF
--- a/client/stylesheets/guessqueue.scss
+++ b/client/stylesheets/guessqueue.scss
@@ -18,6 +18,10 @@
 
 .guess-info {
   flex: 1 0 50%;
+  overflow-x: hidden;
+  .breakable {
+    word-wrap: break-word;
+  }
 }
 
 .guess-button-group {
--- a/client/stylesheets/notifications.scss
+++ b/client/stylesheets/notifications.scss
@@ -59,12 +59,17 @@ ul.notifications {
     }
 
     .content {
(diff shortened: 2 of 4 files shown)


grantaelliott:
Good thinking.  Makes sense to me.

On a related note, should we take this opportunity to address the very limiting size of the read-only input in the guess queue?  While scrolling works, at a glance it frequently cuts off guesses.  Was a read-only input chosen just to make copy-paste easy?  (This could easily be a separate issue / PR of course.)

zarvox:
Ease of copy-paste was the primary reason we used a readonly input in the first pass of this view IIRC.  Earlier versions of jolly-roger didn't have clipboard integration the way the answer queue notifications do; it's possible that we should consider doing a similar thing here instead of the autoselecting input, but also most answers are less than 20 characters long and won't overflow the default-sized input anyway?

Issue #220, opened 2019-01-18
Consider moving navigation list on hunt page


grantaelliott:
The list of three navigation links on the hunt landing page is oddly located alongside the puzzle filter controls, despite not being related to filtering (or in some cases, puzzles).  It would make sense to promote these in a way that emphasizes their role as navigation, perhaps as a horizontal list at the top of page.

The "Add a puzzle" button for operators is also located here, but both makes sense in the context and would fit nicely in the layout if the list of links is moved elsewhere.

zarvox:
Fixed by #224

PR #224, opened 2019-01-18
Style list of links as navigation

grantaelliott:
Addresses #220

On the puzzle list page (hunt landing page) move list of links to the top and style to emphasize that they are navigation and not part of the filter controls.  Leave "Add puzzle" button inline with the filter controls.  Adjust flow of the remaining controls.

CSS is a little quick - I'm sure someone more skilled with flexbox could do this more elegantly.


zarvox:
Sure, this is better for operators, and not thaaaat much worse for non-operators.  :shipit: I say

PR #225, opened 2019-01-18
Fix CSS glitches involving breaks and wrapping

grantaelliott:
Three fixes:
- Viewer count and external links can no longer be broken by line wraps
- The metadata action row could be slightly too long on narrow phones (or iPad split view).  Adjusted the flex to wrap in the worst case and solve the problem in most cases by simply hiding the external link icons.  (I would rather hide the number of guesses so far, but that's actually a little trickier, so this is fine for now.)
- Match the chat textarea padding to the content above it.

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -319,6 +319,7 @@ table.puzzle-list {
 
 .puzzle-metadata .view-count {
   cursor: pointer;
+  white-space: nowrap;
 }
 
 .puzzle-metadata-answer {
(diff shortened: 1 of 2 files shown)


client/stylesheets/puzzle.scss
```diff
@@ -350,9 +351,21 @@ table.puzzle-list {
...
+}
+
 .puzzle-metadata .puzzle-metadata-external-link-button, .puzzle-metadata .gdrive-button {
   display: inline-block;
   font-weight: bold;
+  white-space: nowrap;
+  /* On very small touch screens or split view, desperately try to prevent line wrap */
+  /* Revisit this */
+  @media (max-width: 320px) and (pointer: coarse) {
+    svg {
```
zarvox: Is this just hiding the font icon?

grantaelliott: Yup.  It's neither the logic nor the behavior I'd ideally want, but it's something I could patch together in a few minutes this morning and it does solve the most common (by far) case.

PR #227, opened 2019-01-19
Guard more instances of tag name lookup by id against weird load order

zarvox:
I've seen both of these callsites crash the React renderer.  I'm hopeful
they're the last for this particular failure mode; I don't think I see other
places where we're doing the client-side join from tag id -> `Tag`.


zarvox:
rephrased without `continue` for the linter; best viewed with `?w=1`

PR #232, opened 2019-01-23
Add types for schemas

ebroder:
This introduces new helpers for converting a type represented in io-ts into a SimpleSchema. It also converts all of our schema definition files to TypeScript, using the conversion logic to produce both a TypeScript type and a schema for each model.

A few notes for reviewing:
* It probably makes sense to start by reviewing `typedSchemas.ts` first, although it's definitely a bit of a doozy.
* For a couple of the types, we can get meaningfully stronger typing in TS than we can get in SimpleSchema, so I split the types into a schema type and a true type.
* I'm not sure I like the structure of module exports


ebroder:
In particular, re: export structure, there are 3 possible things we could want to grab out of the schema module:

- The generated SimpleSchema
- The structured object from io-ts representing the type of the schema
- The actual type of the schema (which isn't currently exported - you have to use `t.TypeOf<typeof FooType>`)

Right now these modules only export the first two, and I think we want all 3. What do you call them? Should the types/overrides live in a separate file from the module that exports the schema?

ebroder:
@zarvox Any thoughts on what we should do here? I'm pretty sure I want to get something like this merged, but curious what you think.

imports/lib/schemas/typedSchemas.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  return new SimpleSchema(schema);
+};
+
+export const inheritSchema = function <
+  PT extends Record<string, any>,
+  PP extends Record<keyof PT, t.Mixed>,
+  CT extends Record<string, any>,
+  CP extends Record<keyof CT, t.Mixed>
+> (
```
zarvox: Consider writing some comments explaining what properties drive us toward this design, for posterity?

PR #238, opened 2019-03-13
Bump bootstrap version

ebroder:
Addresses [CVE-2019-8331](https://nvd.nist.gov/vuln/detail/CVE-2019-8331), which I think probably doesn't affect us, but this will make Github stop yelling at me.


ebroder:
Going to go ahead and merge without review because it's just a patch release bump.

PR #239, opened 2019-04-14
Start adding types to Models

ebroder:
This PR is probably a bit easier to review commit-by-commit.

First, I changed the schemas to export the actual raw Typescript type, in addition to the io-ts type value (which I renamed to a "codec" based on the terms they use in their docs; I tried to go through and use that term consistently for variable names and such). Then I started converting both the `Base` model class and some of the subclasses to Typescript.

I wasn't able to convert everything because there aren't any type declarations for the `nicolaslopezj:roles` package; I'll probably have to write them. Before I do that, I want to spend a little more effort on organizing the `.d.ts` files, since right now they're kind of just randomly dropped wherever.

Another followup step: right now we don't do any _load_-time schema validation. `io-ts` has methods for validating that an object conforms to one of its type values, and I'm pretty sure I want to wire it up (this will require us to be more rigorous about migrations, but I think we can get away with that)


imports/lib/schemas/typedSchemas.ts
```diff
@@ -241,14 +241,14 @@ export const inheritSchema = function <
...
-  childSchemaType: t.InterfaceType<CP, CT>,
+  parentSchemaCodec: t.InterfaceType<PP, PT>,
+  childSchemaCodec: t.InterfaceType<CP, CT>,
   parentOverrides: Overrides<PT>,
   childOverrides: Overrides<CT>
-): [t.InterfaceType<PP & CP>, Overrides<PT & CP>] {
-  const inheritedType = t.type({
-    ...parentSchemaType.props,
-    ...childSchemaType.props,
+): [t.TypeC<PP & CP>, Overrides<PT & CP>] {
```
ebroder: This is the one actually meaningful change in this commit. `t.InterfaceType` captures the map of keys to codecs in its first type parameter, but the actual type is the second type parameter (which defaults to `any`, not very useful). `t.TypeC` is just a `t.InterfaceType` with better defaults: https://github.com/gcanti/io-ts/blob/6f75b1026921b87485332be5dabd3e51397af185/src/index.ts#L767-L768

Without this, taking `t.TypeOf<typeof AnnouncementCodec>` or whatever just returns `any`

ebroder:
@zarvox Any thoughts on this? Don't want to bypass your opinion, but if you are OK with the general direction I can just keep charging forward.

PR #242, opened 2019-05-27
Attempt to fix the Docker build

ebroder:


ebroder:
It doesn't seem like this is totally necessary, but we saw some flakiness from the Ubuntu keyserver, so going to go ahead and land this, since it should be more robust

PR #243, opened 2019-05-28
Add ambient type declarations for nicolaslopezj:roles

ebroder:
As part of adding support, I moved all of the type declarations I already added to a top-level `types/` directory, organized by the module that the types come from (which can be different than the types being added, since a lot of these are monkeypatches)


types/nicolaslopezj:roles/roles.d.ts
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable no-restricted-globals */
+declare module 'meteor/nicolaslopezj:roles' {
+  type RuleFuncThis = { userId: string };
+  type RuleFunc = (this: RuleFuncThis, ...args: any[]) => boolean;
+  type Rule = RuleFunc | boolean
+
+  // eslint-disable-next-line import/prefer-default-export
+  export module Roles {
+    const Role: RoleStatic;
```
ebroder: This is a little shenanigans. There is a `class Foo` declaration style that would have worked better - something like:

```typescript
class Role {
  constructor(name: string);
  readonly name: string;
  function allow(action: string, func: Rule): void;
}
```

but eslint can't seem to handle a class declaration with methods with no body - it throws a parse error. It's fine with this nonsense, though (which I copied from the `meteor/mongo` declarations).

PR #244, opened 2019-06-01
Finish converting almost all models to TypeScript

ebroder:
The only model not converted at this point is the Lock model. That's
blocked on converting Ansible, and there's also some trickiness with the
autovalue for createdAt.


imports/lib/models/settings.js
```diff
@@ -1,18 +0,0 @@
...
-Settings.attachSchema(SettingsSchema);
-
-function queryModifier(q) {
-  // Only allow admins to pull down Settings.
-  if (Roles.userHasRole(this.userId, 'admin')) {
-    return q;
-  }
-
-  // Make the query evaluate to nothing
-  return { $and: [false, q] };
```
ebroder: Technically this was invalid, since the list elements for `$and` are supposed to be query hashes, not booleans. I think if this query had actually run, it would have resulted in an error (which is maybe close enough to desired)

PR #246, opened 2019-06-09
Exclude the _id field from the SimpleSchema

ebroder:
We need to have _id as part of the type so that we can access the field,
but we don't want to declare it as part of the schema. The validation
logic in collection2 doesn't handle the fact that _id is auto-populated
well, so it causes spurious validation errors.


ebroder:
(FTR, I ran into this specifically with the upsert in the `setFeatureFlag` server-side method. I think it might only be an issue in cases where we are doing upserts and not inserts or updates-by-_id, but not positive)

PR #250, opened 2019-06-30
Convert most React components to Typescript

ebroder:


imports/client/components/AllProfileListPage.tsx
```diff
@@ -1,17 +1,25 @@
...
+}
+
+class AllProfileListPage extends React.Component<AllProfileListPageProps> {
   static propTypes = {
     ready: PropTypes.bool.isRequired,
-    profiles: PropTypes.arrayOf(PropTypes.shape(ProfileSchema.asReactPropTypes())).isRequired,
+    profiles: PropTypes.arrayOf(
+      PropTypes.shape(
+        ProfileSchema.asReactPropTypes<ProfileType>()
+      ).isRequired as React.Validator<ProfileType>
```
ebroder: There's some wonkiness because of how the ambient type declarations for `PropTypes` works. It inspects to see if a field is optional (by seeing if its type includes either `null` or `undefined`), and if so, makes the property optional (in the `foo?: string`) sense. That doesn't match our schema type declarations, which accept `undefined` but "require" the property to be present.

PR #258, opened 2020-01-04
Monkeypatch type of findOne to return undefined

ebroder:
findOne isn't guaranteed to find a result; it might return undefined.
However, the DefinitelyTyped declaration for findOne doesn't contemplate
this, which results in a couple of potential bugs.

### DIFF
--- a/imports/client/components/AccountForm.tsx
+++ b/imports/client/components/AccountForm.tsx
@@ -83,11 +83,11 @@ class AccountForm extends React.Component<AccountFormProps, AccountFormState> {
     this.setState({
       submitState: AccountFormSubmitState.SUBMITTING,
     });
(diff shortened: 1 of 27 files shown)


ebroder:
Oh actually, it looks like this got fixed in DefinitelyTyped upstream, let me rework to be a bit less janky

ebroder:
OK. This is a cleaner approach, although it did blow up the size of the change a bit. More details in the new commit message:

> Update @types/meteor
> 
> This makes some of the types more accurate, especially `Meteor.user()`, `Meteor.userId()`, and `Collection.findOne()`, all of which can return null, but previously weren't typed to do so.
> 
> For user and userId, in most cases we know whether or not we're logged in based on the current active component, so this can mostly be handled by the type assertion operator. Similarly, there are several cases where we don't need to handle null if we assume our database is consistent. However, some of these changes did catch actual areas where we didn't have strong enough null guards.

PR #260, opened 2020-01-05
Upgrade to TypeScript 3.7

ebroder:


imports/client/components/GuessQueuePage.tsx
```diff
@@ -67,10 +67,10 @@ interface GuessBlockProps {
 class GuessBlock extends React.Component<GuessBlockProps> {
   static propTypes = {
     canEdit: PropTypes.bool.isRequired,
-    hunt: PropTypes.shape(HuntsSchema.asReactPropTypes()).isRequired as React.Validator<HuntType>,
-    guess: PropTypes.shape(GuessesSchema.asReactPropTypes()).isRequired as React.Validator<GuessType>,
+    hunt: PropTypes.shape(HuntsSchema.asReactPropTypes<HuntType>()).isRequired as React.Validator<HuntType>,
```
ebroder: I originally typed `asReactPropTypes` as taking `T = never` to try and force specifying a template parameter. That didn't work on 3.4, but it does work on 3.7. (Sadly, we still need the bonus cast at the end because TypeScript and PropTypes don't 100% agree on optional properties vs. properties that include `| undefined`)

PR #261, opened 2020-01-05
Lighten announcement and guess subscriptions

ebroder:
Historically, we've made overly broad subscriptions to have information
ambiently available to us. But that results in fairly heavy data fetches
on initial page load (e.g. we fetch all puzzles just in case there's a
pending guess).

By pushing some of the logic to the server, and having a more specific
subscription, we can significantly tighten the scope of data that we
have to fetch.

### DIFF
--- a/imports/server/announcements.ts
+++ b/imports/server/announcements.ts
@@ -4,6 +4,40 @@ import { Roles } from 'meteor/nicolaslopezj:roles';
 import Ansible from '../ansible';
 import Announcements from '../lib/models/announcements';
 import PendingAnnouncements from '../lib/models/pending_announcements';
+import Profiles from '../lib/models/profiles';
+
+Meteor.publish('announcements.pending', function () {
+  if (!this.userId) {
+    throw new Meteor.Error(401, 'Unauthenticated');
+  }
+
+  const pas = PendingAnnouncements.find({ user: this.userId });
+  const announcementIds = [...new Set(pas.map(pa => pa.announcement))];
+  const announcements = Announcements.find({ _id: { $in: announcementIds } });
+
+  const creatorIds = [...new Set(announcements.map(a => a.createdBy))];
+  const creators = Profiles.find({ _id: { $in: creatorIds } }, { fields: { displayName: 1 } });
+
+  return [pas, announcements, creators];
+});
+Meteor.publish('announcements.all', function (huntId: string) {
+  check(huntId, String);
+  const u = Meteor.users.findOne(this.userId);
+  if (!u) {
+    throw new Meteor.Error(401, 'Unauthenticated');
+  }
+
+  if (!u.hunts.includes(huntId)) {
+    throw new Meteor.Error(403, 'Not a member of that hunt');
+  }
+
+  const announcements = Announcements.find({ hunt: huntId });
+
+  const creatorIds = [...new Set(announcements.map(a => a.createdBy))];
+  const creators = Profiles.find({ _id: { $in: creatorIds } }, { fields: { displayName: 1 } });
+
+  return [announcements, creators];
+});
 
 Meteor.methods({
   postAnnouncement(huntId, message) {
--- a/imports/server/guesses.ts
+++ b/imports/server/guesses.ts
@@ -1,12 +1,38 @@
-import { Meteor } from 'meteor/meteor';
+import { Meteor, Subscription } from 'meteor/meteor';
 import { check } from 'meteor/check';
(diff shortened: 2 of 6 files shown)


ebroder:
@zarvox Curious what your thoughts are on this. I think I like it but could use a second opinion

imports/server/announcements.ts
```diff
@@ -4,6 +4,40 @@ import { Roles } from 'meteor/nicolaslopezj:roles';
...
+  if (!u.hunts.includes(huntId)) {
+    throw new Meteor.Error(403, 'Not a member of that hunt');
+  }
+
+  const announcements = Announcements.find({ hunt: huntId });
+
+  const creatorIds = [...new Set(announcements.map(a => a.createdBy))];
+  const creators = Profiles.find({ _id: { $in: creatorIds } }, { fields: { displayName: 1 } });
+
+  return [announcements, creators];
```
zarvox: I believe this is not *quite* going to do what you want, since publishes are not rerun on result changes.

Namely, what will happen here is that if a new `Announcement` is created by someone who did not wind up in `creatorIds` when the subscription was created, that `Announcement` will get published to the client, but the `Profile` for the creator will not, which will wind up with missing data client-side when you go to render it.

imports/server/announcements.ts
```diff
@@ -4,6 +4,40 @@ import { Roles } from 'meteor/nicolaslopezj:roles';
...
+  }
+
+  const pas = PendingAnnouncements.find({ user: this.userId });
+  const announcementIds = [...new Set(pas.map(pa => pa.announcement))];
+  const announcements = Announcements.find({ _id: { $in: announcementIds } });
+
+  const creatorIds = [...new Set(announcements.map(a => a.createdBy))];
+  const creators = Profiles.find({ _id: { $in: creatorIds } }, { fields: { displayName: 1 } });
+
+  return [pas, announcements, creators];
```
zarvox: This has the same issue as `announcements.all` below -- the `PendingAnnoucements` collection available client-side will get updated reactively later, but `Annoucements` and `Profiles` will not.

imports/server/guesses.ts
```diff
@@ -1,13 +1,39 @@
...
+    const guesses = Guesses.find({ ...filter, hunt: { $in: u.hunts } });
+    const huntIds = [...new Set(guesses.map(g => g.hunt))];
+    const puzzleIds = [...new Set(guesses.map(g => g.puzzle))];
+    const guesserIds = [...new Set(guesses.map(g => g.createdBy))];
+
+    const hunts = Hunts.find({ _id: { $in: huntIds } });
+    const puzzles = Puzzles.find({ _id: { $in: puzzleIds } });
+    const guessers = Profiles.find({ _id: { $in: guesserIds } }, { fields: { displayName: 1 } });
+
+    return [guesses, hunts, puzzles, guessers];
```
zarvox: Again, this will only cover hunts/puzzles/guessers that existed at the time the subscription was initially created, and will not behave correctly if any of that happens later on.  I expect you to wind up with missing data unless you're implicitly pulling in the same data from some other subscription elsewhere.

ebroder:
Hmm. Plausibly we could switch to [publish-composite](https://atmospherejs.com/reywood/publish-composite), which seems designed for exactly this use case?

zarvox:
I could believe that something like `publish-composite` would work for us here, though honestly I'm a little wary of the CPU/memory overhead to carefully track the reactive publish sets for every client.  We'll save bandwidth, but we might spike CPU, which could have the net effect of making the overall system less performant under load.

Probably worth trying and measuring though?

ebroder:
I think we could plausibly break down strategy a bit based on the subscriptions. `guesses.pending` and `guesses.all` should only be loaded by operators (in theory we could explicitly restrict to that, but today it's implied by client code structure). Since the subs should be relatively infrequent, the increased CPU usage seems like it shouldn't be too expensive.

Announcements are a bit more ambiguous since everyone subscribes to them and can potentially view the announcement list page.

Meteor does do some coalescing of subscriptions, right? How does that work? If we got rid of `announcements.pending` entirely and made everyone subscribe to `announcements.all` (so that everyone with the same hunt membership would have the same query) + `PendingAnnouncements` separately, would that work better?

ebroder:
From reading the [docs](https://guide.meteor.com/data-loading.html#lifecycle), it kind of looks like if your publish function directly returns a cursor, there's some coalescing that can happen, but if it doesn't return a cursor, then maybe there's no coalescing, which would be a pretty big argument against `publish-composite` for heavily used subscriptions

PR #263, opened 2020-01-09
Switch meteor method arguments to unknown

ebroder:
Now that `check` is annotated as an assertion function, we can get
sufficient type information just from the check calls. By annotating the
argument types to unknown, we can more accurately reflect that we don't
yet know the types.

### DIFF
--- a/imports/server/hunts.ts
+++ b/imports/server/hunts.ts
@@ -37,10 +37,10 @@ const existingJoinEmail = (user: Meteor.User | null, hunt: HuntType, joinerName:
 };
 
 Meteor.methods({
-  addToHunt(huntId: string, email: string) {
+  addToHunt(huntId: unknown, email: unknown) {
     check(huntId, String);
     check(email, String);
-    if (!this.userId) throw new Meteor.Error(401, 'Unauthorized');
+    check(this.userId, String);
 
     const hunt = Hunts.findOne(huntId);
     if (!hunt) {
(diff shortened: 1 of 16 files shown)


imports/server/hunts.ts
```diff
@@ -37,10 +37,10 @@ const existingJoinEmail = (user: Meteor.User | null, hunt: HuntType, joinerName:
 };
 
 Meteor.methods({
-  addToHunt(huntId: string, email: string) {
+  addToHunt(huntId: unknown, email: unknown) {
     check(huntId, String);
     check(email, String);
-    if (!this.userId) throw new Meteor.Error(401, 'Unauthorized');
```
zarvox: You sure that `this.userId` is always either a full user ID or undefined, and not empty string?  Just want to be sure we're not dropping a necessary authentication check.

imports/server/operator.ts
```diff
@@ -6,14 +6,14 @@ import Ansible from '../ansible';
 Meteor.methods({
   // Temporarily de-op yourself
   stopOperating() {
-    if (!this.userId) throw new Meteor.Error(401, 'Unauthorized');
```
zarvox: ditto

imports/server/hunts.ts
ebroder: Yeah, the [Meteor docs](https://docs.meteor.com/api/methods.html#DDPCommon-MethodInvocation-userId) document the semantics of this field. I added a handful of these checks [when I converted to TypeScript](https://github.com/deathandmayhem/jolly-roger/commit/f1e73440b93c187b240d726dfb38eedda6bec192#diff-f4c92cab507de9189ac164cf7c595bb6R43) so that the typechecker would see the value as non-null later on.

PR #265, opened 2020-01-12
Configure import/order eslint plugin

### LINKED ISSUE #209: Setup import/order eslint rule
As we've moved more towards using modules and imports, the import block at the top of lots of files has gotten kind of excessive. I've noticed a little cognitive overhead to deciding where to put imports, and that seems like something that tooling can just solve for me.

I think we should just enable the `import/order` eslint rule, and let it run its auto-fixed on the codebase. Preferably at a time when nobody else has anything in flight.

(Sadly, this isn't a perfect answer, since import/order doesn't seem to support ordering within groups, but worth the baby steps)

ebroder:
(And use the auto-fix to bring us into compliance.)

Fixes #209

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -64,6 +64,19 @@
     // We don't use imports with extensions
     "import/extensions": ["error", "never"],
 
+    // Imports should be grouped, and sorted within the group. Meteor imports
+    // are their own group, after built-ins and before other external libraries
+    "import/order": ["error", {
+      "alphabetize": {"order": "asc"},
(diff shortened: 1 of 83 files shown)


ebroder:
For what it's worth, it seems like this didn't touch the imports in `client/main.ts`, `server/main.ts`, or `imports/server/migrations/all.ts`. I don't entirely understand why, but I assume it understands that imports that don't import anything are side-effectful and therefore order-sensitive

PR #268, opened 2020-01-13
Minor improvements to puzzle metadata style

grantaelliott:
- Correct an overly restrictive media query from #225, used to minimize content for very narrow windows.
- Allow answers to break. Previously, long answers could force the title to wrap poorly. While not a perfect solution, simply allowing the answer to break (combined with the existing flex behavior) results in a reasonable layout. This change primarily affects mobile and puzzles with exceptionally long answers.

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -331,6 +331,7 @@ table.puzzle-list {
   border-radius: 4px;
   background-color: #00ff00;
   color: #000000;
+  word-break: break-all;
 }
 
 .puzzle-metadata .tag-list {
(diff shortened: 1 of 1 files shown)


zarvox:
Mind posting example screenshots demonstrating the impact of this change?

grantaelliott:
Current (iPhone 8 example):

New, showing answer breaking:


This is not a perfect solution. In some cases, a title may not use the full width allocated to it, due to the positioning of word breaks, so the answer is forced to break even though a different flex allocation would have fit the answer on one line without affecting the title. iPhone SE example:

I don't see a way to better handle this case without introducing an onResize callback, which I think is too large a hit to readability/maintainability.

Issue #234, opened 2019-01-29
Improve "related puzzles" pane


ebroder:
The related puzzles pane on the puzzle page doesn't seem to have proven very useful. I think we intended for it to surface potential answer constraints on individual puzzles, but people don't really do that. At most, it's maybe useful for metas.

We should likely hide or remove the related puzzles pane, although maybe we should leave it around on puzzles tagged with `is:meta`. We could either replace it with more chat, or possibly display a list of useful solving tools or similar.

grantaelliott:
I've found the related puzzles pane useful in the past and would prefer not to remove it entirely, though I agree that extended chat history would be more useful on most puzzles. I would support simply collapsing the pane by default on puzzles that are tagged neither is:meta nor meta-for:*

grantaelliott:
Tentatively closed, by #271. Worth revisiting if we come up with new information worth presenting in some cases.

PR #271, opened 2020-01-15
Suggestion: Automatically collapse related puzzles if unlikely to be useful

grantaelliott:
Per Issue #234, related puzzles are most useful for meta puzzles. For other puzzles, the information should be available, but the space may be better devoted to chat.

This automatically collapses related puzzles for puzzles without meta-related tags. When navigating between puzzles, related puzzles will never be collapsed automatically (so related puzzles stays open when navigating from a meta to a puzzle), but may be expanded (so related puzzles is opened automatically when navigating from a puzzle to a meta, though currently there are no links to other puzzles other than in the related pane, so it was almost certainly already open).

As an example, clicking Spotted Tower from the hunt landing page will cause the related puzzles pane to open, but clicking Topsy Turvy from the same page will leave it collapsed. However, navigating to Topsy Turvy from Spotted Tower will result in the pane staying open.

This behavior is just a suggestion and is quite open to discussion. (We may even decide to close issue #234 with no changes.)

### DIFF
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -1070,6 +1070,7 @@ interface PuzzlePageState {
   showRelated: boolean;
   isDesktop: boolean;
   isStackable: boolean;
+  defaultsAppliedForPuzzle: string;
 }
 
 class PuzzlePage extends React.Component<PuzzlePageProps, PuzzlePageState> {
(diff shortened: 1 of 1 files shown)


grantaelliott:
Any ideas for what else we could put there? From #234 it sounds like you have a sense of how most players are using the tools in practice.

@zarvox Do you have an opinion? It feels strange to unceremoniously merge something like this, that affects visibility.

zarvox:
No objection.  I happen to personally like the related puzzles pane, mostly for navigation purposes, but I recognize that I'm probably an outlier and I probably also have some "I made this!" IKEA effect influencing my perception of it.  There's like a few flows that I think I probably use this for that would go away:

* what metas does this round puzzle feed into?
* what fraction of the puzzles in this puzzle's meta are solved vs. unsolved?
* what are some plausible constraints on this puzzle's solution, based on what we know about other puzzles that feed into it?
* if I'm sick of this puzzle but still want to work on something contributing to the same meta, what else can I work on?

In practice, with a little scrolling, these questions are easily answered by looking at the puzzle list page instead.  And it might make more sense to have side-by-side windows of puzzle list page and a puzzle page when doing that sort of workflow than to carve out a tiny sliver of screen space for the related puzzles section.

And for people who just want to solve puzzles, the related puzzle section is frequently more of a distraction that limits chat history visibility than a boon.

I'm at peace with this.  I'd consider also leaving the related puzzles page in for `administrivia`; I don't think those tend to get much chat usage.

quentinmit:
I'll chime in as someone who also really likes the related puzzles pane. The PR title says "collapse"; as long as the disclosure triangle is still there that seems fine to me.

grantaelliott:
Per @zarvox 'administrivia' tag is now recognized as likely to need related content. Also changed the text at the top of that section to "Related," since we do use the puzzle collection for more than just puzzles. (The word 'Puzzle' appears in a few other places, but changing it here seemed low impact and at least vaguely related to this PR.)

PR #272, opened 2020-01-16
Rewrite SplitPanePlus as fully controlled

grantaelliott:
Rewrite SplitPanePlus to be fully controlled, with associtated rewiring of PuzzlePage. SplitPanePlus now has somewhat different (but hopefully more intuitive) behaviors around the size prop than the underlying SplitPane. Automatic classes are now assigned to individual collapsed panes and the resizer instead of to the parent SplitPane. Min and max sizes are now respected during resize.

Fixes several bugs (The first three could occur whenever Related Puzzles was collapsed):
- Clicking and immediately releasing the resizer when a secondary pane is collapsed no longer clears the collapse state.
- Dragging a collapsed pane and releasing it within the auto-collapse zone no longer clears the collapse state.
- The primary pane no longer overfills the parent container when the secondary pane is collapsed. (This resulted in the bottom of the chat input being cropped whenever Related Puzzles was collapsed.)
- The resizing cursor persists during the drag. (The fix for this was quite inelegant and I'm very open to suggestions.)

Also improves default sizing for various panes on PuzzlePage and reduces the time for the resizer fade effect.

### DIFF
--- a/imports/client/components/SplitPanePlus.tsx
+++ b/imports/client/components/SplitPanePlus.tsx
@@ -73,6 +73,12 @@ import SplitPane, { Props as SplitPaneProps } from 'react-split-pane';
       Styles applied to the resizer
 */
 
+/* Why is this necessary? These should already be in SplitPaneProps */
+interface FullSplitPaneProps extends SplitPaneProps {
+  pane1ClassName?: string;
+  pane2ClassName?: string;
+}
+
 interface SplitPanePlusProps {
   children: React.ReactNode[],
   size: number,
@@ -136,11 +142,9 @@ class SplitPanePlus extends React.Component<SplitPanePlusProps, SplitPanePlusSta
     // This is an inelegant way of preventing the browser from going into selection mode and
     // overriding the cursor. It also has to be accompanied by additional cursor styles for the
     // pane below (in puzzle.scss).
-    if (this.ref && this.ref.current) {
-      const resizerEl = this.ref.current.querySelector(':scope > .SplitPane > .Resizer');
-      if (resizerEl) {
-        resizerEl.addEventListener('mousedown', (ev) => { ev.preventDefault(); });
-      }
+    const resizerEl = this.resizerNode();
+    if (resizerEl) {
+      resizerEl.addEventListener('mousedown', (ev) => { ev.preventDefault(); });
     }
   }
 
@@ -163,8 +167,8 @@ class SplitPanePlus extends React.Component<SplitPanePlusProps, SplitPanePlusSta
       if (this.props.onPaneChanged) {
         this.props.onPaneChanged(newSize, newCollapsed, 'resize');
       }
-      this.forceUpdate();
     }
+    this.forceUpdate();
   }
 
   onChange = (size: number) => {
@@ -214,29 +218,20 @@ class SplitPanePlus extends React.Component<SplitPanePlusProps, SplitPanePlusSta
   }
 
   primaryPaneNode(): HTMLElement | null {
-    const root = this.splitPaneNode();
-    const className = `Pane${this.props.primary === 'first' ? 1 : 2}`;
-    return root ? _.find(root.childNodes, (n) => (
-      n instanceof Element &&
-      n.classList.contains(className)
-    )) as HTMLElement : null;
+    return this.findChildByClass(`Pane${this.props.primary === 'first' ? 1 : 2}`);
   }
 
   secondaryPaneNode(): HTMLElement | null {
-    const root = this.splitPaneNode();
-    const className = `Pane${this.props.primary === 'first' ? 2 : 1}`;
-    return root ? _.find(root.childNodes, (n) => (
-      n instanceof Element &&
-      n.classList.contains(className)
-    )) as HTMLElement : null;
+    return this.findChildByClass(`Pane${this.props.primary === 'first' ? 2 : 1}`);
   }
 
   resizerNode(): HTMLElement | null {
+    return this.findChildByClass('Resizer');
+  }
+
+  findChildByClass(className: string): HTMLElement | null {
     const root = this.splitPaneNode();
-    return root ? _.find(root.childNodes, (n) => (
-      n instanceof Element &&
-      n.classList.contains('Resizer')
-    )) as HTMLElement : null;
+    return root && _.find(root.children, (n) => n.classList.contains(className)) as HTMLElement;
   }
 
   measure(elem: HTMLElement | null): number {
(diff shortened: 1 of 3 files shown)


imports/client/components/SplitPanePlus.tsx
```diff
@@ -1,92 +1,130 @@
...
+    paneClassName?, pane1ClassName?, pane2ClassName?: string
+      Class names applied to both, the first, and the second pane respectively, in addition to
+      'Pane', 'Pane1', and 'Pane2' as well as 'Vertical' or 'Horizontal' (depending on split)
+    paneStyle?, pane1Style?, pane2Style?: React.CSSProperties
+      Styles applied to both, the first, and the second pane respectively
+    resizerClassName?: string
+      Class names applied to the resizer, in addition to 'Resizer' as well as 'Vertical' or
+      'Horizontal' (depending on split)
+    resizerStyle?: React.CSSProperties
+      Styles applied to the resizer
```
ebroder: Nit, and we can follow up later, but it'd be nice to move these comments onto the prop interface definition, since the types already live there

imports/client/components/SplitPanePlus.tsx
```diff
@@ -95,9 +133,14 @@ class SplitPanePlus extends React.Component<SplitPanePlusProps, SplitPanePlusSta
...
-    if (this.props.collapsed === 0) {
-      this.recordSize(this.measure(this.primaryPaneNode()!));
+    // This is an inelegant way of preventing the browser from going into selection mode and
+    // overriding the cursor. It also has to be accompanied by additional cursor styles for the
+    // pane below (in puzzle.scss).
+    if (this.ref && this.ref.current) {
+      const resizerEl = this.ref.current.querySelector(':scope > .SplitPane > .Resizer');
+      if (resizerEl) {
+        resizerEl.addEventListener('mousedown', (ev) => { ev.preventDefault(); });
+      }
```
ebroder: FYI for future reference, we're now on a version of Typescript that has safe navigation, so you could just write this as `const resizerEl = this.ref?.current?.querySelector(...)`

imports/client/components/SplitPanePlus.tsx
```diff
@@ -111,168 +154,175 @@ class SplitPanePlus extends React.Component<SplitPanePlusProps, SplitPanePlusSta
...
     }
     return this.ref.current.firstChild;
   }
 
-  primaryPaneNode() {
+  primaryPaneNode(): HTMLElement | null {
     const root = this.splitPaneNode();
     const className = `Pane${this.props.primary === 'first' ? 1 : 2}`;
     return root ? _.find(root.childNodes, (n) => (
       n instanceof Element &&
```
ebroder: I think these may also need to change to `HTMLElement` for the cast below to be safe

imports/client/components/SplitPanePlus.tsx
```diff
@@ -111,168 +154,175 @@ class SplitPanePlus extends React.Component<SplitPanePlusProps, SplitPanePlusSta
...
   }
 
   render() {
-    const paneProps: SplitPaneProps = _.extend({}, this.props);
-
-    const defaultPaneStyles = { overflow: 'auto' };
-    const maxRangeStyles = (this.props.split === 'vertical' ?
-      { maxWidth: '100%' } :
-      { maxHeight: '100%' }
+    const paneProps: { [key: string]: any; } = _.pick(
```
ebroder: Can you declare this as, like, `Partial<SplitPaneProps>`?

imports/client/components/SplitPanePlus.tsx
```diff
@@ -111,168 +154,175 @@ class SplitPanePlus extends React.Component<SplitPanePlusProps, SplitPanePlusSta
...
-      defaultPaneStyles,
-      this.props.primary === 'first' ? maxRangeStyles : {},
-      this.props.pane1Style);
-    paneProps.pane2Style = _.extend({},
-      defaultPaneStyles,
-      this.props.primary === 'first' ? {} : maxRangeStyles,
-      this.props.pane2Style);
-
+    const defaultPaneStyle: React.CSSProperties = { overflow: 'auto' };
+    paneProps.paneStyle = _.extend(defaultPaneStyle, this.props.paneStyle);
```
ebroder: FWIW, I think this will mutate `defaultPaneStyle`, which is probably not what you want. `{ ...defaultPaneStyle, ...this.props.paneStyle }` should work to create a new object.

(I also kiiiind of want to move away from using underscore because the type definitions are a bit weaker than a lot of the modern ES6+ native equivalents - e.g. `_.extend` is defined as taking `any` and returning `any`, which isn't super helpful)

ebroder: Oh I...missed that you were just defining this above. Still could probably use object-spread syntax.

imports/client/components/SplitPanePlus.tsx
grantaelliott: That's a good point, but I'm afraid it's getting pushed to next week...

imports/client/components/SplitPanePlus.tsx
grantaelliott: I tried to do this (in splitPaneNode because the above code disappears with a trivial refactor) and got some very strange behavior. I'll leave it for now and revisit save navigation next week.

imports/client/components/SplitPanePlus.tsx
grantaelliott: Not entirely sure what you mean by this. Maybe you misread the ternary? If so, I can't blame you because this is terrible. Cleaned it up dramatically.

imports/client/components/SplitPanePlus.tsx
grantaelliott: I thought so too and wasted a lot of time last night trying to make it work, but for some reason pane1ClassName and pane2ClassName (which clearly work, because I'm using them) throw errors as not existing. For now, I just added them to SplitPaneProps, but that's obviously not great. Any ideas?

imports/client/components/SplitPanePlus.tsx
grantaelliott: Just sloppy on my part. Good catch. Here it was functionally fine, but a few lines below pane1Style could leak into pane2Style.

I had no idea you could do that with object spread. Neat.

imports/client/components/SplitPanePlus.tsx
ebroder: It looks like they're just missing from the type declarations in upstream: https://github.com/tomkp/react-split-pane/blob/master/index.d.ts#L7-L28. Probably worth filing a bug over at some point

grantaelliott: Well I feel like a fool.  I glanced at https://github.com/tomkp/react-split-pane/blob/master/src/SplitPane.js#L352-L378, decided it was in there, and never again considered the possibility this was just a bug on that side. Until it's fixed, as we ok with the patched in approach I used?

ebroder: Yeah, react-split-pane isn't actually written in Typescript, so they have the type definitions separate from the code...which means they can be wrong

PR #273, opened 2020-01-16
Suppress 4 eslint warnings for react/no-danger

grantaelliott:
Per conversation with @ebroder we aren't likely to change these four uses of dangerouslySetInnerHTML, so the warnings being generated aren't drawing attention to anything useful. I suggest adding the eslint-disable as an acknowledgment of the issue and moving on without the warnings.


ebroder:
FWIW, I'd be in favor of just disabling this rule entirely in eslintrc, rather than having to keep adding more overrides if we end up using dangerouslySetInnerHTML elsewhere - IMO the rule doesn't really add any signalling beyond the name of the attribute itself

PR #274, opened 2020-01-16
Relocate viewers in puzzle metadata

grantaelliott:
Despite doing the same thing (opening a modal), the viewers and guess button have different presentations. The former is an unstyled link with a popover and the latter is a button. Since the action isn't navigation, standardize on button. Also move the viewers button to the bottom row, as it doesn't make a lot of sense inline with the title and contributes to needless line breaks there.

Change the text for both buttons to be more concise (and invoke a verb). The viewers button now reads "See \<n\> hunter(s)" while the guess button reads either "Guess (\<n\> so far)" or "See \<n\> guess(es)" depending on the solve status.

Change the labels for document links to reflect Google's naming convention (which conveniently shortens them) and use different icons for sheets and docs.

With those changes, everything fits on one line on most modern phones (using iPhone 8 as a guide). Below 375px (e.g. iPhone SE at 320px), the link labels are hidden. The link labels are hidden with css media queries, which is sufficient as long as the desktop view isn't allowed to reduce the width below 375px.

### DIFF
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -231,9 +231,12 @@ class ViewersModal extends React.Component<ViewersModalProps, ViewersModalState>
   }
 }
 
-interface ViewCountDisplayProps {
+interface ViewCountDisplayParams {
   count: number;
   name: string;
+}
+
+interface ViewCountDisplayProps extends ViewCountDisplayParams {
   subfetchesDisabled: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


grantaelliott:
Desktop (Tablet view, not shown, is similar to iPhone 8 view with more space):

iPhone 8:

iPhone SE:

imports/client/components/PuzzlePage.tsx
```diff
@@ -250,29 +253,21 @@ class ViewCountDisplay extends React.Component<ViewCountDisplayProps> {
   };
 
   render() {
-    const text = `(${this.props.count} viewing)`;
+    const text = `See ${this.props.count} ${this.props.count === 1 ? 'hunter' : 'hunters'}`;
```
ebroder: I find this text to be a bit more confusing relative to what was there before - "hunters" doesn't clearly scan to me as "people currently looking at this puzzle". "See {count} viewers" maybe?

grantaelliott: My thinking was to make it more consistent with the "hunter" language used elsewhere, but you're right. I'll change it to "viewer(s)".

Issue #235, opened 2019-01-29
Guess notification should link to JR puzzle page


ebroder:
When a new guess comes in, operators frequently want to stop by the puzzle page to see where the guess came from, do some sanity checking, etc. But the notification links directly to the guess submission page on the Hunt website. We should link to the Jolly Roger puzzle page instead, maybe with an additional link (using the font awesome puzzle piece?) to the actual submission page.

grantaelliott:
It looks to me like the Jolly Roger page is the more important one, but I'm not a frequent operator. It might be smart (and easy) just to replace the existing inline link with a link to the JR page.

Regarding the puzzle-piece icon, we currently use that for external link to the puzzle on the puzzle list page (but then we use the external link icon for the same thing on the JR puzzle page). We should be careful about introducing contradictory uses. (Looking at it now, I think I'd support using the external link icon for external puzzle links everywhere.)

grantaelliott:
Closed by #275

PR #275, opened 2020-01-17
Suggestion: Add JR link to guess notification

grantaelliott:
This is a suggested implementation of #235, adding a JR link to guess notifications. Adds separate buttons, each with a tooltip, for opening the JR page or the puzzle.

notification

Per conversation on #235, this adopts the convention of using the external link icon for puzzle links and further uses the puzzle piece icon for JR links. If we adopt this convention, we should replace the icons generated by puzzle listings with the external link icon (probably a good idea anyway).


grantaelliott:
Per conversation with @ebroder, switched to skull-crossbones for JR links and puzzle-piece for puzzle links.

PR #276, opened 2020-01-17
Suggestion: Use external link icon for external puzzle links

grantaelliott:
Use external link icon for puzzle links in listings, consistent with the use of this icon on the puzzle page. This is consistent with #275 and, if that is adopted, this probably should be as well.

I really have no strong opinions about what icon gets used for what, but it'd be nice to be consistent. Puzzle-piece for JR link and external-link for puzzle link seems a reasonable convention, but I'm very open to suggestions. (I vaguely recall supporting puzzle-piece for links years ago...)


grantaelliott:
Per slack conversation with @ebroder, we're going with skull-crossbones for JR links and puzzle-piece for puzzle links, so Puzzle is correct as is. Closing.

PR #278, opened 2020-01-17
Get rid of desiredLayout hack

ebroder:
If we're a bit more clever with our CSS (and accept one very minor UI change - namely that the disconnect alert is always at a fixed non-scrolling position), then we can have fixed markup for the app framework and only need custom markup and styling for the actual puzzle page.

This gets rid of the fairly awkward hack where the App component was going digging through the resolved routes under it (which will make the react-router v5 upgrade easier).

### DIFF
--- a/client/stylesheets/app.scss
+++ b/client/stylesheets/app.scss
@@ -1,21 +1,11 @@
-.app-content-fullscreen {
-  position: fixed;
-  top: 50px;
-  bottom: 0px;
-  left: 0px;
-  right: 0px;
-}
-
-.connection-status-fullscreen {
+.connection-status {
   position: fixed;
   top: 50px;
   left: 0px;
   right: 0px;
-  z-index: 1;
-}
-
-.app-content-scrollable {
-  padding-top: 70px;
+  // This z-index is chosen to be higher than any z-index used by Bootstrap
+  // (which are all in the 1000-1100 range)
+  z-index: 10000;
 }
 
 /* Prevent mobile safari zoom */
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -1309,7 +1309,5 @@ const tracker = withTracker(({ params }: PuzzlePageParams) => {
 });
 
 const PuzzlePageContainer = tracker(crumb(PuzzlePage));
-// Mark this page as needing fixed, fullscreen layout.
-(PuzzlePageContainer as any).desiredLayout = 'fullscreen';
 
 export default PuzzlePageContainer;
--- a/client/stylesheets/index.css
+++ b/client/stylesheets/index.css
@@ -1,3 +1,10 @@
+.jolly-roger {
+    /* This value is taken from the Bootstrap documentation to support
+    navbar-fixed-top:
+    https://getbootstrap.com/docs/3.4/components/#navbar-fixed-top */
+    padding-top: 70px;
+}
+
 .jolly-roger .navbar-brand {
     padding: 0;
     float: none;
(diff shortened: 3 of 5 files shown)


ebroder:
FYI @grantaelliott this seems like the sort of thing you would have feelings about. @zarvox and I tested this IRL and our belief is that all of the existing layout should continue to work. The only UI change that I'm aware of is making the status bar always be `position: fixed`, instead of before, where it was inline. This seems like a fine change - for long pages like the puzzle list page, you actually probably want feedback that you're offline even if you're scrolled down.

client/stylesheets/app.scss
```diff
@@ -1,20 +1,12 @@
-.app-content-fullscreen {
+.connection-status {
   position: fixed;
   top: 50px;
-  bottom: 0px;
   left: 0px;
   right: 0px;
+  z-index: 10000;
```
zarvox: This was specifically chosen to be greater than bootstrap's modals' z-indices, which are in the low 1000s.  (Might be worth a comment to that effect.)

client/stylesheets/app.scss
```diff
@@ -1,20 +1,12 @@
...
-.connection-status-fullscreen {
-  position: fixed;
-  top: 50px;
-  left: 0px;
-  right: 0px;
-  z-index: 1;
-}
-
-.app-content-scrollable {
+body {
```
zarvox: Putting this on `<body>` in the middle of `app.scss` still strikes me as odd, but I'll live

but maybe this belongs in a different file?

client/stylesheets/app.scss
```diff
@@ -1,20 +1,12 @@
...
   padding-top: 70px;
```
zarvox: Did we arrive at this number via any particular set of principles?  Is this fundamentally 50px navbar + 20px content padding?

ebroder: This is taken straight from the bootstrap docs: https://getbootstrap.com/docs/3.4/components/#navbar-fixed-top

client/stylesheets/app.scss
ebroder: It's possible we could put this on the `.jolly-roger` div, but I'd want to re-test it to make sure it works consistently

imports/client/components/PuzzlePage.tsx
```diff
@@ -1177,24 +1177,26 @@ class PuzzlePage extends React.Component<PuzzlePageProps, PuzzlePageState> {
...
-              <PuzzlePageContent
-                puzzle={activePuzzle}
-                allTags={this.props.allTags}
-                guesses={this.props.allGuesses}
-                displayNames={this.props.displayNames}
-                document={this.props.document}
-                isDesktop={this.state.isDesktop}
-              />
-            </SplitPanePlus>
+            <div className="puzzle-page-inner">
```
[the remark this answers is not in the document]
grantaelliott: If there's a reason for the inner div to stay, in the name of consistency we could add it to the narrow view as well, move the .narrow to the new inner div, and update the .puzzle-page.narrow style to .puzzle-page-inner.narrow. But it isn't necessary.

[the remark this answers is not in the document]
ebroder: I thought we needed a container with `position: fixed` so that we could then put the `position: absolute` container inside of it, but that doesn't actually seem to be necessary, so we can drop it.

client/stylesheets/app.scss
ebroder: Added a comment

client/stylesheets/app.scss
ebroder: This works on the `.jolly-roger` div, so that's a bit cleaner

client/stylesheets/app.scss
ebroder: (Added a comment)

PR #282, opened 2020-01-18
Switch to rendering emoji with twemoji

ebroder:


ebroder:
I thinks this introduces a bunch of XSS, so I think I need to try harder

PR #303, opened 2020-01-25
Add search bar to guess queue page

zarvox:
This makes it possible to filter down from a hunt's worth of submissions to
particular puzzles or guesses.  Of course ctrl-f worked before, but this makes
it easy to scan over guesses.

### DIFF
--- a/imports/client/components/GuessQueuePage.tsx
+++ b/imports/client/components/GuessQueuePage.tsx
@@ -3,9 +3,14 @@ import { Roles } from 'meteor/nicolaslopezj:roles';
 import { withTracker } from 'meteor/react-meteor-data';
 import { _ } from 'meteor/underscore';
 import classnames from 'classnames';
+import { Location } from 'history';
 import React from 'react';
+import Button from 'react-bootstrap/lib/Button';
+import FormControl from 'react-bootstrap/lib/FormControl';
+import FormGroup from 'react-bootstrap/lib/FormGroup';
+import InputGroup from 'react-bootstrap/lib/InputGroup';
 import { withBreadcrumb } from 'react-breadcrumbs-context';
-import { Link } from 'react-router';
+import { Link, browserHistory } from 'react-router';
 import Guesses from '../../lib/models/guess';
 import Hunts from '../../lib/models/hunts';
 import Profiles from '../../lib/models/profiles';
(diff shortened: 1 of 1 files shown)


imports/client/components/GuessQueuePage.tsx
```diff
@@ -138,17 +145,91 @@ type GuessQueuePageProps = GuessQueuePageParams & {
...
+
+  compileMatcher = (searchKeys: string[]): (g: GuessType) => boolean => {
+    return (guess) => {
+      const puzzle = this.props.puzzles[guess.puzzle];
+      const guessText = guess.guess.toLowerCase();
+
+      // Try to find a reason to be interested in this guess.
+      for (let i = 0; i < searchKeys.length; i++) {
+        const key = searchKeys[i].toLowerCase();
+        if (guessText.indexOf(key) !== -1 || puzzle.title.toLowerCase().indexOf(key) !== -1) {
```
grantaelliott: Is matching arbitrary substrings in the title the user-expected behavior? I think I’d expect it to be constrained to leading (or potentially leading or trailing) substrings of each word in the title, especially given the implicit disjunction. (For guesses, which frequently lack word breaks, arbitrary substrings make sense of course.)

You should do whatever you prefer, as it makes no difference and I don’t think we have a consistent style, but in my opinion using searchKeys.some() would be more consistent with the rest of the code (and more concise) than this for-loop.

zarvox: Grant, you have a knack for asking the right questions to make me think about a problem more clearly. :)

Thoughts:

* I do explicitly want the filter to be type-ahead style filtering, where adding additional constraints will never make the list of results get longer.  I like having the live-update on each keystroke, and making sure that the results have this property of stability is something that I value.
* On reflection, I'm not sure having disjunctive search results makes the most sense here, or anywhere?  The original implementation for filter-search in the puzzle list page is disjunctive, and I think that might have been because the control flow was easier for December 2015 Drew to hack together quickly and get working, and not because I thought about it real hard and concluded this was the best approach.
* Arbitrary substring matching is probably not quite how people usually use search.  I think I started out with the attitude that it was better to make it easier to get a match than not in case there's punctuation or something, but that might also have been that it was fewer function calls to just `indexOf()` the whole title rather than doing word-splitting on each puzzle title.  That said, I can't remember a time that the permissive mid-word-matching has been useful rather than surprising, so I'm willing to try more restricted matching here
* Prefix-matching is compatible with the constraints around incrementally shrinking the result set; postfix-matching is not.

I think I'll switch this to doing conjunction on multiple keys, prefix search on title words, and keep looking for any substring in answer phrases.

PR #313, opened 2020-09-30
Take mostly easy updates

ebroder:
This brings basically everything up to date except for packages related to
- marked (API changes to how HTML sanitization works)
- bootstrap (v3 -> v4)
- react-router (v3 -> v5 at this point)


imports/client/components/PuzzlePage.tsx
```diff
@@ -425,7 +424,6 @@ const chatInputStyles = {
     padding: '9px 4px',
     borderWidth: '1px 0 0 0',
     resize: 'none' as 'none',
-    maxHeight: '200px',
```
zarvox, line 428: Any reason for dropping this style?  I think it was intended to avoid the chat pane from growing so large it takes over the entire left panel.

ebroder: It's no longer allowed in the new version of `react-textarea-autosize`, and specifying it is an error: https://github.com/Andarist/react-textarea-autosize/blob/master/CHANGELOG.md#major-changes.

I think that previously it was intercepting that field and using it internally, rather than actually passing it through to the underlying style spec.

PR #314, opened 2020-09-30
Upgrade to Bootstrap 4

ebroder:
This takes out one of the two major outstanding upgrades (the other being react-router) by getting us onto the latest branches of BS and react-bootstrap.

My process was mostly
- Upgrade the npm packages
- Fix imports that no longer existed (including `react-bootstrap` moving component level imports from `react-bootstrap/lib/` to `react-bootstrap/`, plus some classes that got removed)
- Fix all of the type errors
- Navigate through all the pages and see what looked broken

BS4 is slightly stylistically different than BS3 - the text defaults to slightly bigger, and the colors on buttons and such are a bit bolder. I left that as-is for now, although if there are strong objections I can try to shrink things back down. Overall I didn't try to get pixel-perfect with the old version, but I did try to make sure that nothing looked egregiously bad and that everything loaded and seemed visible.

I'll wait to merge until I can get an active review, since this was so invasive.


imports/client/components/Puzzle.tsx
```diff
@@ -64,7 +64,7 @@ class Puzzle extends React.PureComponent<PuzzleProps, PuzzleState> {
   editButton = () => {
     if (this.props.canUpdate) {
       return (
-        <Button onClick={this.showEditModal} bsStyle="default" bsSize="xs" title="Edit puzzle...">
+        <Button onClick={this.showEditModal} variant="secondary" size="sm" title="Edit puzzle...">
```
zarvox, line 67: this went from bsSize=xs to size=sm -- intentional?

imports/client/components/PuzzleModalForm.tsx
```diff
@@ -1,8 +1,10 @@
...
-import Alert from 'react-bootstrap/lib/Alert';
-import ControlLabel from 'react-bootstrap/lib/ControlLabel';
-import FormControl from 'react-bootstrap/lib/FormControl';
-import FormGroup from 'react-bootstrap/lib/FormGroup';
+import Alert from 'react-bootstrap/Alert';
+import FormControl, { FormControlProps } from 'react-bootstrap/FormControl';
+import FormGroup from 'react-bootstrap/FormGroup';
+import FormLabel from 'react-bootstrap/FormLabel';
+import Row from 'react-bootstrap/Row';
+import Col from 'react-bootstrap/esm/Col';
```
zarvox, line 7: what's this `esm/` doing here? https://react-bootstrap.github.io/layout/grid/#col-props suggests it should just be `'react-bootstrap/Col'`

package.json
```diff
@@ -27,6 +27,7 @@
     "history": "^4.10.1",
     "io-ts": "^2.2.10",
     "io-ts-types": "^0.5.11",
+    "jquery": "^3.5.1",
```
zarvox, line 30: I guess `bootstrap` pulls in `jquery`, even though we're only really using the CSS?  Presumably if we don't import it anywhere we're not actually paying for it in our bundle, right?

package.json
```diff
@@ -36,9 +37,10 @@
     "monocle-ts": "^2.3.3",
     "mustache": "^4.0.1",
     "newtype-ts": "^0.3.4",
+    "popper": "^1.0.1",
```
zarvox, line 40: What pulls this in?  also is this intended to be https://www.npmjs.com/package/popper.js or https://www.npmjs.com/package/popper I'm confused

imports/client/components/Puzzle.tsx
ebroder: Yeah, buttons [no longer have an "xs" size](https://getbootstrap.com/docs/4.5/migration/#:~:text=Dropped%20the%20.btn-xs%20class%20entirely).

imports/client/components/PuzzleModalForm.tsx
ebroder: Bah, oversight. VS Code was very determined to pull its imports from there instead of top-level. Pushed a fix.

package.json
ebroder: Yeah, I think that's right. `bootstrap` has both `jquery` and `popper` (below) as peer dependencies, meaning that we're supposed to include them ourselves. However, I think that `react-bootstrap` side-steps all of the built-in JS in bootstrap (certainly I don't load it anywhere) so neither should end up in the bundle.

zarvox (approved):
LGTM

We can chase the look-and-feel stuff separately; lets land this

grantaelliott:
I'm more than happy to work on the look-and-feel aspects, but I won't have time to do so until this weekend.

PR #315, opened 2020-10-06
Upgrade to react-router v5

zarvox:
react-router v5 is a completely incompatible change, so this necessitated some rewriting.

`params` have moved to live under the `match` prop, and their expected contents can be a type parameter to `RouteComponentProps`.

The previous approach we had used for our `Authenticator` is no longer viable under react-router v4+'s model, since it is no longer possible for a parent component to determine if any of its children match before it renders.  This makes it possible to do async loading of code for routes, but means that we can no longer have a single authenticator component in the tree.  Instead, the approach I took was to add wrapper components that produced the same sorts of redirects as needed for auth-required routes (or any of the three unauthenticated routes if you happen to be logged in already).


imports/client/components/Routes.tsx
```diff
@@ -1,69 +1,45 @@
...
+              <AuthenticatedRoute path="/users/:userId" component={ProfilePage} />
+              <AuthenticatedRoute path="/users" component={AllProfileListPage} />
+              <AuthenticatedRoute path="/setup" component={SetupPage} />
+
+              {/* Unauthenticated routes - if user already logged in, get redirected to /hunts */}
+              <UnauthenticatedRoute path="/login" component={LoginForm} />
+              <UnauthenticatedRoute path="/reset-password/:token" component={PasswordResetForm} />
+              <UnauthenticatedRoute path="/enroll/:token" component={EnrollForm} />
+
+              {/* Routes available to both authenticated and unauthenticated users */}
```
ebroder, line 40: I feel like maybe we should just kill this comment since it doesn't seem like we're ever going to have any of these

imports/client/components/GuessQueuePage.tsx
```diff
@@ -130,35 +130,36 @@ class GuessBlock extends React.Component<GuessBlockProps> {
...
   };
 
   setSearchString = (val: string) => {
-    const newQuery = val ? { q: val } : { q: undefined };
-    browserHistory.replace({
+    const newSearch = val ? `q=${val}` : '';
+    this.props.history.replace({
       pathname: this.props.location.pathname,
-      query: { ...this.props.location.query, ...newQuery },
+      search: newSearch,
```
ebroder, line 162: (Same here re: `URLSearchParams`)

ebroder (approved):
OK, I think this overall makes sense to me. We _really_ should add some basic tests for auth behavior though. Maybe I'll try to rig up a basic test suite later.

imports/client/components/App.tsx
```diff
@@ -87,7 +93,7 @@ class AppNavbar extends React.Component<AppNavbarProps> {
   }
 }
 
-const AppNavbarContainer = withTracker(() => {
+const AppNavbarContainer = withTracker((_props: RouteComponentProps) => {
```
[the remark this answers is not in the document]
zarvox: Because we do make use of it in `AppNavbar` (so that we can access the history object on logout, so that we redirect before other components have a chance to observe `Meteor.userId()` returning null).

And if I don't annotate the type here, then typescript complains (because the props passed to `AppNavbar` might not include all of the `RouteComponentProps`, which we need them to).

imports/client/components/PuzzleListPage.tsx
```diff
@@ -75,15 +76,15 @@ class PuzzleListView extends React.Component<PuzzleListViewProps, PuzzleListView
...
   };
 
   setSearchString = (val: string) => {
-    const newQuery = val ? { q: val } : { q: undefined };
-    browserHistory.replace({
+    const newSearch = val ? `q=${val}` : '';
+    this.props.history.replace({
       pathname: this.props.location.pathname,
-      query: _.extend(this.props.location.query, newQuery),
+      search: newSearch,
```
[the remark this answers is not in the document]
zarvox: Yeah, this is definitely a bit janky.  I'll follow up on this in a separate patchset.

imports/client/components/GuessQueuePage.tsx
zarvox: Ditto.

imports/client/components/PuzzleListPage.tsx
```diff
@@ -24,7 +25,7 @@ import RelatedPuzzleGroup from './RelatedPuzzleGroup';
 
 /* eslint-disable max-len */
 
-interface PuzzleListViewProps {
+interface PuzzleListViewProps extends RouteComponentProps {
   // eslint-disable-next-line no-restricted-globals
   location: Location;
```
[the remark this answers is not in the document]
zarvox: Oh yeah, this can be cleaned up.  Nice catch.

imports/client/components/Routes.tsx
zarvox: Done.

PR #317, opened 2020-10-07
Set up a basic test suite

ebroder:
Meteor's client/server split in testing makes this a little annoying to wire up. In particular, I wanted to be able to define server-side helper methods for client-side tests, but that meant that a single file was loaded on both client and server _and_ needed some client-only imports, hence the slightly messy use of require.

I think this is prooobably still a good idea but it's definitely a bit fiddly.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,26 +1,37 @@
 {
   "compilerOptions": {
-    "target": "esnext",
-    "module": "commonjs",
-    "esModuleInterop": true,
-    "allowSyntheticDefaultImports": true,
-    "baseUrl": ".",
-    "paths": {
-      "include": [
-        "./types"
-      ]
-    },
-    "alwaysStrict": true,
-    "jsx": "react",
-    "allowJs": false,
-    "noImplicitAny": true,
-    "noImplicitReturns": true,
-    "noImplicitThis": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
+    /* Basic Options */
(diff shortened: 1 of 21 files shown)


tsconfig.json
```diff
@@ -1,26 +1,37 @@
...
-      ]
-    },
-    "alwaysStrict": true,
-    "jsx": "react",
-    "allowJs": false,
+    /* Basic Options */
+    "target": "es2018",
+    "module": "esNext",
+    "lib": ["esnext", "dom"],
+    "allowJs": true,
```
zarvox, line 7: Do we actually need to `allowJs`?  We seem to be .ts/.tsx-only right now, and I'd be happy to keep it that way, unless there's something I'm missing.

tsconfig.json
```diff
@@ -1,26 +1,37 @@
...
+    "resolveJsonModule": true,
+    "types": ["node", "mocha"],
+    "esModuleInterop": true,
+    "preserveSymlinks": true
+  },
+  "exclude": [
+    "./.meteor/**",
+    "./packages/**"
+  ]
+}
```
zarvox, line 37: nit: files should have trailing newlines

zarvox (approved):

tsconfig.json
ebroder: Yeah I think that's fine. This tsconfig comes originally from `meteor create --typescript` and I think I basically didn't edit it afterwards, but agree we shouldn't be using raw JS.

PR #319, opened 2020-10-21
Use URLSearchParams to manage query strings

### LINKED ISSUE #316: Use URLSearchParams where we put query strings in URLs
See https://github.com/deathandmayhem/jolly-roger/pull/315#discussion_r500620512 and https://github.com/deathandmayhem/jolly-roger/pull/315#discussion_r500622421 -- we should probably use a more structured type, and serialize/deserialize carefully when changing the URL

zarvox:
Fixes #316


ebroder (approved):
Thanks for following up

PR #330, opened 2020-11-19
Rewrite README and split out other documentation by purpose

zarvox:
`README.md` now explains what Jolly Roger is and why you'd want to use it.
`DEVELOPING.md` explains how to set up your own instance for development.
`OPERATION.md` is a stub, but in time should help explain what else you'll need
to bring to run Jolly Roger in anger.

We should probably come up with a new place to write about internals.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,135 +1,106 @@
-Jolly Roger
-===========
+# Jolly Roger
 
 ![Jolly Roger](public/images/hero.png)
 
-The next generation Virtual HQ for Death and Mayhem, written in
-Meteor. It uses [react-router][react-router] for navigation and
-[Collection2][collection2] for schema support.
+Jolly Roger is a coordination tool for collaboratively solving puzzlehunts like the [MIT Mystery Hunt][http://web.mit.edu/puzzle/www/].
+The team Death and Mayhem created Jolly Roger in 2015 for the 2016 MIT
+Mystery Hunt, and it is still actively maintained and in production service.
 
-Getting Started
----------------
+At its core, it allows tracking puzzles and guesses in an ever-evolving hunt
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,135 +1,106 @@
...
+
+Users who do not link a Google account (or don't have or use one) can still
+participate in the sheet anonymously.
+
+### Per-puzzle chat rooms
+
+Jolly Roger provides its own persistent chat with each puzzle.  No need to
+constantly jump between the spreadsheet and a third-party chat service.  No
+losing relevant discussion to 10000-message retention limits in the middle
+of the Hunt.  Supports Markdown formatting.
```
ebroder, line 32: Technically it's a variant of Slack's fake-Markdown (`_` and `*` are not italic and bold respectively in true markdown). That's going to end up on the list of things we'll have to deal with on Discord, which uses true Markdown.

README.md
```diff
@@ -1,135 +1,106 @@
...
+### Tag-based structure
+
+In many hunts, it's not always clear from the beginning what puzzles will be
+related or may contribute to which metapuzzles.  Jolly Roger solves this by
+eschewing a top-down hierarchical round structure and instead allowing
+multiple tags to be applied to each puzzle to construct dynamic groupings as
+you gain more information about the hunt structure over time.
+
+We've found this approach able to be capable of modeling every hunt
+structure we've thrown at it.
```
ebroder, line 47: With sufficient contortion, here's looking at you Holiday Hunt

ebroder (approved):

README.md
zarvox: Oh, that's a really good point; I'd missed that before.

PR #333, opened 2020-11-20
Add `Send` button to chat input area

### LINKED ISSUE #287: Add chat submission button
The chat system currently has no button to submit text, relying instead on the enter key. On some platforms (e.g. Touchscreen devices without a paired keyboard) there is currently no way to submit. We should add this button.

We could potentially hide the button on non-mobile devices, but it's more accessible to keep it and I don't think there's much benefit to omitting it.

zarvox:
On mobile devices the keyboards often don't have an Enter key, so a separate
"Send" button is needed.

Fixes #287

button idle

resized to content on desktop

mobile size

mobile size with content

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -280,6 +280,16 @@ table.puzzle-list {
   }
 }
 
+.chat-input-row {
+  display: flex;
+  flex-direction: row;
+  align-items: stretch;
+  justify-content: space-between;
+  textarea {
+    flex: 1 1 auto;
+  }
+}
+
 .chat-history {
   flex: 1 1 auto;
   overflow-y: auto;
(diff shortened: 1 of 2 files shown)


zarvox:
I'm gonna go ahead and merge this, since it seems pretty straightforward, but feel free to poke me to take another look if somehow this becomes busted on Safari or something.

grantaelliott:
Yup, problems on Safari (and likely elsewhere). The chat-input-row is scaling down with the chat-section. I'm pretty sure it just needs a flex-shrink:0 or more completely flex:none.

Thanks for adding this, by the way.

client/stylesheets/puzzle.scss
```diff
@@ -280,6 +280,16 @@ table.puzzle-list {
   }
 }
 
+.chat-input-row {
+  display: flex;
```
grantaelliott, line 284: I believe this needs flex:none; to render as intended in Safari.

zarvox: Fixed via https://github.com/deathandmayhem/jolly-roger/commit/f40316407639ef2b5f942ed9f98411ab229f92bf -- thanks for testing and coming up with the fix!

PR #335, opened 2020-11-25
Polish chat controls

grantaelliott:
Minor changes to chat controls:
* Style tweaks, including improved control consistency
* Text area always gets focus after submit (previously never)
* Send button disabled when no text entered
* Textarea and button behave as one control, with shared focus and tab-index

Before:
before-no-focus
After:
after-no-focus

Before with content and focus:
before-focus
After with content and focus:
after-focus

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -1,3 +1,6 @@
+@import "../../node_modules/bootstrap/scss/functions";
+@import "../../node_modules/bootstrap/scss/variables";
+
 // Puzzle components helper styles
 // Puzzle lists/tables
 .puzzle-group {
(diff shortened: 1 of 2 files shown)


client/stylesheets/puzzle.scss
```diff
@@ -286,8 +289,23 @@ table.puzzle-list {
   flex-direction: row;
   align-items: stretch;
   justify-content: space-between;
+  border-top: $input-border-width solid $input-border-color;
+  overflow: hidden;
+  &:focus-within {
```
zarvox, line 294: today I learned about `focus-within`!

zarvox (approved):

Issue #233, opened 2019-01-29
Keyboard shortcut to close celebrations


ebroder:
Currently, when a celebration pops up, you have to click the "X" in order to close it (before it finishes playing its audio, or whatever). It should also be possible to hit escape to close out of a celebration.

PR #339, opened 2020-11-25
Reinstitute celebrations and support close on esc press

grantaelliott:
Reinstitutes celebrations (removed as part of multiple answer support). A celebration occurs if the answers field contains an answer that was not previously present. Multiple celebrations are queued if necessary.

Also addresses #233, closing the celebration when esc is pressed.

### DIFF
--- a/imports/client/components/Celebration.tsx
+++ b/imports/client/components/Celebration.tsx
@@ -14,10 +14,12 @@ class Celebration extends React.Component<CelebrationProps> {
 
   componentDidMount() {
     this.timer = window.setTimeout(() => { this.onClose(); }, 7000);
+    document.addEventListener('keydown', this.onKeyDown);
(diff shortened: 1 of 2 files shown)


zarvox:
I'm vaguely remembering that we had weirdness around celebrations when people would suspend/resume their devices and then get like a zillion of them on reconnect, or in all their tabs at once (which would be overpoweringly loud).  I'm gonna hold off on merging this until I have a chance to think a bit more about the corner cases of user experience here, but will consider this an impetus to do that thinking again. :)

(I'm generally in support of celebrations, though, especially since we're effectively all remote this year)

grantaelliott:
Ah, that makes sense. Feel free to close this PR and/or just grab the esc keydown patch for the future behavior fix.

PR #341, opened 2020-11-25
Fix two links in README

grantaelliott:
Fix a markdown typo and a filename mismatch. (Assuming DEVELOPMENT was the intended filename, based on sibling OPERATION.)


zarvox (approved):
Thanks for the cleanups!

PR #345, opened 2020-11-28
Add mode without guess queue

aldeka:
This adds a hasGuessQueue attribute to the Hunt model. Said attribute is on by default, so the user guess -> operator approval flow will still be the case for instances unless this is explicitly set by the operator.

When hasGuessQueue is false, users can able to submit answers directly, without the guess UI or operator approval. In that mode, additionally, users can remove answer(s) from the list on the puzzle page (if, e.g. they made a typo the first time).

### DIFF
--- a/imports/server/guesses.ts
+++ b/imports/server/guesses.ts
@@ -130,19 +130,12 @@ Meteor.methods({
       guess: answer,
       state: 'correct',
     });
+
     const savedAnswer = Guesses.findOne(answerId);
     if (!savedAnswer) {
       throw new Meteor.Error(404, 'No such correct guess');
     }
-    Guesses.update({
-      _id: savedAnswer._id,
-    }, {
-      $set: {
-        state: 'correct',
-      },
-    });
     addChatMessage(savedAnswer, 'correct');
-
     Puzzles.update({
       _id: savedAnswer.puzzle,
     }, {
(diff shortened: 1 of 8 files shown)


imports/server/guesses.ts
```diff
@@ -86,6 +97,78 @@ Meteor.methods({
...
+      _id: savedAnswer.puzzle,
+    }, {
+      $addToSet: {
+        answers: savedAnswer.guess,
+      },
+    });
+    GlobalHooks.runPuzzleSolvedHooks(savedAnswer.puzzle);
+  },
+
+  removeAnswerFromPuzzle(puzzleId: unknown, answer: unknown) {
```
zarvox, line 149: This feels like it'd be better taking a `guessId` than a puzzleId + answer, since the former is reliably unique and the latter is not.

imports/server/guesses.ts
```diff
@@ -86,6 +97,78 @@ Meteor.methods({
...
+        hunt: 1,
+      },
+    });
+    const huntId = puzzle && puzzle.hunt;
+    const hunt = Hunts.findOne({ _id: huntId });
+    if (!huntId || !hunt || hunt.hasGuessQueue) {
+      throw new Error(`Hunt ${huntId} does not support self-service answers`);
+    }
+
+    const guess = Guesses.findOne({ puzzle: puzzleId, guess: answer });
```
zarvox, line 167: If there are two guesses that have the same answer, and they're both wrong, you could get into a situation where this `findOne()` always retrieves the one that's already marked incorrect and the other one can never be changed, no matter how many times you hit the X button.  Using guess IDs would avoid this corner case.

zarvox:
cc @ebroder do you know if there's a reason builds/checks didn't run on this PR?  I note that it's from a different tree (aldeka/jolly-roger rather than deathandmayhem/jolly-roger) so I can understand not running arbitrary code by default, but I'm not sure how to mark this PR as "yes please run tests/actions" so checks will actually run.

ebroder:
Hmm. It looks like [workflows don't run on forks by default](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#:~:text=Workflows%20don't%20run%20on%20forked%20repositories%20by%20default), but it's possible to enable them in the repository settings (I think [here](https://github.com/aldeka/jolly-roger/settings/actions) for @aldeka's fork).

I think we also can set the repository to trigger on the [`pull_request` event](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request), although it's not entirely clear to me whether that re-fires when someone pushes an update to a PR.

zarvox:
> it's not entirely clear to me whether that re-fires when someone pushes an update to a PR

FWIW it looks like it does!

PR #346, opened 2020-11-28
Add hunt homepage link to landing pages

### LINKED ISSUE #292: Add link to hunt home
It would be nice for Jolly Roger to link to hunt homepages. Currently, we usually do this through an administrivia item. The link would likely appear near the top of the hunt landing page (the page dominated by the list of puzzles).

(We could also conceivably add a mechanism to assign links to groups, so that round pages could be directly linked, but that is likely more trouble than it's worth.)

grantaelliott:
Add an optional field to Hunts to store hunt homepage urls, rendered as a link on the puzzle list page. Also, no longer wait for puzzle and tag data to render the hunt navigation links. General cleanup of PuzzleListPage, including removal of redundant and unreferenced markup.

No migration added, as it appears we aren't creating migrations for the creation of optional fields. Please correct me if I'm wrong.

Closes #292

### DIFF
--- a/imports/client/components/PuzzleListPage.tsx
+++ b/imports/client/components/PuzzleListPage.tsx
@@ -2,6 +2,8 @@ import { Meteor } from 'meteor/meteor';
 import { Roles } from 'meteor/nicolaslopezj:roles';
 import { withTracker } from 'meteor/react-meteor-data';
 import { _ } from 'meteor/underscore';
+import { faExternalLinkAlt } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import React from 'react';
 import Button from 'react-bootstrap/Button';
 import ButtonToolbar from 'react-bootstrap/ButtonToolbar';
@@ -13,8 +15,10 @@ import ToggleButton from 'react-bootstrap/ToggleButton';
 import ToggleButtonGroup from 'react-bootstrap/ToggleButtonGroup';
 import { RouteComponentProps } from 'react-router';
 import { Link } from 'react-router-dom';
+import Hunts from '../../lib/models/hunts';
 import Puzzles from '../../lib/models/puzzles';
 import Tags from '../../lib/models/tags';
+import { HuntType } from '../../lib/schemas/hunts';
 import { PuzzleType } from '../../lib/schemas/puzzles';
 import { TagType } from '../../lib/schemas/tags';
 import PuzzleList from './PuzzleList';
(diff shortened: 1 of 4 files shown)


imports/client/components/PuzzleListPage.tsx
```diff
@@ -424,37 +428,52 @@ interface PuzzleListPageProps extends PuzzleListPageWithRouterParams {
...
+            allTags={this.props.allTags}
+          />
+        </div>
       );
     }
   }
 }
 
 const PuzzleListPageContainer = withTracker(({ match }: PuzzleListPageWithRouterParams) => {
+  const huntHandle = Meteor.subscribe('mongo.hunts', { _id: match.params.huntId });
```
zarvox: No need to subscribe here -- we've already subscribed to `mongo.hunts` with that `huntId` up in HuntApp.tsx, which is higher in the tree.  By the time we're rendering PuzzleListPage, we're guaranteed we've already got the hunt.

So remove this line, don't worry about tracking if it's ready or not, and just always pass the hunt down.

imports/client/components/PuzzleListPage.tsx
```diff
@@ -424,37 +428,52 @@ interface PuzzleListPageProps extends PuzzleListPageWithRouterParams {
   canUpdate: boolean;
   allPuzzles: PuzzleType[];
   allTags: TagType[];
+  hunt?: HuntType;
```
zarvox, line 422: Make this just `hunt: HuntType;` and use a TS assertion `!` below where you put it in the object.

(separately: in most places I'd rather include keys in objects and have their values possibly be `undefined` than have optional keys, since that's generally how we get data out of Meteor so it's less noisy and I anticipate it being slightly easier for the JIT to optimize, since there's only one set of fields to recognize for the "struct".  The way to express that would be `hunt: HuntType | undefined;`)

imports/client/components/PuzzleListPage.tsx
```diff
@@ -424,37 +428,52 @@ interface PuzzleListPageProps extends PuzzleListPageWithRouterParams {
...
+  const huntHandle = Meteor.subscribe('mongo.hunts', { _id: match.params.huntId });
   const puzzlesHandle = Meteor.subscribe('mongo.puzzles', { hunt: match.params.huntId });
   const tagsHandle = Meteor.subscribe('mongo.tags', { hunt: match.params.huntId });
 
   // Don't bother including this in ready - it's ok if it trickles in
   Meteor.subscribe('subscribers.counts', { hunt: match.params.huntId });
 
-  const ready = puzzlesHandle.ready() && tagsHandle.ready();
+  const ready = huntHandle.ready() && puzzlesHandle.ready() && tagsHandle.ready();
+  const hunt = Hunts.findOne({ _id: match.params.huntId });
```
zarvox, line 472: Go ahead and make this:

```
// Hunt is guaranteed to have been loaded by HuntApp already.
const hunt = Hunts.findOne({ _id: match.params.huntId })!;
```

The `!` at the end asserts to typescript that this will not be `undefined`.

imports/client/components/PuzzleListPage.tsx
```diff
@@ -470,6 +489,7 @@ const PuzzleListPageContainer = withTracker(({ match }: PuzzleListPageWithRouter
       canUpdate: Roles.userHasPermission(Meteor.userId(), 'mongo.puzzles.update'),
       allPuzzles: Puzzles.find({ hunt: match.params.huntId }).fetch(),
       allTags: Tags.find({ hunt: match.params.huntId }).fetch(),
+      ...(hunt ? { hunt } : { }),
```
zarvox: With the `!` above, you can make this just `hunt,`

imports/client/components/PuzzleListPage.tsx
```diff
@@ -424,37 +428,52 @@ interface PuzzleListPageProps extends PuzzleListPageWithRouterParams {
   canUpdate: boolean;
   allPuzzles: PuzzleType[];
   allTags: TagType[];
+  hunt?: HuntType;
 }
 
 class PuzzleListPage extends React.Component<PuzzleListPageProps> {
   render() {
+    const leadLinks = this.props.hunt?.homepageUrl && (
```
zarvox, line 427: once you've asserted hunt is always truthy you can remove the `?` after `hunt` here

imports/client/components/PuzzleListPage.tsx
grantaelliott: I was wondering how we felt about that redundancy in the name of modularity. It's not hard to imagine a change breaking this assumption, but I think I'm ok with it.

imports/client/components/PuzzleListPage.tsx
grantaelliott: Summarizing a productive Slack conversation with @zarvox:

Since HuntApp already takes care of the hunt subscription and catches non-existent hunts, the assertion is reasonable in this case and, from a maintenance perspective, has essentially the same impact as removing the redundant subscription. In the more general case, our preference would be to use required props that may be undefined rather than optional props. Either approach eliminates awkward constructions, like the conditional spread below, that are required with optional props.

Optional props are used for this case in quite a few other places, which we may want to clean up.

zarvox:
Looks good to me code-wise.

Design-wise, I find the link a bit large.  It takes up a fair bit of vertical space at the top.  Could we fit the link in somewhere/somehow that would avoid using so much vertical space?  Maybe another item in the Announcements/Guess queue/Hunters list?

Screenshot_20201128_174219

aldeka:
UI proposal: the quickest and dirtiest way (aka doesn't involve potentially frobbing breadcrumbs) would be to add 'Hunt Homepage' to the announcements/guess queue/hunters row.

grantaelliott:
I was trying to avoid adding the link to that navigation row without style changes, as I believe an external link should be visually distinct from Jolly Roger navigation, but yeah the stand-alone link isn't great either. (Relatedly, that navigation row could potentially be elevated to appear on all hunt-related pages with the addition of a fourth "Puzzle List" option and .active styling, in which case the distinction would be even more important.) I'll see if I can come up with something quickly this afternoon, but otherwise adding it to the row is a reasonable interim choice - after all, it's not exactly a top priority problem.

grantaelliott:
Gave it a visually distinct style within the row. The choice of color would probably have to change if this navigation were to add active styling. (It actually looks pretty good in dark gray, but I thought that created some confusion with the secondary buttons directly below it.)

updated navigation row

zarvox (approved):
LGTM.  I rebased for you and will merge on successful build.

PR #348, opened 2020-11-28
Run tests on PRs

ebroder:


ebroder:
Maybe we only want to run on `pull_request`? It seems like we're going to get 2 builds otherwise?

PR #349, opened 2020-11-29
Add `PublicSettings` for server-specific settings that should be public

zarvox:
"public" here means "published to all clients at all times, no ACLs
whatsoever".  It's a transport mechanism for things you'd put in source code if
you didn't want to support different values in different deployments.

This will be used for WebRTC STUN/TURN server configuration, and could be used
for rebranding.

### DIFF
--- a/imports/lib/models/public_settings.ts
+++ b/imports/lib/models/public_settings.ts
@@ -7,7 +7,7 @@ PublicSettings.attachSchema(PublicSettingsSchema);
 
 // All public settings are accessible by all clients at all times, including to
 // not-logged-in-users.
-PublicSettings.publish();
+Meteor.publish('mongo.public_settings', () => PublicSettings.find());
 
 // Public settings should always be subscribed to by all clients.
 if (Meteor.isClient) {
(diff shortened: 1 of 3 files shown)


imports/lib/models/public_settings.ts
```diff
@@ -0,0 +1,17 @@
+import { Meteor } from 'meteor/meteor';
+import PublicSettingsSchema, { PublicSettingType } from '../schemas/public_settings';
+import Base from './base';
+
+const PublicSettings = new Base<PublicSettingType>('public_settings');
+PublicSettings.attachSchema(PublicSettingsSchema);
+
+// All public settings are accessible by all clients at all times, including to
+// not-logged-in-users.
+PublicSettings.publish();
```
ebroder, line 10: I don't think you can use the `Base#publish` helper for collections that you want to be accessible to non-logged-in-users: https://github.com/deathandmayhem/jolly-roger/blob/e99fd810fafc26b3c0648ec0ac64d6ab7474b4d4/imports/lib/models/base.ts#L139-L141

Probably easiest to just publish it directly

```suggestion
Meteor.publish('mongo.public_settings', () => PublicSettings.find());
```

ebroder (approved):

PR #350, opened 2020-12-01
Render tagless group consistently

### LINKED ISSUE #347: Puzzles with no group have different behavior in puzzle list
In the PuzzleListView, the implicit group of puzzles with no group currently renders slightly differently:

- The section heading ("Puzzles in no group") has no preceding down caret, so it appears to be spaced differently (and it has a trailing colon)
- The section cannot be collapsed

It's a very minor concern, but I think this group should be collapsible, like the others. Was this a deliberate decision?

grantaelliott:
Render tagless puzzle groups in the same way as ordinary groups, including collapse behavior. Achieved by giving RelatedPuzzleGroup support for null sharedTag. (Though the .tag class is used to achieve consistent styling, the Tag component has not been extended to support no tag.)

Closes #347

### DIFF
--- a/imports/client/components/PuzzleListPage.tsx
+++ b/imports/client/components/PuzzleListPage.tsx
@@ -308,7 +308,7 @@ class PuzzleListView extends React.Component<PuzzleListViewProps, PuzzleListView
           return (
             <RelatedPuzzleGroup
               key={g.sharedTag ? g.sharedTag._id : 'ungrouped'}
-              sharedTag={g.sharedTag || null}
+              sharedTag={g.sharedTag}
               noSharedTagLabel="(no group specified)"
               relatedPuzzles={g.puzzles}
               allTags={this.props.allTags}
--- a/imports/client/components/PuzzleListPage.tsx
+++ b/imports/client/components/PuzzleListPage.tsx
@@ -305,33 +305,18 @@ class PuzzleListView extends React.Component<PuzzleListViewProps, PuzzleListView
       case 'group': {
         const puzzleGroups = this.puzzleGroupsByRelevance();
         const groupComponents = puzzleGroups.map((g) => {
-          if (g.sharedTag) {
-            return (
-              <RelatedPuzzleGroup
-                key={g.sharedTag._id}
-                sharedTag={g.sharedTag}
-                relatedPuzzles={g.puzzles}
-                allTags={this.props.allTags}
-                includeCount={false}
-                layout="grid"
-                canUpdate={this.props.canUpdate}
-              />
-            );
-          } else {
-            return (
-              <div key="ungrouped" className="puzzle-list-ungrouped">
-                <div>Puzzles in no group:</div>
-                <div className="puzzles">
-                  <PuzzleList
-                    puzzles={g.puzzles}
-                    allTags={this.props.allTags}
-                    layout="grid"
-                    canUpdate={this.props.canUpdate}
-                  />
-                </div>
-              </div>
-            );
-          }
+          return (
+            <RelatedPuzzleGroup
+              key={g.sharedTag ? g.sharedTag._id : 'ungrouped'}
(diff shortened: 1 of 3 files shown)


imports/client/components/PuzzleListPage.tsx
```diff
@@ -305,33 +305,18 @@ class PuzzleListView extends React.Component<PuzzleListViewProps, PuzzleListView
...
-                    canUpdate={this.props.canUpdate}
-                  />
-                </div>
-              </div>
-            );
-          }
+          return (
+            <RelatedPuzzleGroup
+              key={g.sharedTag ? g.sharedTag._id : 'ungrouped'}
+              sharedTag={g.sharedTag || null}
```
zarvox, line 311: Prefer `undefined` over `null` in general, unless there's a concrete reason you have to use `null` (like compatibility with a particular API).

```suggestion
              sharedTag={g.sharedTag}
```

imports/client/components/RelatedPuzzleGroup.tsx
```diff
@@ -12,12 +12,12 @@ import puzzleInterestingness from './puzzleInterestingness';
 
 function sortPuzzlesByRelevanceWithinPuzzleGroup(
   puzzles: PuzzleType[],
-  sharedTag: TagType,
+  sharedTag: TagType | null,
```
zarvox, line 15:
```suggestion
  sharedTag: TagType | undefined,
```

imports/client/components/RelatedPuzzleGroup.tsx
```diff
@@ -37,7 +37,9 @@ function sortPuzzlesByRelevanceWithinPuzzleGroup(
 }
 
 interface RelatedPuzzleGroupProps {
-  sharedTag: TagType;
+  sharedTag: TagType | null;
```
zarvox, line 40:
```suggestion
  sharedTag: TagType | undefined;
```

imports/client/components/RelatedPuzzleGroup.tsx
```diff
@@ -37,7 +37,9 @@ function sortPuzzlesByRelevanceWithinPuzzleGroup(
...
+  // noSharedTagLabel is used to label the group only if sharedTag is null.
+  noSharedTagLabel: String;
```
zarvox, line 42: I was confused at first by this since we don't specify it in RelatedPuzzleGroups.tsx, but then I saw the `defaultProps` down below.  It feels a little weird, but I can roll with it, and RelatedPuzzleGroups doesn't wind up with an undefined `sharedTag` ever.

imports/client/components/PuzzleListPage.tsx
grantaelliott: I used null just because it’s harder to do by accident, so type checking offers a little more benefit, but I’m happy to change it.

imports/client/components/RelatedPuzzleGroup.tsx
grantaelliott: It actually is passed through in the RelatedPuzzleGroup on PuzzleListPage (which now can support a null sharedTag). My thinking was that we use RelatedPuzzleGroup in a few contexts and the message most helpful here (that no group has even been set) wouldn’t necessarily apply to other instances of “default” groups in the future.

imports/client/components/RelatedPuzzleGroup.tsx
```diff
@@ -37,7 +37,9 @@ function sortPuzzlesByRelevanceWithinPuzzleGroup(
 }
 
 interface RelatedPuzzleGroupProps {
-  sharedTag: TagType;
+  sharedTag: TagType | null;
+  // noSharedTagLabel is used to label the group only if sharedTag is null.
```
grantaelliott, line 41:
```suggestion
  // noSharedTagLabel is used to label the group only if sharedTag is undefined.
```

PR #354, opened 2020-12-03
Add in-app first-user-creation flow for server admin

zarvox:
It's actually kinda hard to set up a new production Jolly Roger deployment,
since there's no good way to get a meteor shell with a connection to the real
DB to create the first account.  You either have to figure out how to get a
shell hooked up to your production DB (which we don't document), or you have to
bring a row from your development users table via a mongo client, or something else
ridiculous.  We can do better.

This change adds:

* a pseudocollection that simply indicates whether there are any
  users in the Meteor users table
* a Meteor method to provision the initial admin user provided there
  are no users in the Meteor users table yet
* a page with a form to enter an email address/password to be used with the
  aforementioned Meteor method
* logic in the root redirector to send you to that form if you are not logged in
  and the server has no users

Then, when setting up a new server, you navigate to the root page, it redirects
to the first-account-creation page, and you can just create the admin user
through the UI.

Screenshot_20201203_032249

cc @aldeka since this was a pain point in setting up a test Jolly Roger instance for Rages

### DIFF
--- a/imports/client/components/FirstUserForm.tsx
+++ b/imports/client/components/FirstUserForm.tsx
@@ -1,0 +1,141 @@
+import { Meteor } from 'meteor/meteor';
+import React from 'react';
+import Alert from 'react-bootstrap/Alert';
+import Button from 'react-bootstrap/Button';
+import FormControl, { FormControlProps } from 'react-bootstrap/FormControl';
+import FormGroup from 'react-bootstrap/FormGroup';
+import FormLabel from 'react-bootstrap/FormLabel';
+
+enum SubmitState {
+  IDLE = 'idle',
+  SUBMITTING = 'submitting',
+  SUCCESS = 'success',
+  ERROR = 'error',
+}
+
+interface FirstUserFormProps {
+}
+
+interface FirstUserFormState {
+  email: string;
+  password: string;
+  submitState: SubmitState;
+  submitError: string;
+}
+
+class FirstUserForm extends React.Component<FirstUserFormProps, FirstUserFormState> {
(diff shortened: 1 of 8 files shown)


zarvox:
Argh.  The acceptance tests pass locally, but sometimes fail (due to what appears to be variance in system performance) on CI.

What's more, there's a 2 second timeout for each test somewhere in one of these test frameworks, so I can't make the sleeps longer.  There's probably a way to tune the 2000-msec timeout somewhere, but I'll need to track it down (and I feel lousy just throwing timeouts at problems).  Not sure if there's a good answer for "keep waiting until the conditions are met" with an overall test runner timeout.

Alternately, make some way to get a promise that resolves when all the Meteor subs are ready or something?  Seems pretty overkill for this though.

ebroder:
> Alternately, make some way to get a promise that resolves when all the Meteor subs are ready or something? Seems pretty overkill for this though.

Would it be ridiculous to wire something up using `Meteor.connection._subscriptions`? 😈

ebroder:
Heh. Hilariously, I already added [deanius:promise](https://atmospherejs.com/deanius/promise) which has the hooks to do it, but the `readyPromise` method doesn't seem to show up on the subscription objects in `Meteor.connection._subscriptions` womp womp.

ebroder (approved):

zarvox:
I don't see a great way to get a promise to trigger for what I want, but I can:

1. Set test-specific timeouts
2. Roll my own poll loops so the test isn't just twiddling its thumbs for seconds at a time.

I think I'm cool merging this once it passes the build (which I expect it will).

PR #361, opened 2020-12-09
Move related puzzles to per-tag popovers

### LINKED ISSUE #336: Proposal: Move Related Puzzles to modal
Particularly in light of the renewed importance of chat this year, I'd like to revisit the question of what to do with Related Puzzles. In the past, I've been a proponent of keeping it at least available as always visible, but I now think it would be better to move it to a modal, with a trigger in the Metadata pane (as with Viewers and Guesses). The current behavior is to collapse the Related Puzzles automatically except on puzzles with meta tags.

Pros:
* Prioritize chat
* More space for related puzzles in the modal (less wrapping of titles and, in extreme cases, cutoff of answers)
* Unify the behavior of Guesses, Viewers, and Related Puzzles as parts of the metadata
* Eliminate the short-height layouts
* Simplify the interface visually
* Remove ability to hide chat on low-height mobile layouts

Cons:
* Lose at-a-glance pattern spotting (now one click away)
  * In non-metas, spotting patterns in sibling answers is already lost for most users as Related Puzzles is automatically collapsed.
  * In metas, child answers are usually copied into the sheet quickly.
  * Losing at-a-glance pattern spotting between sibling metas is notable, though it doesn't come up often.
* Lose at-a-glance awareness of a new child answer
  * The notification system serves the same purpose and is much more obvious.

I'd be interested in hearing opinions on this, particularly from those who frequently work on metas.

(The impetus for rethinking this was actually a long-standing but very minor issue for some mobile users that is now unavoidable even within the current generation iPhone lineup. On the puzzle page in narrow view, the transition from adjustable panes to tabs may occur when the browser chrome is hidden, creating a jarring user experience. Currently this affects the iPhone 12 and 12 Pro.)

grantaelliott:
Closes #336 

Based on a suggestion from @zarvox, this moves related puzzles out of the sidebar and into popovers opened by hovering over tags. On mobile, the popover can be opened by tapping the tag and closed by tapping outside. The popovers resize appropriately in small viewports and can be scrolled as necessary.

Tested in Safari, Chrome, and Safari for iOS.

This required factoring out RelatedPuzzleList from RelatedPuzzleGroup. RelatedPuzzleGroups is no longer in use and was removed.

Several upcoming PRs propose style and layout improvements. (e.g. fixes for minor style issues in the table view that are more apparent in this presentation than they were previously.)


client/stylesheets/puzzle.scss
```diff
@@ -633,6 +600,16 @@ table.puzzle-list {
   left: 0;
 }
 
+.popover.related-puzzle-popover {
+  max-width: none;
+  display: flex;
+  flex-direction: column;
+
+  .popover-body {
+    overflow: scroll;
```
zarvox, line 609: Any reason why this is `scroll` rather than `auto`?
```suggestion
    overflow: auto;
```

imports/client/components/PuzzlePage.tsx
```diff
@@ -54,109 +50,53 @@ const FilteredChatFields: ('_id' | 'puzzle' | 'text' | 'sender' | 'timestamp')[]
...
-const DefaultChatHeightFraction = 0.6;
 
 // PuzzlePage has some pretty unique properties:
 //
 // * It's the only page which iframes other sites.  Doing so requires that we
 //   specify the absolute size and position of the iframe, which makes us need
 //   position: fixed.
-// * There's up to four interesting pieces of content shown on this page:
-//   * Related puzzles
+// * There's up to three interesting pieces of content shown on this page:
```
zarvox, line 62: Thanks for updating all the comments in through here.  The simplification is great. :)

zarvox (approved):
This feels fantastic.  More space on the page for chat & presence, contextually-relevant popovers, and more code removed than added.  :100:

Issue #291, opened 2020-01-21
Overhaul emoji support


zarvox:
We should probably come up with a uniform emoji support philosophy.

We probably want to figure out what fields should support emoji, and with what font face(s), in what order of preference, and how.  For instance, right now, whitespace in answers are getting emoji font treatment, which makes them wide and blocky.  I believe the same may be true of numerals in answer blocks.  Let's audit all of that.

We should also figure out how to make this work without enabling XSS.

We may wish to support github/slack-style colon-codes for emoji substitution/emoji picking.

zarvox:
https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range seems like it might help us out here with the immediate issue of "whitespace gets emoji-width spacing under `.answer`" -- we can declare an `@font-face` naming a new `font-family`, then only apply it to particular unicode ranges (e.g. U+1????).

Then we can apply those `font-family`s to `.answer`, and since they won't cover U+20, they'll fall back to monospace.

ebroder:
Now that it's no longer the middle of hunt, I do want to spend some time thinking about what the right behavior here is, and I wonder if it's worth thinking about it independently for answers vs. chat.

For chat, it feels like we should try to find something wholistic that handles Slack-style formatting (`*`, `_`, `>`, `` ` ``, etc., but also emoji). For answers, my intuition is that I still would mildly prefer to plug in twemoji, so that we get cross-platform consistency, rather than continue to fight with native font rendering.

zarvox:
This is done!  We have a standard set of fonts, and specify specific unicode ranges for which to prioritize the use of native emoji fonts, and we apply that throughout the app.

PR #364, opened 2020-12-11
Fix emoji spacing

grantaelliott:
Addresses the space formatting issue discussed in #291, using @zarvox's suggestion, but does not address the bigger questions raised by @ebroder.

With this change, emoji fonts are prioritized for blocks that people tend to describe as emoji (see below), after which we prefer monospace. For characters not available there, we fall back to the rest of the emoji fonts. My intent was to prefer monospace wherever it makes sense, conceding that it probably doesn't for clearly pictographic symbols.

The prioritized emoji include the self-evident 1F300-1FAFF, but also 2300-23FF (Miscellaneous Technical, which happens to include various timepiece symbols), 2600-26FF (Miscellaneous Symbols), 2700-27BF (Dingbats), and 1F100-1F1FF (Enclosed Alphanumeric Supplement, which includes the regional indicator symbols used for flags). It intentionally does not include arrows and the like, for which I think we should prefer monospace variants when available. That said, I could easily be persuaded to change any of this - the important change is that 0020 return to monospace.

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -134,7 +134,7 @@ table.puzzle-list {
 @font-face {
   font-family: "Emoji";
   src: "Noto Color Emoji", "Apple Color Emoji", "Segoe UI Emoji";
-  unicode-range: U+1F300-1FAFF, U+2300-23FF, U+2600-27BF;
+  unicode-range: U+1F300-1FAFF, U+1F100-1F1FF, U+2300-23FF, U+2600-27BF;
 }
 
 .answer, #jr-puzzle-guess {
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -131,9 +131,15 @@ table.puzzle-list {
   margin: 2px;
 }
 
+@font-face {
+  font-family: "Emoji";
+  src: "Noto Color Emoji", "Apple Color Emoji", "Segoe UI Emoji";
+  unicode-range: U+1F300-1FAFF, U+2300-23FF, U+2600-27BF;
+}
+
 .answer, #jr-puzzle-guess {
   text-transform: uppercase;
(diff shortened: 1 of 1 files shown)


client/stylesheets/puzzle.scss
```diff
@@ -131,9 +131,15 @@ table.puzzle-list {
...
+@font-face {
+  font-family: "Emoji";
+  src: "Noto Color Emoji", "Apple Color Emoji", "Segoe UI Emoji";
+  unicode-range: U+1F300-1FAFF, U+1F100-1F1FF, U+2300-23FF, U+2600-27BF;
+}
+
 .answer, #jr-puzzle-guess {
   text-transform: uppercase;
-  font-family: "Noto Color Emoji", "Apple Color Emoji", "Segoe UI Emoji", monospace;
+  font-family: "Emoji", monospace, "Noto Color Emoji", "Apple Color Emoji", "Segoe UI Emoji", sans-serif;
```
zarvox, line 142: Just out of curiosity: do we expect the additional fallbacks after monospace ("Noto Color Emoji", "Apple Color Emoji", "Segoe UI Emoji", sans-serif) to actually ever be reached?  (Seems harmless to have them just in case, and if there's a range that isn't covered by the system monospace font one of those others is more likely to have at least *some* glyph, I guess?)

zarvox (approved):

client/stylesheets/puzzle.scss
grantaelliott: Mostly it just felt like good practice. I didn’t prioritize most code points that are basically stylized characters. Ideally, monospace will handle those but in practice I imagine it frequently won’t. Similarly, I expect some unusual scripts (One day, there will be an answer in Linear B...) will fall all the way through and specifying sans feels better than not.

But in practice, I doubt it’s that different from just letting the system fallbacks take over. And if any of those cases actually comes up, we’ll probably patch the font list to handle it.

PR #367, opened 2020-12-11
Improve puzzle list presentation

grantaelliott:
General improvement of puzzle list presentation format and style.
- **All formats**
  - Answers now display stacked one-per-line: _Improved visual inspection of all answers_
  - Answers that wrap are indented: _Distinguishes wrapped answers at a glance_
  - Change alignment to first baseline (Formerly vertically centered): _Visually aligns the two fonts used for title and answer (particularly noticeable in popover tables). In grid format, also places interface elements consistently in rows that wrap._
  - Tags now break at colons: _Better wrapping_
- **Grid format** (as seen on the main puzzle list)
  - Answers now wrap first at word-breaks, then by letter
  - Move answers to a new line on narrow viewports: _Less text wrapping_
  - Move edit button into its own flex item: _Title no longer wraps under it_
  - Resize edit button to fit in the same line-height
  - Swap puzzle-link and viewer count: _Places the viewer count next to the JR link, which is a more natural association, without significantly changing existing user experience_
  - Minor adjustments to flexbox spacing

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -20,9 +20,6 @@
 }
 
 .puzzle {
-  margin-bottom: 4px;
-  vertical-align: top;
-
   &.unsolved {
     background-color: #f0f0f0;
   }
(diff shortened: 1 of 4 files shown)


zarvox (approved):
Tried this out.  Looks and feels great.

I still observe some alignment oddities in Firefox (ideally single-row text would feel like it winds up centered vertically within the box with colored background), but I'm also seeing them before this patchset, so no reason not to go for this.

grantaelliott:
Precise layout of the single-row case was a goal here, so I'd like to get that resolved. Can you post a screenshot?

zarvox:
Chrome on Linux seems fine:

Screenshot_20201212_011606

Firefox on Linux (same machine), the text (indeed, the whole baseline) looks slightly too high, and is especially noticeable inside the tags:

Screenshot_20201212_011657

grantaelliott:
Unfortunately, this is going to be frustrating to fix because that's not what I see. Safari, Chrome, and Firefox on my Mac all render it with the expected alignment. We do have different fonts, however, which could affect this. At least the consistent baseline helps.

I have an upcoming PR standardizing fonts. (It seemed out of scope for this PR.) Perhaps that will address this issue. In the meantime, I suggest we land this PR, as it's still an improvement on other points, and it sounds like this issue precedes it.

grantaelliott:
I think I've tracked this down and I'm pretty sure it's just a font issue. Chrome and Firefox on your machine are just substituting different fonts for Helvetica Neue, neither of which is metric-compatible - nor are they compatible with each other.

In any case, I'm pretty confident that this will be fixed by standardizing fonts, which I was already working on anyway.

PR #368, opened 2020-12-11
Suggestion: Remove "Tags:" label from puzzle page

grantaelliott:
We actually [discussed this in January](https://github.com/deathandmayhem/jolly-roger/pull/215#discussion_r248921246), but it might be worth revisiting in light of recent layout changes.

Removing the label frees up some space and generally makes the layout cleaner, in my opinion, and hopefully the purpose of tags is pretty self-evident even without the label. I added a "No tags yet" label, so the + button makes sense in that context.

When I put this together, I had forgotten that we already discussed this. @zarvox, please feel free to just cite the old discussion if your opinion hasn't changed.

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -214,6 +214,11 @@ table.puzzle-list {
   margin: 0 0 0 6px;
 }
 
+.tag-list-empty-label {
+  color: #808080;
+  margin-right: 4px;
+}
+
 /*Guessing and History Modal */
 .guess-history-table td.answer {
   word-break: break-all;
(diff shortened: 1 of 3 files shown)


zarvox (approved):
Seems reasonable.

zarvox:
(discussed this privately, but adding it to the PR for posterity)

I didn't recall/think of my previous objections again until I read them where you linked them, and I'm less attached to them now.  While it's true that people might not know what to call the tags, it's not the only place in the UI that the term "tags" shows up.  And the alignment thing is real.

While in general most puzzles will never be tagless (because the operator who created them places ~all puzzles in a group), we refer to Tags in the puzzle edit dialog (which only operators see), in the search bar ("tags" is in the placeholder text), and if I still have qualms about needing to give a name to the things I'd be happy enough to just put it in the popover (e.g. make the top line of the popover "Puzzles with tag group:pod-of-dolphins" or something like that instead of just the tag itself)

PR #369, opened 2020-12-11
Suggestion: Full-width metadata

grantaelliott:
This makes the puzzle metadata span the entire top of the puzzle page, rather than having chat span the entire left.

Pros
- Arguably more natural to have the puzzle information start at the left of the screen, directly below navigation
- More space for puzzle title and tags list before wrapping

Cons
- Less vertical space for chat

This is intended as a demonstration of the layout to spur discussion. My preference is weakly for this layout, but I can see both arguments.

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -227,24 +227,21 @@ table.puzzle-list {
   bottom: 0px;
   left: 0px;
   right: 0px;
-
-  .Pane > .puzzle-content, .Pane > .chat-section {
-    position: absolute;
-    top: 0;
-    bottom: 0;
-    left: 0;
-    right: 0;
-  }
-
-  &.narrow {
-    display: flex;
-    flex-direction: column;
-  }
-}
-
-.puzzle-content {
   display: flex;
   flex-direction: column;
+
+  > .SplitPanePlus {
+    flex: 1 1 auto;
+    position: relative;
+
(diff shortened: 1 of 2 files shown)


zarvox:
I'm weakly against this on the grounds that we're giving up chat space (which is usually the most useful non-spreadsheet part of the page) on the desktop layout (which is the most used layout) to...reduce the risk of wrapping in the metadata if we encounter a very long puzzle title or have a ton of tags?  I'd rather have more chat history or call participants visible.

There is definitely a reasonable point here with "have the puzzle information start at the left of the screen, directly below navigation" though.  How would you feel about flipping ChatSection to sit on the right of the same division of Metadata and Document? e.g.

```
 ____________ ___
|  metadata  | c |
|____________| h |
|  document  | a |
|            | t |
|____________|___|
```

zarvox:
Oh, I'm remembering now that one reason I think we kept chat on the left was because the operator queue was overlaid on the right and made it hard to read the chat backlog for operators.  Hmm.

grantaelliott:
Agreed on all points (though I’d add that wrapping in the metadata ends up costing the document some height). I’d be onboard with just flipping chat, but unfortunately I think there’s too much precedent to make flipping notifications a user-friendly idea.

Maybe we just leave this one as is. I’ll leave the PR open for now.

grantaelliott:
Closing this for now. Feel free to reopen if you have an idea.

PR #373, opened 2020-12-16
Add screenshots to README

zarvox:
Hopefully this makes it easier to get a sense of what Jolly Roger is,
looks like, and how you'd use it.


zarvox:
See rendered with new screenshots at https://github.com/deathandmayhem/jolly-roger/blob/f747f3fd18bcb7e78274b73f7612031afb710182/README.md

PR #378, opened 2020-12-24
Fix emoji font usage

zarvox:
The "Platform Emoji" font-family was never getting used, because we were not
specifying valid `src` declarations for local fonts.  If a bare string is
provided, it is interpreted as a URL; we must specify `local()` to imply a
system-local font.

Once we started applying `local()` correctly, I found that we were missing the
ZWJ character needed to display the Jolly Roger pirate flag character: 🏴‍☠️
so I added it to the `unicode-range` for Platform Emoji.

Promoting "Platform Emoji" to lead the Bootstrap theme font families means that
we get reasonable emoji behavior throughout the app -- we'll use the platform
emoji fonts for those unicode ranges, and fall back to Source Sans Pro (or
Source Code Pro for monospace usages), and then try emoji fonts one more time
in case we missed the unicode-range in Platform Emoji and the other fonts don't
provide those glyphs.  This way we get consistent emoji support in puzzle
titles, input boxes, and so on, even if the Source fonts or monospace fonts
provide (poor, black-and-white) glyphs for those codepoints.


grantaelliott (approved):
Thank you for catching this. I need to go back and figure out how I missed it in testing.

PR #379, opened 2020-12-24
Fix tag editor in puzzle creation modal

### LINKED ISSUE #372: In "Add puzzle" dialog, `Tags` input's tag-specific X doesn't remove the tag from the tag list
Specifically, the X highlighted in red:

Screenshot_20201215_012940

The X for the whole input box *does* work, but removes all the tags input, rather than a single one.

zarvox:
https://react-select.com/props explains a bit more about what arguments we get
passed here.  We should treat "no value" as "empty list", and otherwise update
the tags list in our state with whatever the callback says we have now.

Fixes #372.


grantaelliott (approved):
Looks good to me.

PR #383, opened 2021-01-02
Puzzle metadata pane cleanup

grantaelliott:
Cleanup the puzzle metadata pane:

- Remove redundant title text and rearrange remaining content (see below)
- Move icons to the left of link text, for consistency
- Restore button height to the natural btn-sm styles and add an icon to the guess button, making them more obviously buttons. Previously, button height was reduced to fit them comfortably on the same row as text, but given the very flat theme and the similarly styled tags nearby, functionality was potentially ambiguous.
- Move verbose guess count text into a button badge
- Change answer background color to match the desaturated green used for solved puzzles elsewhere, both for consistency and to further distinguish answers from buttons
- Move the tag editor (the add tag interface) to its own line, avoiding the layout oddity of the editor increasing the height of the tag list
- Fix the formatting of tag remove buttons
- Refactor the metadata render method, to improve legibility

metadata

Per conversation with @zarvox, this removes the redundant title from the metadata pane. The remainder of that line then needs to be accommodated elsewhere, but the result is (usually) more compact than before (two lines for unsolved puzzles), freeing a little space for the document. The new layout (consistent across screen widths) is:

1. Puzzle and document links, edit button, guess button (between 0 and 4 of these may be present depending on puzzle type, layout type, and viewer's role)
2. Answers (if solved)
3. Tags

This does represent a reorganization of a familiar layout. My goals were as follows:
- Action buttons and links should stay together, in a consistent location unaffected by line wraps (i.e. either the first row or the last)
- Answers should remain prominent and predictably placed. Long or multiple answers should behave predictably.
- The tag list works well in the bottom row, where the new popover can only occlude the document

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -123,6 +123,11 @@ table.puzzle-list {
 // Tags
 .tag-list {
   display: inline;
+
+  .tag-editor {
+    display: inline;
+    min-width: 200px;
+  }
 }
 
 .tag {
@@ -517,6 +522,11 @@ table.puzzle-list {
   align-items: flex-start;
   align-content: flex-start;
   flex-wrap: wrap;
+
+  .tag-editor {
+    flex-basis: 100%;
+    margin: 2px 0;
+  }
 }
 
 .puzzle-metadata-row .btn {
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -1,5 +1,9 @@
 @import "theme";
 
+$solved-puzzle-background-color: #dfffdf;
+$unsolved-puzzle-background-color: #f0f0f0;
+$administrivia-puzzle-background-color: #dfdfff;
+
 // Puzzle components helper styles
 // Puzzle lists/tables
 .puzzle-group:not(:last-child) {
@@ -20,15 +24,15 @@
 
 .puzzle {
   &.unsolved {
-    background-color: #f0f0f0;
+    background-color: $unsolved-puzzle-background-color;
   }
 
   &.solved {
-    background-color: #dfffdf;
+    background-color: $solved-puzzle-background-color;
   }
 
   &.administrivia {
-    background-color: #dfdfff;
+    background-color: $administrivia-puzzle-background-color;
   }
 }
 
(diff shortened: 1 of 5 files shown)


client/stylesheets/puzzle.scss
```diff
@@ -1,5 +1,9 @@
 @import "theme";
 
+$solved-puzzle-background-color: #dfffdf;
```
zarvox, line 3: love giving names to these and using them consistently!

imports/client/components/PuzzlePage.tsx
```diff
@@ -460,8 +457,65 @@ class PuzzlePageMetadata extends React.Component<PuzzlePageMetadataProps> {
...
+        className="puzzle-metadata-external-link-button"
+        href={this.props.puzzle.url}
+        target="_blank"
+        rel="noreferrer noopener"
+      >
+        <FontAwesomeIcon fixedWidth icon={faPuzzlePiece} />
+        {' '}
+        <span className="link-label">Puzzle</span>
+      </a>
+    );
```
zarvox, line 473: Just for cleanliness: for these conditionally-generated elements, let's either use the ternary operator `(condition ? <element /> : null )` to make the value explicitly `null` or use `!!` to force the value to be `false` rather than accidentally some other false-y value that might get rendered differently.  (I've accidentally used the falsiness of 0 and wound up rendering the string 0 when I meant to render nothing.)

zarvox (approved):
LGTM

Nice job recapturing space.  I'm really pleased with the cleanup we've managed to complete this year.

One thought for future work: should we be trying to make it visibly louder when we've fully-solved a puzzle (e.g. correct answer count === expected answer count)?  Or do you think that it'll be sufficiently clear as-is?

zarvox:
The previous test run ( https://github.com/deathandmayhem/jolly-roger/runs/1635208203?check_suite_focus=true ) appears to have hung after passing the authentication acceptance tests.  Not sure what it was waiting for exactly.

I've kicked it off again; let's see what happens?

zarvox:
Dunno what's up with the `pull_request` builder hanging sometimes.  Seems to get stuck either in teardown of the first test set after all the tests passed, or hanging while trying to do the tests in `./tests/down`.  @ebroder any ideas?

In the meantime, I'm gonna merge this since I'm pretty convinced it's fine (since the `push` builder passed).

PR #384, opened 2021-01-02
Make viewer and caller lists collapsible

grantaelliott:
The call pane can take up a lot of space when many players are viewing a puzzle. As this space is better used for chat, give players the option to collapse the lists of viewers and callers. The default state is expanded caller list and collapsed viewer list.

My original intent was to collapse the two together, permitting even more space savings when fully collapsed, as caller and viewer counts could be displayed on a single line. However, the natural interface for this is the familiar collapsing hierarchy from the puzzle list page and it's a little odd to link collapses in that presentation. Additionally, perhaps the most natural state is to see callers but not passing-through viewers. The independent controls require an extra line in the fully collapsed state, but I think the tradeoff is worthwhile.

As always, open to suggestions.

### DIFF
--- a/imports/client/components/CallSection.tsx
+++ b/imports/client/components/CallSection.tsx
@@ -156,24 +156,22 @@ class RTCCallSection extends React.Component<RTCCallSectionProps> {
             <FontAwesomeIcon fixedWidth icon={callersHeaderIcon} />
             {`${callerCount} caller${callerCount !== 1 ? 's' : ''}`}
           </header>
-          {this.props.callersExpanded && (
-            <div className="people-list">
-              {this.renderSelfBox()}
-              {this.props.signalsReady && this.props.selfParticipant && others.map((p) => {
-                return (
-                  <CallLinkBox
-                    key={p._id}
-                    rtcConfig={this.props.rtcConfig!}
-                    selfParticipant={this.props.selfParticipant!}
-                    peerParticipant={p}
-                    localStream={this.props.localStream}
-                    audioContext={this.props.audioContext}
-                    deafened={this.props.deafened}
-                  />
-                );
-              })}
-            </div>
-          )}
+          <div className={classnames('people-list', { collapsed: !this.props.callersExpanded })}>
+            {this.renderSelfBox()}
+            {this.props.signalsReady && this.props.selfParticipant && others.map((p) => {
+              return (
+                <CallLinkBox
+                  key={p._id}
+                  rtcConfig={this.props.rtcConfig!}
+                  selfParticipant={this.props.selfParticipant!}
+                  peerParticipant={p}
+                  localStream={this.props.localStream}
+                  audioContext={this.props.audioContext}
+                  deafened={this.props.deafened}
+                />
+              );
+            })}
+          </div>
         </div>
       </>
     );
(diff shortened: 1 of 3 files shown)


imports/client/components/CallSection.tsx
```diff
@@ -143,25 +152,28 @@ class RTCCallSection extends React.Component<RTCCallSectionProps> {
...
-                  selfParticipant={this.props.selfParticipant!}
-                  peerParticipant={p}
-                  localStream={this.props.localStream}
-                  audioContext={this.props.audioContext}
-                  deafened={this.props.deafened}
-                />
-              );
-            })}
-          </div>
+          {this.props.callersExpanded && (
```
zarvox, line 159: This is not viable as written.  `CallLinkBox` manages the actual audio call lifecycle and audio playback; we can't make the rendering of `CallLinkBox` conditional on collapse state or collapsing can't happen without disconnecting the audio call (but confusingly, still showing you as wanting to participate).

What would you think about making `people-list` have `overflow: auto` and specifying a `max-height` instead?

zarvox (changes requested):
Didn't look super closely at the `ViewerPersonBox` bits, but we can't make `CallLinkBox` render conditionally or the actual audio calls will be 100% broken on collapse.

imports/client/components/CallSection.tsx
grantaelliott: Thanks for catching that. I did not realize CallLinkBox was functional. I'll fix tonight (probably just using `display:none`).

imports/client/components/ChatPeople.tsx
```diff
@@ -118,6 +124,8 @@ class ChatPeople extends React.Component<ChatPeopleProps, ChatPeopleState> {
       error: '',
       muted: false,
       deafened: false,
+      callersExpanded: true,
+      viewersExpanded: false,
```
zarvox, line 128: Can I convince you to start with viewers expanded as well?

I think it's easier to collapse things if you discover that you want to see more of the chat backscroll than it is to notice the viewer count and ambiently become aware of other people that might be motivated to join a call if you were the first to click the "join audio call" button.

zarvox (approved):
Implementation looks good now.  Thanks for reworking it!

My personal inclination is to leave presence visible by default -- it's easier to discover collapse rather than expand, even at our peak we usually don't have huge numbers of folks on the same puzzle, and I think having clearer ambient presence is something we'll particularly appreciate for social-connectedness reasons this year.  I could probably be convinced either way, but wanted to share my perspective on the options.

If you plan to collapse viewers all the time, and changing the default sounds overly burdensome since you'll now have to collapse every tab you open, one other thing we could consider is making the state slightly sticker -- by making it either a `localStorage` key shared by all tabs and responded to on update (via storage watchers).

grantaelliott:
Seems reasonable. I'll update viewers to default open as well.

As for making changes persistent, I think syncing collapse across open tabs would be an unexpected behavior. I'd prefer supporting a per-user default in profile. Thoughts?

zarvox:
> As for making changes persistent, I think syncing collapse across open tabs would be an unexpected behavior. I'd prefer supporting a per-user default in profile. Thoughts?

That'd work for me!

grantaelliott:
Sounds good. I suggest we merge this now (as soon CI finishes...) and I will work on personalization features soon.

PR #386, opened 2021-01-03
Cleanup future handling in lock implementation

ebroder:
Any time that an event fires that will resolve the future, we want to make sure to clear all other watches so it doesn't resolve multiple times.

I think this should fix the error that @zarvox reported:

```
I20210102-01:38:47.423(-8)? Exception in removedobserveChanges callback: Error: Future resolved more than once
I20210102-01:38:47.423(-8)?     at Future.return (/home/zarvox/.meteor/packages/meteor-tool/.1.12.0.4zjno.40411w++os.linux.x86_64+web.browser+web.browser.legacy+web.cordova/mt-os.linux.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:261:10)
I20210102-01:38:47.423(-8)?     at removed (imports/server/models/lock.ts:58:13)
I20210102-01:38:47.424(-8)?     at runWithEnvironment (packages/meteor.js:1286:24)
I20210102-01:38:47.424(-8)?     at packages/meteor.js:1299:14
I20210102-01:38:47.424(-8)?     at packages/mongo/observe_multiplex.js:178:30
I20210102-01:38:47.424(-8)?     at Array.forEach (<anonymous>)
I20210102-01:38:47.425(-8)?     at Function._.each._.forEach (packages/underscore.js:139:11)
I20210102-01:38:47.425(-8)?     at Object.task (packages/mongo/observe_multiplex.js:172:9)
I20210102-01:38:47.426(-8)?     at Meteor._SynchronousQueue.SQp._run (packages/meteor.js:917:16)
I20210102-01:38:47.426(-8)?     at packages/meteor.js:894:12
```

### DIFF
--- a/imports/server/models/lock.ts
+++ b/imports/server/models/lock.ts
@@ -5,7 +5,7 @@ import Ansible from '../../ansible';
 import LockSchema, { LockType } from '../schemas/lock';
 
 declare const Npm: any;
-const Future = Npm.require('fibers/future');
+const Future: typeof import('fibers/future') = Npm.require('fibers/future');
 
 // 10 seconds
 export const PREEMPT_TIMEOUT = 10000;
@@ -23,7 +23,7 @@ const Locks = new class extends Mongo.Collection<LockType> {
       return this.insert(<any>{ name });
     } catch (e) {
       if ((e.name === 'MongoError' || e.name === 'BulkWriteError') && e.code === 11000) {
-        return null;
+        return undefined;
       }
 
       throw e;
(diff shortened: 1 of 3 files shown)


zarvox:
If I make a couple of edits in a row to a server file, I observe multiple lock preemptions, followed by a steady-state where there is no lock present in the DB at all (!).  That seems bad, and suggests that somehow we're winding up with multiple attempts by the same process to preempt locks?

E.g.

* add comment line, `:w`
* remove comment line, `:w`
* observe:

```
                                              
Unable to resolve some modules:

  "fibers/future" in /home/zarvox/git/jolly-roger/imports/server/models/lock.ts (os.linux.x86_64)
                                              
If you notice problems related to these missing modules, consider running:
                                              
  meteor npm install --save fibers            
                                              
                                              
Unable to resolve some modules:

  "fibers/future" in /home/zarvox/git/jolly-roger/imports/server/models/lock.ts (os.linux.x86_64)
                                              
If you notice problems related to these missing modules, consider running:
                                              
  meteor npm install --save fibers            
                                              
I20210102-17:32:12.741(-8) (migrations_server.js:80) Migrations: Not migrating, already at version 27
=> Meteor server restarted
I20210102-17:32:17.598(-8)? Prempting lock: id=ENemuuKi6ac3muAGt name=discord-bot
I20210102-17:32:17.601(-8)? Prempting lock: id=ENemuuKi6ac3muAGt name=discord-bot

```

and then from `meteor mongo`:

```
meteor:PRIMARY> db.jr_locks.find({})
meteor:PRIMARY> 
```

This seems like an invalid final state -- my understanding is that I should expect the state of the DB to converge to a lock regularly being refreshed.  My guess is that we're not successfully deduping calls to `Locks.withLock` within the same process.

I don't think we can just `.defer()` that future; we need to avoid generating another if we already have done so once.

ebroder:
I think I got a little overly clever with the import, but otherwise I still think this fix is correct for solving the issue you reported where the Future got resolved multiple times. For the circumstance where there was no lock in the DB, I think that's because there was no process in the critical section, which I believe is fixed by #387

zarvox:
Ahhh, that makes more sense.

PR #392, opened 2021-01-03
Cache Discord API objects from the gateway

ebroder:
Set up a mongo cache that is updated whenever the Discord gateway client gets
notifications of object updates. In theory, we could use this to replace our own
API fetches of state.

### DIFF
--- a/imports/lib/schemas/discord_cache.ts
+++ b/imports/lib/schemas/discord_cache.ts
@@ -7,6 +7,7 @@ import { buildSchema, Overrides } from './typedSchemas';
 export const DiscordCacheCodec = t.type({
   _id: t.string,
   createdAt: date,
+  updatedAt: t.union([date, t.undefined]),
   snowflake: t.string,
   type: t.string,
   object: t.object,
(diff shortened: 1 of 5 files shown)


ebroder:
(I'm planning to switch over the existing code to use this cache in a separate PR, but I think this one is ready to go)

imports/lib/schemas/discord_cache.ts
```diff
@@ -0,0 +1,38 @@
...
+export const DiscordCacheCodec = t.type({
+  _id: t.string,
+  createdAt: date,
+  snowflake: t.string,
+  type: t.string,
+  object: t.object,
+});
+
+const DiscordCacheOverrides: Overrides<t.TypeOf<typeof DiscordCacheCodec>> = {
+  createdAt: {
```
zarvox, line 16: Any reason to omit (or include) an `updatedAt`?

ebroder: No strong reason, I can add one.

zarvox (approved):
LGTM

Looks like the roles cache is a little tricky, because roles are scoped to the guild that owns them, but presumably they contain some sort of back-reference?

ebroder:
> roles are scoped to the guild that owns them, but presumably they contain some sort of back-reference?

They have a `guild` property, if that's what you mean?

PR #396, opened 2021-01-05
Add an audio-levels self-test with some guidance

zarvox:
This change adds an audio self-test section to OwnProfilePage to let you test your microphone setup locally, and provides more detail about the absolute values of the spectrogram in dB.

It also allows the user to wire up loopback audio playback for testing purposes, though a nicer thing to do (if we have the time later to revisit) would be to have some sort of record/playback functionality.

I would particularly appreciate feedback on the copy; I'm not really sure about if my guidance around levels is correct.  I can also adjust the absolute scale of the spectrogram if desired.

Screenshot_20210104_141013


ebroder:
FWIW I would be mildly in favor of wiring in a 1-2s delay on the loopback. I found that very helpful when I was playing with settings.

ebroder (approved):

grantaelliott (approved):
Since the microphone stream is requested with autoGainControl, the guidelines will be to some extent self-fulfilling, which may lead users to address the wrong problem. That said, I think it's the right choice to test with the same settings used in chat, but we should be aware of the caveats. In particular, this makes it tricky to offer guidance for adjusting speaking volume.

Unfortunately, I'm concerned that the real-time FFT is hard to interpret actionably, especially with so few actions a typical user with a laptop mic can take. In my experience, the voice level was in the "good" range no matter how far I got from the mic and it was difficult to identify what represented background noise. I suggest we primarily consider the FFT just a visual confirmation that the microphone is working. I think guidance text should be cast in terms of suggested resolutions, so perhaps it reduces to "When you aren't speaking, all bars should be below -80dBFS. If they aren't, consider moving to a quieter location." (What is the origin of these thresholds, by the way?)

Loopback audio is more intuitively interpretable (and actionable) and it's unfortunate that feedback bars most users from this test. Record and playback (instead of real-time loopback) would have wider applicability, though that's likely too much work to be worthwhile.

Ultimately, I think the most important purpose of this test is verifying that the microphone is successfully being accessed by the web app. I think this serves that purpose as-is. My opinion is that the FFT and specific guidance about voice levels can be downplayed, but that's up to you.

(Also, it's kind of a pedantic point, but it's strange to quote an absolute dB level without a stated reference. I think the units on this are dBFS and it'd arguably be more clear to use that instead of dB at least once.)

zarvox:
I found a couple-second delay still produced feedback, just synchronized to the delay interval.  I don't care enough about this to do record/replay right now, but would be +1 on someone doing it.  Loopback is also complicated by the fact that most users control device volume, and can turn it up or down as desired.  But I guess most people will usually have that volume at a "comfortable" level.

In that sense, the real goal is to make all peers have similar spectrograms (which automatic gain control ~solves for us), and let the user modulate overall volume to a comfortable level.

> (What is the origin of these thresholds, by the way?)

My local testing speaking comfortably at my usual volume levels:

* in Chrome using a Kinect from a couple feet away
* in Chrome using a second desk mic
* in Safari with the MacBook Air internal mic
* in Firefox with the same internal mic

They were all in about the recommended ranges, so I figure this will be relatively reproducible with whatever setup people have on hand.  I'll be surprised if the browsers' built-in AGC doesn't make ~everyone's setup fall within these guidelines.

Thanks for the reviews!

PR #404, opened 2021-01-12
Workaround react-bootstrap autoFocus bug by disabling animations

### LINKED ISSUE #359: Guess submission form should autofocus the guess input when shown

zarvox:
See https://github.com/react-bootstrap/react-bootstrap/issues/5102

This (`animation={false}`) is the simple workaround for the react-bootstrap
issue linked above.  We could do some slightly more involved workarounds to
preserve the animation, but they'd break some abstractions (have to have a ref
of the child to focus, which limits the code's reusability) and honestly I find
the site feels slightly snappier without them.

So we just turn off `Modal` animations entirely.

Fixes #359.

### DIFF
--- a/imports/client/components/ModalForm.tsx
+++ b/imports/client/components/ModalForm.tsx
@@ -52,8 +52,20 @@ class ModalForm extends React.Component<ModalFormProps, ModalFormState> {
   };
 
   render() {
+    // Note: the animation=false is working around a regression in
+    // react-bootstrap that otherwise breaks autofocus:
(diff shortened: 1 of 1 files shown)


grantaelliott (approved):
Looks good. Do we want to open an issue reminding ourselves to reenable animations once react-bootstrap fixes this?

zarvox:
Filed https://github.com/deathandmayhem/jolly-roger/issues/406 to track reenabling if upstream fixes things.

PR #405, opened 2021-01-12
Polish puzzle list page

grantaelliott:
- Multiline answer indents didn't render correctly in Safari, as ch does not size as expected with font fallback. Since we know the font, express in terms of em instead.
- Adjust puzzle group margin to align with header
- Fix inconsistent margins in puzzle list page controls
- Make puzzle list page control markup consistent
- Prefer standard to outline style buttons in radio and toggle bottoms, as the latter use the same colors for hover and active
- Use icon for clear button

### DIFF
--- a/imports/client/components/PuzzleListPage.tsx
+++ b/imports/client/components/PuzzleListPage.tsx
@@ -7,6 +7,7 @@ import {
   faMap,
   faReceipt,
   faUsers,
+  faEraser,
 } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import React from 'react';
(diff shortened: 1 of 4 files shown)


imports/client/components/PuzzleListPage.tsx
```diff
@@ -341,55 +342,57 @@ class PuzzleListView extends React.Component<PuzzleListViewProps, PuzzleListView
...
+                  id="jr-puzzle-search"
+                  as="input"
+                  type="text"
+                  ref={this.searchBarRef}
+                  placeholder="Filter by title, answer, or tag"
+                  value={this.getSearchString()}
+                  onChange={this.onSearchStringChange}
+                />
+                <InputGroup.Append>
+                  <Button variant="danger" onClick={this.clearSearch}>
```
zarvox, line 385: Can we swap this to `light` as well?  In red, its loudness on the page greatly exceeds its importance.

zarvox: Or at least `secondary`.

zarvox (approved):
Make the clear button not-red and I call this good :)

zarvox:
LGTM

PR #409, opened 2021-01-13
Make initial page load faster for operators

ebroder:
Historically, in order to display pending guesses, the operator view has loaded all hunts and puzzles. If a Jolly Roger instance has a lot of history, that can be quite expensive. Instead, expose a special subscription that includes pending guesses + the specific hunts/puzzles that they're linked to. This requires more state tracking on the server, but since there are relatively few operators it seems worth that cost.

(FWIW, this could be made even more efficient if we projected puzzles and hunts for just the fields that we care about, but that creates some risk to type safety that I don't really want to touch. It's also possible we could generalize some of this logic for other use cases, e.g. pending announcements + announcements, although we couldn't lean on the assumption that it will only be used by a few people)

Since this is a bit speculative and not a blocker for this weekend, I'm not planning to merge this without a code review.

### DIFF
--- a/imports/server/guesses.ts
+++ b/imports/server/guesses.ts
@@ -1,11 +1,16 @@
 import { check } from 'meteor/check';
-import { Meteor } from 'meteor/meteor';
+import { Meteor, Subscription } from 'meteor/meteor';
+import { Mongo } from 'meteor/mongo';
 import { Roles } from 'meteor/nicolaslopezj:roles';
 import Ansible from '../ansible';
+import Base from '../lib/models/base';
 import Guesses from '../lib/models/guess';
 import Hunts from '../lib/models/hunts';
 import Puzzles from '../lib/models/puzzles';
+import { BaseType } from '../lib/schemas/base';
 import { GuessType } from '../lib/schemas/guess';
+import { HuntType } from '../lib/schemas/hunts';
+import { PuzzleType } from '../lib/schemas/puzzles';
 import { sendChatMessage } from './chat';
 import GlobalHooks from './global-hooks';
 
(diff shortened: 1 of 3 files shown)


imports/server/guesses.ts
```diff
@@ -57,6 +62,167 @@ function transitionGuess(guess: GuessType, newState: GuessType['state']) {
   }
 }
 
+// It's easier to create one observer per object, rather than try and have a
+// single observer with an $in query. It means we don't have to start/stop the
+// observer. Identical queries from different users will get deduped on the
```
zarvox, line 67: > Identical queries from different users will get deduped on the server side

Where does that happen, out of curiosity?  I'm not seeing anything obvious in this code, so I'm wondering if there's some fanciness in Meteor that I'm not aware of.

imports/server/guesses.ts
```diff
@@ -57,6 +62,167 @@ function transitionGuess(guess: GuessType, newState: GuessType['state']) {
...
+        // The order of operations here is important to avoid transient
+        // inconsistencies
+        if (huntUpdated) {
+          this.huntRefCounter.incref(fields.hunt!);
+        }
+        if (puzzleUpdated) {
+          this.puzzleRefCounter.incref(fields.puzzle!);
+        }
+        this.sub.changed(Guesses.tableName, id, fields);
+        if (huntUpdated) {
```
zarvox, line 189: I'd have expected that you'd unref in the reverse order that you ref -- so I'd expect to decrement puzzle refs before hunt refs, so there's never a moment the client has a `Puzzle` referencing a `Hunt` that it no longer has.  Is my intuition off here somehow?

imports/server/guesses.ts
```diff
@@ -57,6 +62,167 @@ function transitionGuess(guess: GuessType, newState: GuessType['state']) {
...
+          this.huntRefCounter.decref(this.guesses[id].hunt);
+        }
+        if (puzzleUpdated) {
+          this.puzzleRefCounter.decref(this.guesses[id].puzzle);
+        }
```
zarvox, line 194: I think after the decrefs are done, you should update your saved values in `this.guesses[id]` with the new field values?  Otherwise if there's another change, you'll be dropping refs on the original values, not the most-recent-predecessor values.  (Not like puzzle or hunt should ever be changing on guesses, but on principle...)

imports/server/guesses.ts
```diff
@@ -57,6 +62,167 @@ function transitionGuess(guess: GuessType, newState: GuessType['state']) {
...
+  guesses: Record<string, GuessType>
+
+  huntRefCounter: RefCountedObserverMap<HuntType>;
+
+  puzzleRefCounter: RefCountedObserverMap<PuzzleType>
+
+  constructor(sub: Subscription) {
+    this.sub = sub;
+
+    this.guessCursor = Guesses.find({ state: 'pending' });
```
zarvox, line 163: Shouldn't at least this query be filtering down to hunts that the operator is a member of?  Normally that visibility is restricted on the publish side, via `<T extends BaseType>.publish(huntsMatchingCurrentUser)`, but here I think we've lost that filtering.

zarvox (commented):
This seems really great, and likely to make our lives easier.  While in practice I think we could get away with this as-is, there's a few changes (comments inline) that I think would make this slightly more resilient/correct.

If you happen to have any rough numbers on if this has any noticeable impact on CPU or RAM usage, that'd be cool, but also I highly doubt this will be expensive -- most of the time, there are no guesses in the pending state, and usually there's only a tiny handful, so it's not like there's likely to be a ton of child reactive queries.  Probably cheaper than observing all puzzles and all hunts!

imports/server/guesses.ts
ebroder: Meteor will dedup identical cursors - happens somewhere around here: https://github.com/meteor/meteor/blob/3cd6e5ebc66216d07e15ef9cc2f326621934e828/packages/mongo/mongo_driver.js#L1295-L1319

imports/server/guesses.ts
ebroder: Yeah, that's right. I was mostly thinking about hunt+puzzle in relation to guess, but not in relation to each other.

imports/server/guesses.ts
ebroder: Ah yeah, I had this line but lost it in a refactor. Pushed a fix.

imports/server/guesses.ts
ebroder: (pushed a fix)

imports/server/guesses.ts
ebroder: Good catch, pushed a fix

ebroder:
> If you happen to have any rough numbers on if this has any noticeable impact on CPU or RAM usage

I don't have any good measurements here, but it seems hard to see how this could be worse. In steady state, we're only subscribing to guesses (which we were doing before). We only create watches on puzzles/hunts when there are open guesses. The puzzle watch in particular is O(n) on oplog entries where it previously was O(1), but n is going to be quite small, so I think it's mostly a question of whether it's more expensive to watch and skip an oplog entry or more expensive to actually push an oplog update to clients. (My money is on the former)

zarvox (approved):
Thanks for the fixes!  Looks like our two PRs conflicted, so you'll need to rebase, but otherwise, LGTM.

Issue #337, opened 2020-11-25
Add chat notifications based on username mentions


grantaelliott:
In the past, the Slack firehose channel allowed you to receive a notification when you were mentioned by username. Discord's ability to set custom alerting logic is somewhat more limited, but even if it weren't, it would be nice to offer this within Jolly Roger itself.

Specifically, it would be good if you could receive an alert that your name has been called out in chat (likely using @). Since we don't have usernames, this would probably require a dropdown to search for the user (similar to the UI used for tag entry). It may also make sense to add a handle field to user profiles, since we have many users as well known by their handles as by their names.

zarvox:
This is a solid feature spec, thanks for filing!

I think this will also need some nontrivial serverside support, since I think it'd be unsupportable load-wise to have every client watching every puzzles chatmessages all the time (and that would blow per-tab RAM usage through the ceiling).

If we do wind up doing this, we probably want some plan around load-testing or deferred work queue, because I can see the fanout of checking every user's pingwords against every chat message sent adding up to a fair bit of CPU/DB load.

This probably also *ought* to take ACLs into consideration?  Don't want people getting notifications about puzzles in hunts they don't belong to.

Were you anticipating these being the in-tab notifications that we have for general announcements/guess queue for operators, or desktop notifications?

grantaelliott:
I'm assuming these would be the same in-page notifications we have for announcements and operator requests. Supporting desktop notifications as a fallback (when not connected) would be very nice, with the big caveat that iOS still doesn't have support.

ebroder:
I'm going to rework this ticket to reflect support for configurable keywords. I think there's still an open issue around username notification. Here's the original text of the issue from before my edits:

> In the past, it's been possible for users to set up notifications in Slack's firehose channel for keywords of interest to them or their own names. This may be possible in Discord as well, but even if so, it would be nice to offer this within Jolly Roger itself. The existence of this feature would also eliminate one of the arguments in favor of maintaining a firehose channel (or even of mirroring chat entirely).
> 
> I propose two types of such notifications, each delivered with a link back the corresponding Jolly Roger page.
> 
> * Alert that your name has been called out in chat (likely using @). Since we don't have usernames, this would probably require a dropdown to search for the user (similar to the UI used for tag entry). It may also make sense to add a handle field to user profiles, since we have many users as well known by their handles as by their names.
[...]

PR #418, opened 2021-01-15
Add experimental dingwords and in-app notifications

zarvox:
The simplest thing that could work, labeled as experimental, and guarded by a feature flag in the event the naive approach turns out to generate way too much server load.

We add:

* a Profile field `dingwords`, with a list of strings
* a way to edit it in OwnProfilePage
* a circuit breaker `disable.dingwords` on the setup page
* a `DingwordHookset` to, for each chat message sent, check if any users want to get notified, and generate notifications accordingly
* a reactive smart subscription `chatNotifications` which ensures that the puzzle and hunt are available clientside if a chat notification references them
* a new NotificationCenter notification type which links to the puzzle, and shows the sender and the message, and allows the notification to be dismissed

No desktop notifications (permissions and getting them to be interactive from the right tab if multiple tabs are open is more complicated than I want to deal with right now).  No sounds either -- we're likely to have lots of audio usage during hunt anyway, and the less we generate annoying noise the better off we'll be.

Relates to #337 (but only covers basic text, not @-mentions)

### DIFF
--- a/imports/client/components/NotificationCenter.tsx
+++ b/imports/client/components/NotificationCenter.tsx
@@ -16,12 +16,14 @@ import CopyToClipboard from 'react-copy-to-clipboard';
 import { Link } from 'react-router-dom';
 import Flags from '../../flags';
 import Announcements from '../../lib/models/announcements';
+import ChatNotifications from '../../lib/models/chat_notifications';
 import Guesses from '../../lib/models/guess';
 import Hunts from '../../lib/models/hunts';
 import PendingAnnouncements from '../../lib/models/pending_announcements';
 import Profiles from '../../lib/models/profiles';
 import Puzzles from '../../lib/models/puzzles';
 import { AnnouncementType } from '../../lib/schemas/announcements';
+import { ChatNotificationType } from '../../lib/schemas/chat_notifications';
 import { GuessType } from '../../lib/schemas/guess';
 import { HuntType } from '../../lib/schemas/hunts';
 import { PendingAnnouncementType } from '../../lib/schemas/pending_announcements';
@@ -320,6 +322,32 @@ function ProfileMissingMessage(props: ProfileMissingMessageProps) {
   );
 }
 
+interface ChatNotificationMessageProps {
+  // TODO: add the requisite fields
+  cn: NotificationCenterChatNotification;
+  onDismiss: () => void;
+}
+function ChatNotificationMessage(props: ChatNotificationMessageProps) {
+  return (
+    <li>
+      <MessengerSpinner />
+      <MessengerContent dismissable>
+        <Link to={`/hunts/${props.cn.hunt._id}/puzzles/${props.cn.puzzle._id}`}>
+          {props.cn.puzzle.title}
+        </Link>
+        <div>
+          {props.cn.senderDisplayName}
+          {': '}
+          <div>
+            {props.cn.cn.text}
+          </div>
+        </div>
+      </MessengerContent>
+      <MessengerDismissButton onDismiss={props.onDismiss} />
+    </li>
+  );
+}
+
 interface NotificationCenterAnnouncement {
   pa: PendingAnnouncementType;
   announcement: AnnouncementType;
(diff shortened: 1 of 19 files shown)


imports/lib/schemas/chat_notifications.ts
```diff
@@ -0,0 +1,51 @@
...
+  // The sender of the chat message that tripped the user's dingwords.
+  // System messages are forbidden from triggering dingword notifications.
+  sender: t.string,
+
+  // The puzzle to which this chat was sent.
+  puzzle: t.string,
+  // The hunt in which the puzzle resides.
+  hunt: t.string,
+  // The message body. Plain text.
+  text: t.string,
```
ebroder, line 21: I had kind of assumed we'd record a foreign key to the chat message but I guess this makes it easier to track/deal with

imports/server/chat-notifications.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+Meteor.publish('chatNotifications', function () {
+  if (!this.userId) {
+    throw new Meteor.Error(401, 'Not logged in');
+  }
+
+  if (Flags.active('disable.dingwords')) {
+    return;
+  }
```
ebroder, line 87: publish functions aren't reactive, right, so this won't re-render if the flag changes?

ebroder: I guess that's presumably harmless since we won't be generating new chat notifications, and the client would resubscribe based on flag changes

imports/server/hooks/dingword-hooks.ts
```diff
@@ -0,0 +1,78 @@
...
+    }, {
+      fields: { _id: 1 },
+    }).fetch().map((u) => {
+      return u._id;
+    });
+
+    // Fetch the profiles of all those users.
+    const memberProfiles = Profiles.find({
+      _id: {
+        $in: huntMembers,
```
ebroder, line 35: nit: might be nice to throw `'dingwords.0': {$exists: true}` as a query to capture "has any dingwords" (and you could index on that too)

zarvox: Ahhhhh, nice!  I tried something like that with `$ne: []` but tsc didn't like the types.

I was given to understand that most queries having to do with array fields were not indexable?  but maybe `.0` is special?

imports/server/chat-notifications.ts
zarvox: Yeah, the latter is more broadly what saves us here.

imports/server/hooks/dingword-hooks.ts
ebroder: Arrays are definitely indexable. If you index an array field, then it explodes out the array and creates an index entry for each value (which is a perf nightmare if your arrays get big, but not a risk here). But indexing a single array element shouldn't explode or do anything unusual.

imports/lib/schemas/chat_notifications.ts
zarvox: Yeah, it made slightly more sense to me to denorm it?  I guess you could always make the reactive subscribe do even more chaining, but given that the actions you can take are limited to "view the puzzle", it's not *that* useful to have a reference to a specific message.

imports/client/components/NotificationCenter.tsx
```diff
@@ -320,6 +322,32 @@ function ProfileMissingMessage(props: ProfileMissingMessageProps) {
...
+  // TODO: add the requisite fields
+  cn: NotificationCenterChatNotification;
+  onDismiss: () => void;
+}
+function ChatNotificationMessage(props: ChatNotificationMessageProps) {
+  return (
+    <li>
+      <MessengerSpinner />
+      <MessengerContent dismissable>
+        <Link to={`/hunts/${props.cn.hunt._id}/puzzles/${props.cn.puzzle._id}`}>
```
ebroder, line 335: FTR, this could just be `/hunts/${props.cn.puzzle.hunt}/puzzles/${props.cn.puzzle._id}` and we wouldn't actually need the hunt sub at all, right?

(I think that's not true of pending guesses, since we need the hunt's config for URL transformation)

imports/server/hooks/dingword-hooks.ts
zarvox: Another option would be to recompute a per-hunt dingwords cache each time a user saves their profile, shifting the load from "every chat message" to "every dingword configuration change" (and I do anticipate the latter being much less frequent an operation).  We'd have to do some locking to avoid dropping dingword updates if multiple profile saves are in flight at once though.

ebroder:
Or just set up a singleton in-memory subscription on Profiles

imports/client/components/NotificationCenter.tsx
zarvox: I think you're right.  I'm just slightly paranoid about ever making a foreign key available if the target isn't already present client-side, since 1) we seem to be pretty aggressive about assuming that we can `findOne()` them, and 2) many, many of our bugs resulting in totally blank layout in old hunts were the result of attempting to traverse `undefined`s. 😛

PR #431, opened 2021-03-04
Auto-nest groups in group view on PuzzleListPage

### LINKED ISSUE #426: PuzzleListPage: consider nesting groups
One thing that comes up is that rounds with multiple metas wind up with the puzzle showing up in multiple groups, and so sub-groups where all the puzzles in the sub-group exist within a parent group get listed in a way that duplicates the whole child group, if it was entirely contained by the parent group.

In this case, it might be neater for presentation purposes to do some slightly cleverer topological sorting of groups, and if puzzles in group B are a strict subset of puzzles in group A, rendering it nested under group A instead of rendering all the puzzles in that group under group A and again under (separately) group B.

This would help a bit making things map more closely to the natural underlying hierarchy when one exists, while keeping the possibility of non-nested groups getting rendered exactly as they do today.

zarvox:
This change adds logic that recognizes when `group:` tags strictly
dominate each other by puzzles contained, and intelligently nests
them recursively when one group contains another.

Some corner cases worth noting: if there are two groups which have
exactly the same set of puzzles, we cannot tell which is the "outer" or
"inner" group, so we show both groups in full.  (Adding a puzzle to one
but not the other will disambiguate, which usually happens once you've
unlocked enough of a round.)  Similarly, if a group is a strict subset
of several other disjoint groups, we'll show it in each.

To make search behave predictably and preserve nesting group context, we
switch filtering to take place *after* grouping and sorting.  This way,
the user sees only the puzzles that match their search query, but still
under the multiple groups under which they appear without a filter
applied.

Since we now have situations where we have multiple shared tags in
context, we now pass lists of tag IDs to suppress in various views
rather than just a boolean which implies only suppressing the (single)
`sharedTag` value.

With additional tag suppressions, the primary PuzzleListPage view grows
much less noisy, since most tags are groups, and most group tags are
suppressed because most structures are nesting subgroups.

Fixes #426.


zarvox:
A few example screenshots for various behaviors:

Structure for opening round of Mystery Hunt 2018:

Screenshot_20210304_154315

Search preserves group structure:

Screenshot_20210304_155012

Nesting involving deep hierarchy or common subgroups but different contents:

Screenshot_20210304_154401

zarvox:
Did some basic perf optimizations (more Set usage, early exits, caching recomputed data) to bring the overhead of this new approach down from ~8msec to about 1msec on the 2015 and partial-2018 test datasets.  Almost all of render time is spent in either Tracker (pulling records from minimongo) or in React (flushing to real DOM).

imports/client/components/puzzle-sort-and-group.ts
```diff
@@ -0,0 +1,350 @@
...
+  // At the point we're calling this, we ignore the `subgroups` field, which are not yet
+  // populated, and consider only puzzles for determining subgroupiness.
+
+  if (subCand.puzzles.length >= parentCand.puzzles.length) {
+    // If the subgroup is the same size as or larger than the parent candidate,
+    // it can't be contained by the parent candidate group (second criterion above)
+    return false;
+  }
+
+  // first criterion: does every puzzle id in subCand appear in parentCand?
```
ebroder, line 157: Should there be a second criterion?

ebroder: Ah I see they're just out of order

ebroder (approved):

PR #438, opened 2021-03-12
Save last "Show solved" state per Hunt in localStorage

### LINKED ISSUE #423: Make `show solved` toggle sticky within the context of a hunt.
Later on in the hunt, oftentimes there are many solved puzzles and not so many unsolved ones, but the unsolved ones are the most interesting.  However, any time you navigate away from the puzzle list page, the state (show only unsolved puzzles/groups) gets reset.

While we wouldn't want to preserve this state across hunts (we probably don't want to hide information by default, especially if it's been ~a year since the user last touched the app/settings!), within the context of a hunt, making this sticky might make sense.

This seems like the sort of thing I'd encourage keeping in LocalStorage rather than making more serverside state.

zarvox:
And use that value as the default value on PuzzleListPage for that Hunt.
This has the effect of making the toggle "sticky", which makes the "hide
solved" view more useful, especially later on in hunts when there are
many solved puzzles and we're trying to see the ones that remain more
clearly.

Now that we have UI that indicates that we're hiding some subset of
puzzles when applying either a search or an unsolved filter, I'm less
concerned about folks accidentally hiding solved puzzles and being
confused when they can't find the puzzle they're looking for, so I'm
more comfortable with the idea of hiding solved puzzles more of the
time.

Fixes #423.

### DIFF
--- a/imports/client/components/PuzzleListPage.tsx
+++ b/imports/client/components/PuzzleListPage.tsx
@@ -46,6 +46,10 @@ interface PuzzleListViewState {
   showSolved: boolean;
 }
 
+function showSolvedStorageKey(huntId: string): string {
+  return `showsolved-${huntId}`;
+}
+
 class PuzzleListView extends React.Component<PuzzleListViewProps, PuzzleListViewState> {
   addModalRef: React.RefObject<PuzzleModalForm>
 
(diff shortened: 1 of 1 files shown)


zarvox:
The approach taken here allows separate tabs to have separate in-memory values, but on each change all future tabs will take the last-written value.  If we wanted to make this bit sync more immediately between tabs, we could always add a storage listener to update `state` whenever we observe a new write to that key in `localStorage` like we do in AudioConfig.tsx.

Thoughts?

ebroder (approved):
Yeah I agree with not having this take effect on open tabs immediately

Issue #193, opened 2019-01-13
Make more things whitelabelable


zarvox:
There's several places where we've baked "Death and Mayhem" into the source pretty directly:

* https://github.com/deathandmayhem/jolly-roger/blob/master/imports/server/hunts.js#L19-L34
* https://github.com/deathandmayhem/jolly-roger/blob/master/imports/server/accounts.js#L96-L118
* https://github.com/deathandmayhem/jolly-roger/blob/master/imports/client/components/AccountForm.jsx#L169
* https://github.com/deathandmayhem/jolly-roger/blob/master/imports/server/gdrive.js#L89

It would be nice for other teams to be able to brand their own instances with their own team name, and perhaps their own logo.

PR #441, opened 2021-03-21
Make email configurable via a new Setting with templates

zarvox:
This change:

* adds a new Setting type for email configuration which covers all the
  deployment-specific configuration that was previously hardcoded in the
  jolly-roger source
* adds a Meteor method in setup.ts to save this configuration to the DB
* adds UI in SetupPage to edit the From: address and email templates,
  along with an explanation of what variables are available at
  email-rendering time, and to call the method to save the config values
* adds code in imports/server/accounts.ts to monitor that setting for changes and update
  `Accounts.emailTemplates` accordingly when changes are observed, so that the Meteor
  accounts system will use the templates and `From:` as desired.
* adds code to render Mustache templates to produce email contents when
  we want to send email

Related to #193.

Once we merge this, we should update the configuration on our production instance to explicitly specify appropriate values for each of these fields.


ebroder (approved):

zarvox:
Closing the loop: I updated our production deployment's email settings.

PR #442, opened 2021-03-26
Configurable server-served images

zarvox:
(I recommend reviewing this commit-by-commit.)

One of the remaining tasks to make the upstream jolly-roger implementation rebrandable without requiring source modification is to support deployment-specific image assets.  This series of patches make that possible.

* We must be able to serve images based on some site-specific configuration.
* We want to avoid code changes being necessary to achieve custom branding
* Mongo is the most straightforward mutable storage we have
* None of our assets are even close to the maximum allowable mongo document size

So, it makes the most sense to stuff assets in Mongo.

Arguably, the most sensible representation for this data in Mongo would be an EJSON binary, but since our schemas system doesn't appear to have a way to cleanly describe UInt8Array, the easiest thing wound up being to just base64-encode the binary and store it as a string.  If you have better ideas here, go for it.

So we store blobs in mongo, base64'd.

We need to be able to upload these, and honestly I don't want to figure out another authentication scheme, so instead I made a collection UploadTokens which contains single-use tokens for uploading assets.  You make a Meteor method call which checks authorization and returns a token that you can use for a single POST request.

To preserve backwards compatibility (and provide some default branding), we still include the existing assets, but we move them from `public/` (which serves all assets from the web root) to `private/` which makes them available to serverside code via Meteor's `Assets` API.  From there, we serve in priority order:

* The blob from Mongo with a matching asset name
* The binary from the `Assets` populated by the contents of `private/` via Meteor's build system
* A 404.

Then, we just need some code on `SetupPage` to show the current assets, and provide a way to upload new ones.

Screenshot_20210326_020740

Screenshot_20210326_020805

Addresses: #193.


imports/client/components/SetupPage.tsx
```diff
@@ -1521,6 +1521,159 @@ class WebRTCSection extends React.Component<WebRTCSectionProps> {
...
+            Brand icon, 50x50 pixels, shown in the top left of all logged-in pages
+          </BrandingAssetRow>
+          <BrandingAssetRow asset="brand@2x.png">
+            Brand icon @ 2x res for high-DPI displays, 100x100 pixels, shown in the top left of all logged-in pages.
+          </BrandingAssetRow>
+          <BrandingAssetRow asset="hero.png" backgroundSize="contain">
+            Hero image, approximately 510x297 pixels, shown on the login/enroll/password-reset pages.
+          </BrandingAssetRow>
+          <BrandingAssetRow asset="hero@2x.png" backgroundSize="contain">
+            Hero image, approximately 510x297 pixels, shown on the login/enroll/password-reset pages.
```
zarvox, line 1655: Oh whoops I copied this description and didn't update it.  I'll fix that

imports/server/images.ts
```diff
@@ -0,0 +1,111 @@
...
+});
+
+const UPLOAD_TOKEN_VALIDITY_MSEC = 60000; // 60 seconds
+router.post('/:uploadToken', (req, res) => {
+  check(req.params.uploadToken, String);
+  // Look up upload token.  If missing, or too old (>5m), reject with a 403.
+  const uploadToken = UploadTokens.findOne(req.params.uploadToken);
+
+  // Regardless of age, once a token is presented, we should remove it.
+  UploadTokens.remove(req.params.uploadToken);
```
ebroder, line 67: Should we have a sweeper thread on the server in case tokens go unused?

imports/server/images.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  // No match.
+  console.log(`[assets] 404 GET ${req.params.asset}`);
+  res.status(404).send('Sorry, not found');
+});
+
+const UPLOAD_TOKEN_VALIDITY_MSEC = 60000; // 60 seconds
+router.post('/:uploadToken', (req, res) => {
+  check(req.params.uploadToken, String);
+  // Look up upload token.  If missing, or too old (>5m), reject with a 403.
```
ebroder, line 63:
```suggestion
  // Look up upload token.  If missing, or too old (>1m), reject with a 403.
```

imports/server/images.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  // Let's read the uploaded body, which should be just the file contents as a
+  // byte stream, no additional encoding.
+
+  // Aggregate chunks as they stream in.
+  const chunks: Buffer[] = [];
+  req.on('data', (chunk: Buffer) => {
+    chunks.push(chunk);
+  });
```
ebroder, line 83: There's a theoretical DoS/memory exhaustion here, right? I think it'd be safer to keep a running tally of the total bytes uploaded, and bail if that goes over 1M

(Also, FWIW, we could compute the ETag as we go by just calling `update` multiple times, although arguably that's a bit superfluous since the files are small)

zarvox: Strictly speaking, since we do the token check first, it's only DoSable by server admins with valid `UploadToken`s already, which is probably outside our threat model.  That said, it's still reasonable to enforce some resource limits here.

Updating the ETag as we go seems reasonable if we're trying to smooth out the CPU load, but also these things are tiny in the grand scheme of things and I probably can't guess if we'd burn more CPU time in practice making calls to `update` on each chunk vs. the blob as a whole, and I figure we'll only do this a couple times per deployment per year at most in the first place so I didn't plan to overthink it.

imports/server/images.ts
zarvox: Oh yeah I think I meant to do that and forgot

imports/server/images.ts
ebroder: Ah good point on permissions, although something something defense in depth 🤷

imports/server/schemas/blob.ts
```diff
@@ -0,0 +1,23 @@
+import * as t from 'io-ts';
+import { Overrides, buildSchema } from '../../lib/schemas/typedSchemas';
+
+export const BlobCodec = t.type({
+  // Base64-encoded blob contents (since Mongo don't play well with null bytes
+  // in string types, and Meteor doesn't play well with Mongo's binary types)
+  value: t.string,
```
ebroder, line 7: For the record, if we wanted to do this as a binary, I'm pretty sure we could. EJSON [supports](https://docs.meteor.com/api/ejson.html) `Uint8Array` for binary types, so we just (I think) need to define a `io-ts` codec for it and support converting it to a SimpleSchema in `typedSchemas.ts`. Based on the [io-ts docs](https://github.com/gcanti/io-ts/blob/master/index.md#:~:text=success.-,Example,-A) something like:

```
const uint8Array = new t.Type<Uint8Array>(
  'Uint8Array',
  (input: unknown): input is Uint8Array => input instanceof Uint8Array,
  (input, context) => (input instanceof Uint8Array ? t.success(input) : t.failure(input, context)),
  t.identity
);
```

for the former, and then just add a check for `fieldCodec === uint8Array` somewhere in the conditionals in `buildField`.

I don't have a strong opinion about whether it's _worthwhile_ to try and make this work, but I do think it's at least _possible_.

zarvox: I got this to work!  Thanks for the pointers.

imports/server/images.ts
zarvox: Done

imports/server/images.ts
zarvox: Done.

zarvox:
Updated this to use content-addressed assets at `/asset/<file sha256>` for easy future caching.  To support this, we now have a couple additional subscriptions and need to do a bit more server-side rendering in places we might reference an asset, like `site.webmanifest`, the HTML `<head>`, or `browserconfig.xml`.  Each of those is just doing fully-indexed point lookups against `BlobMappings`, which should be cheap.

I wound up unwinding the move of images from public/ to private/; and just put the ones that we have asset-overriding behavior all in `public/images` for consistency.

Lastly, I tacked the relevant commit from #443 on since that seemed easier than trying to keep these changes going separately.

ebroder (approved):
Thanks for the changes. This LGTM (and I will try to follow up with some actual caching in our nginx config)

PR #443, opened 2021-03-28
Make team name configurable

### LINKED ISSUE #193: Make more things whitelabelable

zarvox:
A single commit tacked on to #442; look at that one first; then look at the last commit of this branch.

We add one more `Setting` value to wrap up a string containing the
server name.  We plumb it to the client via a simple pseudocollection.
(Since there's only one such value, it didn't seem worth resurrecting
`PublicConfig`.)

The team name value is used in ~three places:

* in client code, on the login page
* in server code, for puzzle Google doc/sheet titles (two usages)
* in client code, in the user help text talking about the $TEAMNAME
  Discord server.

This change completes our work on custom branding.  With this patchset,
it is now possible to operate a Jolly Roger instance for a team that is
not Death and Mayhem, using the exact same code as upstream, which
should help encourage any teams that wish to operate a Jolly Roger
instance of their own to develop and use the upstream builds, rather
than forking.

As a bonus, it means that our "operating a Jolly Roger instance"
instructions could potentially just tell people to use our public Docker
images directly and bring their own `MONGO_URL`, `MONGO_OPLOG_URL` and
`MAIL_URL`.

Fixes #193.


ebroder (commented):
This LGTM if you want to tack it onto #442 (or just change the merge target?)

zarvox:
I'll just tack this onto #442; thanks.

PR #450, opened 2021-07-03
Only import the icons that we actually use

zarvox:
Using `bundle-analyzer`, I noted that we were spending a substantial
fraction of our client bundle size on the `free-solid-svg-icons`
package.  Digging in, the main file appears to include all thousand
icons, then export them.  Since reify (the Meteor build tool) doesn't do
any tree-shaking, that means we're making clients load every icon in the
icon set, despite only using like two dozen or so of them.

`free-solid-svg-icons` appears to support the same sort of selective
inner import strategy as react-bootstrap to avoid pulling in more code
than intended, so I converted all of our icon imports to that pattern.

free-solid-svg-icons effective size in bundle before: 672  kB
free-solid-svg-icons effective size in bundle after:   16.2kB

Total client bundle size before: 2.23MB
Total client bundle size after:  1.57MB

29.6% smaller.  That should be good for page load times.


ebroder (approved):
Nice catch

Issue #427, opened 2021-01-20
Chat text moves without new messages appearing


catoverdrive:
I noticed a few times this past weekend that the text in the chat window seemed to move even though new messages weren't being sent. This is only mildly annoying, but definitely distracting if you're trying to monitor the chat while puzzling. I think (but haven't verified) this might happen as new viewers open the puzzle or join the audio call causing the box above the chat to resize. Of course, it could also be my eyes playing tricks on me.

(Browser: Google Chrome Version 87.0.4280.141)

zarvox:
Ohhhhh it's probably the length of the audiocall + viewers pane growing (or shrinking) by a row.  Yeah, that sounds like something we could probably be smarter about handling -- forcing scroll to the bottom when the height allocated to the chat history pane changes due to viewer/caller join/part.

PR #454, opened 2021-07-30
Convert ~all client-side code to function components/hooks style

zarvox:
This is a long, mostly mechanical conversion of our components to the more recent Hooks style.

I tried not to make significant behavioral changes, but there's a couple of things that necessitated an alternate approach to get off higher-order components:

* Breadcrumbs.  I've replaced the venerable `react-breadcrumbs-context` with a `useBreadcrumb` hook.  This was very finicky to get right.  I've tried testing all sorts of navigation and field updates, and I think this does the right thing in all circumstances.
* DocumentTitle.  I've replaced this with a simple `useDocumentTitle` that uh, is inspired by the breadcrumbs stuff after some hard-earned lessons on what can work and what can't.  Some day, maybe we should go through things and update the document title any time we update breadcrumbs?  Or even derive the document title from the breadcrumb `title` values?  But this patchset is already disgustingly long as-is, so that's getting backburnered.

I have not yet converted `SplitPanePlus`; it does a bunch of stuff with DOM interactions that I might have to think harder about.

For reviewing: I definitely suggest familiarizing yourself with the [standard hooks API](https://reactjs.org/docs/hooks-reference.html) first.  Then, probably best to just read in commit order, one component-at-a-time.  It's a ton of churn in terms of lines, and indentation, but hopefully despite the reordering, the function bodies are similar enough that it all makes sense.


imports/client/components/PuzzlePage.tsx
```diff
@@ -110,116 +112,131 @@ interface ChatMessageProps {
...
+    if (!messagePane) {
+      return;
+    }
 
     // Include a 5 px fudge factor to account for bad scrolling and
     // fractional pixels
-    this.shouldScroll = (messagePane.clientHeight + messagePane.scrollTop + 5 >= messagePane.scrollHeight);
-  };
+    shouldScroll.current = (messagePane.clientHeight + messagePane.scrollTop + 5 >= messagePane.scrollHeight);
+    console.log(`shouldScroll: ${shouldScroll.current}`);
```
ebroder, line 162: Is this debugging cruft? (Why didn't eslint complain?)

ebroder (approved):
This is great. I'm in favor purely on the basis of keeping up with upstream best practice, but I find some of the restructuring that hooks require to be more legible, so that seems like a win. Thanks for all the time you put into this.

imports/client/components/PuzzlePage.tsx
zarvox: Yeah.  I probably should have taken it out for this PR, though I likely still need it to look into a proper fix for #427.

eslint didn't complain because we disable the `max-len` and `no-console` rules at the top of this file.

PR #457, opened 2021-08-06
Security updates

ebroder:


zarvox:
I thought popper was the thing that powered the popovers on hover -- if you remove it, do we lose e.g. the related-puzzles popovers?

We seem to import from `@popperjs/core` in a couple places in imports/client/components/Tag.tsx

zarvox:
Oh I see you commented that that's a different package.

We probably *should* have an explicit dep on popperjs if we're importing it in our source, rather than relying on it transitively.

ebroder:
Turns out there's an eslint for that :)

.meteor/versions
```diff
@@ -11,6 +11,7 @@ babel-runtime@1.5.0
 base64@1.0.12
 binary-heap@1.0.11
 boilerplate-generator@1.7.1
+bundle-visualizer@1.2.3
```
zarvox, line 14: Probably don't actually want to check this in.

zarvox:
unintended `bundle-visualizer` dep aside, LGTM

.meteor/versions
ebroder: Gah, damnit. Fixed

zarvox (approved):
Thanks for the cleanups!

PR #458, opened 2021-08-06
Add async loading for SetupPage and RTCDebugPage

### LINKED ISSUE #111: Incorporate dynamic-import
Once we've upgraded to Meteor 1.5, we can split out the app into multiple JS bundles. There's some [additional documentation](https://blog.meteor.com/dynamic-imports-in-meteor-1-5-c6130419c3cd) on the Meteor blog; it seems like it integrates well with async/await.

We should figure out where we want to split the app into multiple bundles and use dynamic require to do that.

ebroder:
This is made pretty straightforward with the introduction of React.lazy and the Suspense component. The only thing we really need to provide is a placeholder component to use as a loading indicator.

This doesn't seem to make much of a dent in the bundle size, likely because most of the bundle is dependencies, and I don't think removing these pages allows for pruning many dependencies, but it still seems like good hygiene and a good leaping off point for more code splitting later.

Here's what the loading component looks like for reference:

Screenshot from 2021-08-06 11-29-33

For what it's worth, I'm somewhat tempted to switch some of the "loading..." text placeholders to use the spinner instead, since I think the UI is a little cleaner. I'll probably play with it some and see if it works.

I'm going to say this fixes #111, although there is more work to do if we want.

### DIFF
--- a/imports/client/components/Routes.tsx
+++ b/imports/client/components/Routes.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Suspense } from 'react';
 import { Route, Switch } from 'react-router';
 import { BreadcrumbsProvider } from '../hooks/breadcrumb';
 import useDocumentTitle from '../hooks/use-document-title';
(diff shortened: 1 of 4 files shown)


ebroder:
FWIW I think my next stab at code splitting will be pulling out components that depend on `react-select`, since I think those are somewhat infrequently used and should let us actually trim a dependency from the default bundle.

imports/client/components/Routes.tsx
```diff
@@ -8,37 +8,41 @@ import EnrollForm from './EnrollForm';
...
+
 const Routes = React.memo(() => {
   useDocumentTitle('Jolly Roger');
 
   return (
     <BreadcrumbsProvider>
-      <Switch>
-        {/* Index redirect */}
-        <Route exact path="/" component={RootRedirector} />
+      <Suspense fallback={<Loading />}>
```
zarvox, line 26: Does putting `Suspense` here mean that if we were to e.g. make other things (like, say, Select2) dynamically load, it'd block the whole Route from rendering instead of just the thing we're trying to lazy-load?

https://reactjs.org/docs/react-api.html#reactsuspense suggests that:

> The best practice is to place `<Suspense>` where you want to see a loading indicator, but to use `lazy()` wherever you want to do code splitting.

So I'm suspicious of using Suspense way down in the root of the Routes page.  Maybe it makes more sense to inline a `<Suspense>` under the `component=` for each of the two lazy-loaded routes?

ebroder: That's where the [React docs](https://reactjs.org/docs/code-splitting.html#route-based-code-splitting) suggest putting the `Suspense` component if you're splitting by route. I think if we make more things load selectively, I'd add additional `Suspense` components further in the component tree - my assumption is that you can nest them and the innermost one gets the trigger?

zarvox: Fair enough.  I guess if that turns out not to be true, we'll find out when we get there, and in the meantime this is a win.

zarvox (approved):

PR #460, opened 2021-08-08
Make react-select dynamically imported

ebroder:
The puzzle edit modal and the tag editor are the only 2 places where we use react-select, which is an otherwise fairly expensive require (about 60K in the final bundle). By pushing the react-select usages to use dynamic-import we can slim down the default fetch.

I tested this using the Chrome dev tools with a custom throttling profile that added 2 seconds of latency.

Here's what the tag selector looks like:

suspense-tag-editor

And here's what the puzzle edit modal looks like:

suspense-puzzle-modal

### DIFF
--- a/client/stylesheets/loading.scss
+++ b/client/stylesheets/loading.scss
@@ -1,8 +1,11 @@
-.loading {
-  position: fixed;
+.loading-fullsize {
+  position: absolute;
+  top: 0;
+  left: 0;
   height: 100%;
   width: 100%;
   display: flex;
   justify-content: center;
   align-items: center;
+  background: rgba(255, 255, 255, 0.5);
 }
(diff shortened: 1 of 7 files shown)


client/stylesheets/loading.scss
```diff
@@ -1,8 +1,11 @@
-.loading {
-  position: fixed;
+.loading-fullsize {
```
zarvox, line 1: Is there supposed to be a `loading-inline` class defined somewhere, or is the implication that `.loading-inline` as used in Loading.tsx carries no styles?

zarvox (approved):

client/stylesheets/loading.scss
ebroder: No styling currently. Mostly wanted to preserve future optionality.

zarvox: Seems reasonable; just wanted to make sure I wasn't missing something.

PR #462, opened 2021-08-09
Remove some unnecessary uses of underscore

ebroder:
Three main types of changes here:

* Removing the underscore import from files where it wasn't used at all. Typescript makes it an error to have unused variables, but I think it misses this because it starts with (is) an underscore.
* Replacing underscore with native equivalents. I think in the early days of working on jolly-roger, I used underscore more heavily because I assumed that it new something about, like, weird Javascript prototypical inheritance that I didn't, but in practice there's really no reason to use `_.map` or `_.filter`.
* Replacing underscore where it obfuscates typing information. The type annotations on underscore are mostly pretty good, but a few of them are quite bad - `_.extend` and `_.omit` in particular have a tendency to just return `any`. Plus the spread operator and destructuring assignment in modern JS are basically equivalent anyway.

I thought about introducing a lint rule (I've been using https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore as my goto for "when are there easy native equivalents") but decided not to for the time being. I also looked into whether we could eliminate underscore entirely, but `_.indexBy`, `_.groupBy`, and `_.throttle` are hard to replace (and `_.sortBy` is a little awkward). Since it's only 14K of bundle, it didn't seem worth it.

### DIFF
--- a/imports/client/components/HuntApp.tsx
+++ b/imports/client/components/HuntApp.tsx
@@ -1,7 +1,6 @@
 import { Meteor } from 'meteor/meteor';
 import { Roles } from 'meteor/nicolaslopezj:roles';
 import { useTracker } from 'meteor/react-meteor-data';
-import { _ } from 'meteor/underscore';
 import React, { useCallback, useMemo } from 'react';
 import Alert from 'react-bootstrap/Alert';
 import Button from 'react-bootstrap/Button';
@@ -139,7 +138,7 @@ const HuntApp = React.memo((props: RouteComponentProps<HuntAppParams>) => {
       _id: huntId,
     });
     const user = Meteor.user();
-    const member = !!user && _.contains(user.hunts, huntId);
+    const member = !!user && user.hunts.includes(huntId);
     return {
       ready: userHandle.ready() && huntHandle.ready() && deletedHuntHandle.ready(),
       hunt: Hunts.findOneAllowingDeleted(huntId),
--- a/imports/client/components/HuntApp.tsx
+++ b/imports/client/components/HuntApp.tsx
@@ -138,7 +138,7 @@ const HuntApp = React.memo((props: RouteComponentProps<HuntAppParams>) => {
       _id: huntId,
     });
     const user = Meteor.user();
-    const member = !!user && user.hunts.includes(huntId);
+    const member = user?.hunts?.includes(huntId) ?? false;
     return {
       ready: userHandle.ready() && huntHandle.ready() && deletedHuntHandle.ready(),
       hunt: Hunts.findOneAllowingDeleted(huntId),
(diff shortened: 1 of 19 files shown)


imports/client/components/UnauthenticatedRoute.tsx
```diff
@@ -50,7 +49,7 @@ const UnauthWrapper = React.memo((props: UnauthWrapperProps) => {
   }
 
   if (tracker.userId) {
-    const state = _.extend({ pathname: '/', search: undefined }, props.location.state);
+    const state = { pathname: '/', search: undefined, ...(props.location.state as any) };
```
ebroder, line 52: FWIW this (and other `as any` casts that I added) aren't actually any worse than what we had before, but the `_.extend` or `_.omit` were masking it.

imports/client/components/HuntApp.tsx
```diff
@@ -139,7 +138,7 @@ const HuntApp = React.memo((props: RouteComponentProps<HuntAppParams>) => {
       _id: huntId,
     });
     const user = Meteor.user();
-    const member = !!user && _.contains(user.hunts, huntId);
+    const member = !!user && user.hunts.includes(huntId);
```
zarvox, line 141: Careful -- I think we also need to guard `user.hunts` being undefined.

ebroder: Is that true? It's not defined as an optional property

zarvox: Its seems obvious to me that anything that we're attaching to `Meteor.user()` is going to have *some* transient state where the property isn't present.

I agree that most users will have hunts with at least one hunt, as a result of other flows calling `Meteor.users.update`, but the first user obviously won't until they create a hunt.  So yeah, I think the type signature is overly optimistic.

ebroder: Fixed by changing the definition to be optional and fixing the fallout

imports/lib/models/base.ts
```diff
@@ -98,28 +97,28 @@ class Base<T extends BaseType> extends Mongo.Collection<T> {
 
   find(selector: FindSelector<T> = {}, options: FindOptions = {}) {
     return super.find(
-      _.extend({ deleted: false }, this[formatQuery](selector)),
+      { deleted: false as any, ...this[formatQuery](selector) },
```
zarvox, line 100: Do you intend for the `deleted: false` to be overrideable at the query site here, or do you mean for users to have to use the appropriate `findDeleted` or `findAllowingDeleted` accessor?

Because with object spreads, the last-mentioned value wins, which means that you can just override it from the query.

zarvox (approved):

imports/lib/models/base.ts
ebroder: Yeah it does seem like the behavior is questionable, but `_.extend` has the same behavior so this isn't a behavior change. I guess I can fix

PR #463, opened 2021-08-10
Remove `nicolaslopezj:roles`

zarvox:
When trying to take the Meteor 2.3.4 update, we discovered that it was being held back by `nicolaslopezj:roles`, which has not been updated since 2018, which had a dependency on an older version of `accounts-base` than is available in Meteor 2.3.4.

Separately, we're interested in potentially making the `operator` role something that can have a finer granularity than "per server" (which is how it currently works).  This change series does some (but nowhere near all) of the refactoring work needed to support per-hunt operator status.

This is not meant to be a particularly pretty or DRY reimplementation of Roles.  This is meant to be a mostly-mechanical stopgap that unblocks taking updates.


imports/client/components/ProfilePage.tsx
```diff
@@ -51,9 +54,9 @@ const ProfilePage = (props: RouteComponentProps<ProfilePageParams>) => {
         discordAccount: undefined,
         muteApplause: undefined,
       },
-      viewerCanMakeOperator: Roles.userHasPermission(Meteor.userId(), 'users.makeOperator'),
-      viewerIsOperator: Roles.userHasRole(Meteor.userId()!, 'operator'),
-      targetIsOperator: Roles.userHasPermission(uid, 'users.makeOperator'),
+      viewerCanMakeOperator: deprecatedUserMayMakeOperator(Meteor.userId()),
```
zarvox, line 57: My best idea for how to make these not rely on the deprecated global calls is to add the route `/hunts/:huntId/hunters/:userId` for hunt-specific user lookup and then only present hunt-specific operator-y context on that page.

Then, the global user pages at `/users/:userId` could drop the operator-specific components.

imports/client/components/NotificationCenter.tsx
```diff
@@ -364,7 +370,7 @@ type NotificationCenterTracker = {
 
 const NotificationCenter = () => {
   const tracker: NotificationCenterTracker = useTracker(() => {
-    const canUpdateGuesses = Roles.userHasPermission(Meteor.userId(), 'mongo.guesses.update');
+    const canUpdateGuesses = deprecatedIsActiveOperator(Meteor.userId());
```
zarvox, line 373: `pendingGuesses` might need to be restructured or made to work for non-operators or scoped to specific hunts or something to make this work reasonably in a tracking-operator-per-hunt world.

imports/server/hunts.ts
```diff
@@ -74,7 +79,90 @@ function renderExistingJoinEmail(setting: SettingType | undefined, user: Meteor.
...
+
+    return huntId;
+  },
+
+  updateHunt(huntId: unknown, value: unknown) {
+    check(this.userId, String);
+    checkAdmin(this.userId);
+    check(huntId, String);
+    // Despite looking right to me, TS seems to trip over the [String] pattern for mailingLists.
+    // @ts-ignore Something is wrong with check's array-of-strings typing
```
zarvox, line 124: Feel free to take a crack at winning a duel with the typechecker here; I gave up and moved on.

(In an ideal world, maybe we'd be able to synthesize these match patterns from the io-ts types or something?  Or use a `Match.Where` that calls the codec to see if it deserializes cleanly?  But maybe we'd rather find some other approach with smaller codesize?  🤷🏻‍♂️ )

imports/server/api/resources/users.ts
```diff
@@ -37,13 +37,14 @@ const ACTIVE_THRESHOLD = 365 * 24 * 60 * 60 * 1000;
 const renderUser = function renderUser(user: Meteor.User, profile: ProfileType) {
   const active = user.lastLogin &&
           Date.now() - user.lastLogin.getTime() < ACTIVE_THRESHOLD;
+  const operator = deprecatedUserMayMakeOperator(user._id);
```
zarvox, line 40: @grantaelliott as the maintainer of the only known jolly-roger API client: do you know if you actually rely on the presence/value of the `operator` field from the jolly-roger API for the POTW site or any of the other hunt sites you've written?  Would love to have the flexibility to drop this field or restructure it on a per-hunt basis so long as it wouldn't break you.

ebroder: This is why we should switch to graphql :joy:

imports/server/hunts.ts
ebroder: It seems like the issue is that Typescript is inferring the type of `[String]` to be an array of unspecified size (`StringConstructor[]`), rather than a tuple type (`[StringConstructor]`), and that resolution happens at assignment time and isn't updated or refined when the object is passed into `check`.

imports/server/hunts.ts
```diff
@@ -74,7 +79,90 @@ function renderExistingJoinEmail(setting: SettingType | undefined, user: Meteor.
...
 }
 
+const SavedDiscordObjectFields = {
+  id: String,
+  name: String,
+};
+
+const HuntShape = {
+  name: String,
+  mailingLists: [String],
```
ebroder, line 89:
```suggestion
  mailingLists: [String] as [StringConstructor],
```

imports/server/hunts.ts
```diff
@@ -74,7 +79,90 @@ function renderExistingJoinEmail(setting: SettingType | undefined, user: Meteor.
...
+
 Meteor.methods({
+  createHunt(value: unknown) {
+    check(this.userId, String);
+    checkAdmin(this.userId);
+    // Despite looking right to me, TS seems to trip over the [String] pattern for mailingLists.
+    // @ts-ignore Something is wrong with check's array-of-strings typing
+    check(value, HuntShape);
+
+    const huntId = Hunts.insert(value as HuntInsertFields);
```
ebroder, line 108:
```suggestion
    check(value, HuntShape);

    const huntId = Hunts.insert(value);
```

imports/server/hunts.ts
ebroder: Also we could use `Match.Where` but when I originally added the typing I didn't make the typing for that specific enough. Right now it's

```
        function Where(condition: (val: any) => boolean): Matcher<any>;
```

and I think it would need to be something like

```
        function Where<T>(condition: (val: any) => val is T): Matcher<T>;
```

zarvox:
@ebroder Is this good now, or is there something else you'd like me to address first?

imports/client/components/ProfilePage.tsx
ebroder: Ugh. We don't need to solve this now but that seems messy. It's already pretty messy that there are 2 different ways to get to the profile list that look the same but show different things

imports/lib/models/base.ts
```diff
@@ -35,8 +36,22 @@ class Base<T extends BaseType> extends Mongo.Collection<T> {
     this.name = name;
     this.tableName = tableName;
 
-    // All models have standard roles
-    this.attachRoles(`mongo.${name}`);
+    // Let admins get away with absolutely anything from the client.
+    // This is an affordance intended for use by a human trying to fix things
+    // in production with the `Models` global manually populated by the
+    // `loadFacades()` global function, and these allow rules should never be
+    // relied upon by any client-side code that is part of the application.
```
ebroder, line 43: Hmm. This makes me wonder if we should figure out some way to make these rules off by default? We could have a Meteor method that enables them per-session or something like that and upgrade `loadFacades()` to, like, `enableAdminMode()` or whatever. Not a today problem but it makes me wonder if we can systematically prevent accidentally relying on these in the future.

ebroder (approved):

imports/lib/models/base.ts
zarvox: That's a nice idea -- we could have these rules also check a global boolean or something that we can toggle via a Meteor method `breakGlass()`, and maybe have it automatically set a 10-minute timer to flip the boolean back to false?

Or it could just be a feature flag we can toggle from `/setup`

imports/client/components/ProfilePage.tsx
zarvox: Yeah -- I think we should probably nudge UI towards the hunt-scoped option.

I think if we're thinking about unrelated entities having accounts, it'd be pretty reasonable to restrict usage of `/users` and `/users/:userId` to admins and make everything else be hunt-scoped.  Might need to relocate `OwnProfilePage` to somewhere a little more generic.

imports/server/hunts.ts
ebroder: Delayed update: I believe that once DefinitelyTyped/DefinitelyTyped#55140 lands and we pick it up, we could replace this with `check(value, Match.Where(HuntCodec.is))`.

imports/server/api/resources/users.ts
grantaelliott: POTW does not depend on the operator field.

PR #476, opened 2021-12-03
Convert most CSS to styled-components

ebroder:
Alright, I think this is ready for a review.

It should be reviewable commit-at-a-time. In most cases, the conversion is comparatively rote; I noted in the commit messages where it wasn't. Some elements like notifications required some structural changes to the CSS (to avoid things like caring whether a parent element was `:last-child`), but none of those should have created. There were a few cases where I decided that the styling change from the CSS was too subtle to be worth the effort of preserving, and I dropped those. There's also a lot of garbage collection of unused styles.

I built up a handful of helpers in `imports/client/components/styling`. Of those, `responsive.tsx` is probably worthy of the most discussion. We get the `@media-breakpoint-*` SCSS mixins from Bootstrap and don't have access to them in the raw-ish CSS we're writing in styled-components. The breakpoints are exposed via CSS variables, but you also can't use `var` or `calc` in `@media` queries in raw CSS. The logic is ripped off pretty directly from what the SCSS mixins are using. I only implemented `mediaBreakpointDown` for now because it's the only one we ever use, but adding `mediaBreakpointUp`/`mediaBreakpointOnly`/etc. should mostly just be a small matter of writing code. The helper function takes a styled-components CSS fragment as an argument - there are examples of usage in `App.tsx` and `PuzzleListPage.tsx`. It's not a perfect DX but I think it's a pretty decent compromise.

For naming, I generally tried to minimize the number of throwaway variables used purely for wrapping components, so styled component names often just replace existing components (e.g. importing BS's `Breadcrumb` as `BSBreadcrumb` and then using `const Breadcrumb = styled(BSBreadCrumb)...`. I used `StyledFoo` as a prefix in a handful of places where the separate variable was unavoidable.

I decided not to convert `PuzzlePage` / `puzzle.scss`. We can save that for a followup, or for @grantaelliott 🙂.


imports/client/components/Celebration.tsx
```diff
@@ -9,6 +10,32 @@ interface CelebrationProps {
...
+  background-color: rgba(0, 0, 0, .20);
+  z-index: 1050;
+`;
+
+const CelebrationContainer = styled.div`
+  background-color: #f0fff0;
+  padding: 24px;
+`;
+
+const CelebrationClose = styled.button`
```
zarvox, line 33: Not super attached to this, but for some reason the first thing I thought was "this makes it harder to tell that the element is actually a `<button>`, which makes it slightly harder to understand what props it's likely to accept" -- maybe call this `CelebrationCloseButton`, which hints at its underlying control?

imports/client/components/GuessQueuePage.tsx
```diff
@@ -138,7 +210,7 @@ interface GuessQueuePageProps extends GuessQueuePageWithRouterParams {
 }
 
 const GuessQueuePage = (props: GuessQueuePageWithRouterParams) => {
-  useBreadcrumb({ title: 'Guess queue', path: `/hunts/${props.match.params.huntId}/guesses` });
+  useBreadcrumb({ title: 'Guess queue', path: `/ hunts / ${props.match.params.huntId} /guesses` });
```
zarvox, line 213: I haven't tested this, so correct me if I'm wrong, but I think this new whitespace here is going to do unwelcome things to the breadcrumb link

imports/client/components/GuessQueuePage.tsx
```diff
@@ -83,27 +155,27 @@ const GuessBlock = React.memo((props: GuessBlockProps) => {
...
           {' from '}
-          <span className="breakable">{props.createdByDisplayName || '<no name given>'}</span>
+          <Breakable>{props.createdByDisplayName || '<no name given>'}</Breakable>
         </div>
         <div>
           {'Puzzle: '}
           <a href={guessURL(props.hunt, props.puzzle)} target="_blank" rel="noopener noreferrer">{props.puzzle.title}</a>
           {' ('}
-          <Link to={`/hunts/${props.puzzle.hunt}/puzzles/${props.puzzle._id}`}>discussion</Link>
+          <Link to={`/ hunts / ${props.puzzle.hunt} /puzzles/${props.puzzle._id} `}>discussion</Link>
```
zarvox, line 178: not sure I grok what the whitespace is doing here

imports/client/components/GuessQueuePage.tsx
```diff
@@ -57,10 +58,81 @@ interface GuessBlockProps {
...
+    border: 1px solid #0000ff;
+    background-color: #f0f0ff;
+  `}
+`;
+
 const formatDate = (date: Date) => {
   // We only care about days in so far as which day of hunt this guess was submitted on
   const day = daysOfWeek[date.getDay()];
-  return `${date.toLocaleTimeString()} on ${day}`;
+  return `${date.toLocaleTimeString()} on ${day} `;
```
zarvox, line 135: Another sorta-odd trailing whitespace

imports/client/components/AccountForm.tsx
```diff
@@ -177,47 +204,42 @@ const AccountForm = (props: AccountFormProps) => {
   }[props.format];
 
   const emailInput = (
-    <div className="at-input form-group">
-      <label className="control-label" htmlFor="at-field-email">Email</label>
+    <div className="form-group">
```
zarvox: I still think (someday) this would all benefit from just using react-bootstrap directly, but that needn't be now

zarvox: Aaaand literally the next commit does that.  This is what happens when I review commit-by-commit 😂

imports/client/components/PuzzleListPage.tsx
```diff
@@ -346,6 +347,52 @@ interface PuzzleListPageTracker {
...
+  }
+`;
+
+const StyledPuzzleListExternalLink = styled(StyledPuzzleListLink)`
+  flex: 0 0 40px;
+`;
+
+const StyledPuzzleListLinkLabel = styled.span`
+  margin-left: 4px;
+  @media (max-width: ${() => window.getComputedStyle(document.body).getPropertyValue('--breakpoint-sm')}) {
```
zarvox: Just out of curiosity: how often does this get computed?  Seems like the sort of thing that could wind up causing JS to block on DOM operations and lead to jank if it's every render or something.

zarvox:
Cool, I feel like this is ready to go, modulo a couple small nits.

imports/client/components/GuessQueuePage.tsx
ebroder: I'm not sure where this whitespace came from. I think an over-zealous auto-formatter

imports/client/components/PuzzleListPage.tsx
ebroder: As far as I can tell, it's only evaluated on load. Right now there are 3 calls to `mediaBreakpointDown` so it gets called 3 times. We could theoretically build out some caching here, but I worry that we'd have to think about, like, load order and stuff like that and would mildly prefer to just not, but I can look into it if you think it's important.

ebroder:
PTAL?

zarvox (approved):

zarvox:
Thanks for all the effort you put into this!  It looks really good! :)

PR #477, opened 2021-12-10
Cleanups

ebroder:
These came up while I was working on the SFU code, but aren't directly tied to any of that code so splitting out for an earlier review.

### DIFF
--- a/server/main.ts
+++ b/server/main.ts
@@ -33,3 +33,10 @@ import '../imports/server/setup';
 import '../imports/server/subscribers';
 import '../imports/server/operator';
 import '../imports/server/users';
+
+// Imports are necessary to make sure the modules are in the bundle
+import ModelsFacade from '../imports/lib/models/facade';
+import SchemasFacade from '../imports/lib/schemas/facade';
+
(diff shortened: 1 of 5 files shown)


server/main.ts
```diff
@@ -33,3 +33,10 @@ import '../imports/server/setup';
 import '../imports/server/subscribers';
 import '../imports/server/operator';
 import '../imports/server/users';
+
+// Imports are necessary to make sure the modules are in the bundle
+import '../imports/lib/models/facade';
```
zarvox, line 38: Seems a little odd to write it with a pair of side-effectful imports and then to go require it separately again?  Any reason not to just:

```js
import ModelsFacade from '../imports/lib/models/facade';
import SchemasFacade from '../imports/lib/schemas/facade';
(global as any).Models = ModelsFacade;
(global as any).Schemas = SchemasFacade;
```

?

ebroder: Oh yeah, that would work - I added the imports after I had the assignment because otherwise the modules aren't available. Fixed.

PR #479, opened 2021-12-12
Upgrade @types/meteor

ebroder:
The newer version of @types/meteor is a bit more sensitive about null
vs. undefined. All of the new sensitivity is in places that don't matter
to us (i.e. we were passing in nullable values that we knew couldn't be
null), so just add some extra non-null assertions.

The types also include their own declarations of
Promise.{await,awaitAll,async,asyncApply,prototype.await} so we no
longer need to provide our own declaration.

(I took the upgrade on my SFU branch...for reasons I no longer remember)


imports/lib/models/document_permissions.ts
```diff
@@ -5,7 +5,7 @@ import Base from './base';
 const DocumentPermissions = new Base<DocumentPermissionType>('document_perms');
 DocumentPermissions.attachSchema(DocumentPermissionSchema);
 DocumentPermissions.publish(function (this: Subscription, q) {
-  return { ...q, user: this.userId };
+  return { ...q, user: this.userId! };
```
zarvox, line 8: I don't think we have anything that actually prevents this from being `undefined/null` if these subs get requested by a connection where no user has logged in?

ebroder:
The Base.publish method checks before calling the query modifier

imports/model-helpers.ts
```diff
@@ -27,7 +27,7 @@ const huntsMatchingCurrentUser = function <T extends HuntModel> (
   // As a note: this will not re-publish if the user's hunt membership
   // changes, so use it carefully (basically, use it when you already
   // know the user is a member of the hunt in question).
-  const u = MeteorUsers.findOne(this.userId);
+  const u = MeteorUsers.findOne(this.userId!);
```
zarvox, line 30: It's not clear to me that `this.userId` is guaranteed to be a string here.  In fact, we appear to be checking that we got a user, and I think this would just turn into `MeteorUser.findOne(undefined)`, which might just return an arbitrary user?

Either way I'd be happier if we just checked `this.userId` explicitly in this function instead of writing `this.userId!`

zarvox (approved):
Ahhh, I missed the check in `publishFunc`.  This all seems good then

ebroder:
I think I found a better approach. PTAL?

imports/lib/models/base.ts
```diff
@@ -167,9 +166,12 @@ class Base<T extends BaseType> extends Mongo.Collection<T> {
           return [];
         }
 
-        let query = q;
+        let query: Mongo.Selector<T> | undefined = q;
         if (modifier) {
-          query = modifier.apply(this, [q]);
+          query = modifier(this.userId, q);
```
zarvox, line 171: Is there a reason to prefer `this.userId` over just capturing `userId` from the first argument to `publish()` here?

ebroder: `userId` is the first argument to `modifier`, which is a functional argument to publish.

zarvox (approved):

PR #483, opened 2021-12-14
Clean up shutdown process

ebroder:
Meteor doesn't set signal handlers by default, so when we set a signal
handler, we were in fact, bypassing the default handling of the signal
(i.e. exiting). Add a helper which runs signal handlers, cleans up after
itself, and then re-signals to actually cause the process to exit.

### DIFF
--- a/imports/server/onExit.ts
+++ b/imports/server/onExit.ts
@@ -1,0 +1,12 @@
+const exitHandlers: (() => void)[] = [];
+
+['SIGINT' as const, 'SIGTERM' as const, 'SIGHUP' as const].forEach((signal) => {
+  process.once(signal, () => {
+    exitHandlers.splice(0).forEach((handler) => handler());
+    process.kill(process.pid, signal);
+  });
+});
+
+export default function onExit(handler: () => void) {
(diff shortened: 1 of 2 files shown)


imports/server/onExit.ts
```diff
@@ -0,0 +1,12 @@
+const exitHandlers: (() => void)[] = [];
+
+['SIGINT' as const, 'SIGTERM' as const, 'SIGHUP' as const].forEach((signal) => {
+  process.once(signal, () => {
```
zarvox, line 4: Checking to be sure I'm understanding this logic:

For each signal type, we register an event handler to be run exactly once upon delivery of that signal.

Upon the first such signal to arrive, the event handler takes the list of exit handlers, and runs each cleanup function in the order they were registered, exactly once.

And then, the event handler signals itself *again* with the original signal, which, because we've no longer got a handler for those signals registered, ultimately causes the process to exit.

Is that right?

zarvox (approved):

imports/server/onExit.ts
ebroder: Yep, that's right

PR #484, opened 2021-12-22
Switch to using a Mediasoup SFU instead of peer-to-peer WebRTC connections

ebroder:
Instead of using peer-to-peer WebRTC connections, run a Mediasoup SFU on the server and connect to it using Mediasoup's client library. On the server, each active call corresponds to a Mediasoup router, which is
created on the server that is handling requests for the first peer to connect.

Other peers might have a Meteor connection to a different server, which implies that all communication needs to happen via the database. The server and client communicate with each other by creating various database records.

For ease of understanding, this replaces the CallParticiant and CallSignal models from the original implementation, and breaks out the client-server communication into many different models. (In theory, I think that most could have been crammed into CallSignals, but this was easier to reason about.)

This gets us down to 2 connections from a given peer - one for sending and one for receiving (the Mediasoup client library only supports uni-directional connections). Structurally, on the client, this means that CallSection is now responsible for connecting to the SFU (instead of each instance of CallLinkBox, which no longer exists, being responsible), and fanning out components for each peer. (A client receives a consumer with an audio track for each peer.)

Using a central SFU server that is known to be internet accessible means that we no longer require STUN/TURN for peer-to-peer NAT busting, so that configuration is all removed.

Additionally, this clamps down the timeouts on dead server garbage collection - since all call traffic goes through the server, a dead server is potentially a much larger impediment to collaboration, so liveness matters more than it used to.

Some outstanding things to do before we should consider merging this:

- [x] Check for symmetry of resource acquisition/teardown (including calls to `observeChanges`)
- [x] Make sure that every Mediasoup DB record has a dead server garbage collection somewhere
- [ ] Check the diff for outstanding `TODOs`
- [x] Do something with `RTCDebugPage`
- [x] Tamp down the logging to something reasonable
- [x] Add additional indication on client when we have a `Peer` but are missing a `Room`, `Router`, or `Transport`, since those can indicate that we're waiting to GC old state from a disappearing server


imports/server/mediasoup.ts
```diff
@@ -0,0 +1,763 @@
...
+    // oplog events are processed in order, so we need a locking scheme like
+    // "discord-bot" in discord-client-refresher
+    try {
+      Rooms.insert({
+        call: peer.call,
+        routedServer: peer.createdServer,
+        createdBy: peer.createdBy,
+      });
+    } catch (e) {
+      // 11000 is a duplicate key error
```
zarvox, line 413: Given how many places we use this, it might be worth extracting a function `isMongoDuplicateKeyError(e)` for use here/below/in locks.ts/puzzle.ts

imports/server/mediasoup.ts
```diff
@@ -0,0 +1,763 @@
...
+    Ansible.log('Peer joined call', { peer: peer._id, call: peer.call, createdBy: peer.createdBy });
+    if (!this.callToPeers.has(peer.call)) {
+      this.callToPeers.set(peer.call, new Set());
+    }
+    this.callToPeers.get(peer.call)!.add(peer._id);
+    this.peerToCall.set(peer._id, peer.call);
+
+    // TODO: ideally we only want this to happen on one server to ensure that
+    // oplog events are processed in order, so we need a locking scheme like
+    // "discord-bot" in discord-client-refresher
```
ebroder, line 405: For the record, I think the potential raciness here is the section of code I'm most uncomfortable with right now - specifically, races where the last peer leaves the call and then a new peer joins at around the same time, potentially leaving us with peers but no room. I think it's unlikely but hard to recover from.

Given that, I'm tempted to stop observing `Peers` here, and instead create/destroy the `Room` in the `mediasoup:join` subscription, using a lock to avoid raciness.

(I think I will still want something similar to the `remoteRoomRemoved` listener, though, to account for server death. Plus, it potentially gives us an easy hook for rebalancing rooms across servers)

imports/server/mediasoup-api.ts
```diff
@@ -0,0 +1,327 @@
...
+    throw new Meteor.Error(404, 'Peer not found');
+  }
+
+  if (peer.createdBy !== this.userId) {
+    throw new Meteor.Error(403, 'Not allowed');
+  }
+
+  const transportRequest = TransportRequests.insert({
+    createdServer: serverId,
+    routedServer: peer.createdServer,
```
zarvox, line 125: This should be router, not peer

imports/server/mediasoup.ts
```diff
@@ -0,0 +1,763 @@
...
+        transportId: transport.id,
+      }, {
+        $set: {
+          dtlsState,
+          createdBy: transport.appData.createdBy,
+        },
+      });
+    }));
+
+    this.producers.forEach((producer) => {
```
zarvox, line 320: This wants to get filtered down to just the call

imports/server/mediasoup.ts
```diff
@@ -0,0 +1,763 @@
...
+  }
+
+  onProducerCreated(producer: types.Producer) {
+    producer.observer.on('close', Meteor.bindEnvironment(() => {
+      this.producers.delete(producer.appData.producerClient);
+      ProducerServers.remove({ producerId: producer.id });
+    }));
+
+    // Create consumers for other existing transports
+    this.transports.forEach(async (transport, key) => {
```
zarvox, line 346: This too needs to get filtered down to just the call

imports/client/components/CallSection.tsx
```diff
@@ -1,137 +1,159 @@
...
+  });
 
-  const spectrumRefCallback = useCallback((spectrum) => {
-    if (spectrum) {
-      spectrum.connect(localStream);
-    }
-  }, [localStream]);
+  const [tracks, setTracks] = useState<MediaStreamTrack[]>([]);
+  useEffect(() => {
+    const captureTracks = () => Meteor.defer(() => setTracks(stream.getTracks()));
```
zarvox, line 145: comment why the eventloop yield is needed (tracks aren't reachable from stream.getTracks() yet)

imports/server/mediasoup.ts
ebroder: Fixed in 1609128

imports/server/mediasoup.ts
ebroder: (Also fixed in 1609128)

imports/client/components/CallSection.tsx
ebroder: Fixed in 0cd500b

imports/client/components/CallSection.tsx
```diff
@@ -161,79 +223,552 @@ const RTCCallSection = (props: RTCCallSectionProps) => {
...
+    }).then(() => setDevice(newDevice));
+  }, [router.rtpCapabilities]);
+
+  useSubscribe(device ? 'mediasoup:transports' : undefined, selfPeer._id, JSON.stringify(device?.rtpCapabilities));
+  const { sendServerParams, recvServerParams } = useTracker(() => {
+    return {
+      // TODO: should we try to detect this specific transport request
+      sendServerParams: Transports.findOne({ peer: selfPeer._id, direction: 'send' }),
+      recvServerParams: Transports.findOne({ peer: selfPeer._id, direction: 'recv' }),
+    };
```
ebroder, line 627: @zarvox Interested in your feedback. If, for whatever reason, we churn the `mediasoup:transports` subscription, then potentially we could end up with 2 different `TransportRequest` records and therefore 4 different `Transport` records. In that case, these `findOne`'s could potentially find transports from the older `TransportRequest`, which presumably is about to be torn down. I'm not actually sure when this would happen.

I see two options:

1. We could solve this by having some way to identify the current `TransportRequest`, but there's no good unique ID right now. We'd have to generate a unique ID, thread that through the subscription, and make sure it gets stored onto the `TransportRequest` and copied onto the `Transports`. We'd need to regenerate the ID whenever the subscription got churned, although since I'm not sure when that would happen I don't know when we'd need to churn the ID.
2. We could do nothing and let this theoretical race exist. It should resolve cleanly (eventually the old `Transports` will get deleted and we'll pick up the new ones) so might just create some unnecessary churn

Thoughts?

imports/server/mediasoup.ts
ebroder: I added a function to `imports/server/mediasoup-api.ts`. I'll leave it as an exercise for future-us to pull it out into a library method and replace the other call sites.

imports/client/components/CallSection.tsx
zarvox: Could always have the findOne sort by createdAt descending if you want to be sure that you’re picking the most up-to-date (to force convergence faster). Otherwise, I wouldn’t worry too much about it unless we’re observing excess churn for whatever strange reason.

ebroder: Yeah I got worried when I realized the race, but having reflected some, I think I'm not actually that worried. I'll edit this comment into a note but I feel OK taking no other action.

imports/server/mediasoup-api.ts
```diff
@@ -0,0 +1,391 @@
...
+  check(tab, String);
+
+  if (!this.userId) {
+    throw new Meteor.Error(401, 'Not logged in');
+  }
+
+  if (Flags.active('disable.webrtc')) {
+    throw new Meteor.Error(403, 'WebRTC disabled');
+  }
+
```
zarvox, line 107: For completeness: maybe we should check that user `this.userId` is a member of hunt `hunt` or throw here

imports/server/mediasoup-api.ts
```diff
@@ -0,0 +1,391 @@
...
+});
+
+Meteor.publish('mediasoup:metadata', function (hunt, call) {
+  check(hunt, String);
+  check(call, String);
+
+  if (!this.userId) {
+    throw new Meteor.Error(401, 'Not logged in');
+  }
+
```
zarvox, line 91: Also for completeness: check hunt membership here too

zarvox (approved):
All right, let's do this!

Thank you for putting this all together and thinking through not just how to make this work well, but also how to try to make this fail and recover well too.

PR #487, opened 2021-12-24
Fix lock preemption

ebroder:
The logic in aa05df90 for lock preemption was backwards (because the
logic is backwards relative to everywhere else we catch a duplicate key
error). This caused us to throw the duplicate key error (and in the
event of any other error, we'd try to preempt the lock 😬)


zarvox (approved):
😬

PR #492, opened 2021-12-31
Switch to using CloudWatch agent for host metrics

ebroder:
I'm not wild about CloudWatch in general as a service, and this does
theoretically cost us some metrics that the Datadog agent is capable of
capturing, but we don't make that much use of those metrics today, and
the CloudWatch agent uses significantly less memory than Datadog.


ebroder:
(This is already running in prod because, again, it's the easiest way to test CFN changes)

zarvox (approved):

PR #498, opened 2021-12-31
Switch to host networking in CloudFormation config

ebroder:
Now that we're running RTC traffic directly through the jolly-roger
container, we should move it to host networking to save on overhead. The
easiest way to make this work was to just switch to the standard nginx
container and provide our own nginx config (which is mostly taken from
the config that nginx-proxy was previously rendering).

(Running in production successfully already)

### DIFF
--- a/cloudformation/jolly-roger.yaml
+++ b/cloudformation/jolly-roger.yaml
@@ -507,7 +507,6 @@ Resources:
                 - containerd.io
                 - htop
                 - moreutils
-                - nvme-cli
                 - python3-dev
                 - python3-pip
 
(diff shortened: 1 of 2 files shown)


cloudformation/jolly-roger.yaml
```diff
@@ -519,22 +518,80 @@ Resources:
...
                     real_ip_header X-Forwarded-For;
 
-                    location /healthcheck {
-                      return 200 "OK\n";
+                    upstream jolly-roger {
+                      server 127.0.0.1:3000;
+                    }
+
+                    server {
+                      server_name deathandmayhem.com;
```
zarvox, line 579: do you mean `${AppUrl}` (or whatever the appropriate substitution would be) here?

zarvox (approved):
Seems reasonable.  We probably want to be templating the nginx config to not tie it to our particular URL, which I think unfortunately means converting all the `$` to `$$`, and then writing `${AppUrl}` in that one spot?

But you can merge this as-is for now

PR #500, opened 2022-01-01
Fix receiving audio on non-Chrome browsers

ebroder:
Apparently support for the MediaStream active event is...spotty. So
instead, have ConsumerManagers explicitly call back into PeerBox with
tracks when they get them, and have a callback in PeerBox manage both
adding/removing tracks from the stream and tracking when there are a
non-0 number of tracks.

### DIFF
--- a/imports/client/components/CallSection.tsx
+++ b/imports/client/components/CallSection.tsx
@@ -247,25 +247,19 @@ const ProducerBox = ({
 };
 
 const ConsumerManager = ({
-  stream,
+  setTrack,
   recvTransport,
   serverConsumer,
 }: {
-  stream: MediaStream;
+  setTrack: (consumer: string, track?: MediaStreamTrack) => void;
   recvTransport: types.Transport,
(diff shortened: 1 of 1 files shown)


imports/client/components/CallSection.tsx
```diff
@@ -326,28 +320,33 @@ const PeerBox = ({
...
+  const audioRef = useRef<HTMLAudioElement>(null);
+  useEffect(() => {
+    if (audioRef.current && tracks.size > 0) {
+      audioRef.current.srcObject = stream;
     }
-
-    // eslint-disable-next-line no-param-reassign
-    audio.srcObject = stream;
-  }, [streamActive, stream]);
+  }, [audioRef, tracks, stream]);
```
zarvox, line 349: ref objects never change, so you don't need to list audioRef in the deps here

ebroder: fixed

zarvox (approved):

PR #504, opened 2022-01-02
Show Discord avatars in ProfileList and search over Discord handle/disambiguator

zarvox:
This makes the two profile list pages slightly more usable by including
recognizable avatars (when present) and also allowing the filter-search
to search over the user's linked Discord handle and disambiguator. This
should help make it easier to identify the jolly-roger user associated
with a Discord user if their Discord profile and handle lack useful
identifying information.

Before:

Screen Shot 2022-01-02 at 12 51 58 AM

After:

Screen Shot 2022-01-02 at 12 51 24 AM


ebroder:
Followup task here is probably to switch to showing Discord avatar on the profile page instead of gravatar (since we never ended up using gravatar anywhere else, we can just kill off that code completely)

ebroder (approved):

PR #507, opened 2022-01-02
Rototill the build to be entirely Docker driven

### LINKED ISSUE #494: Update the Docker image build
The Dockerfile was originally written for Docker Hub, which doesn't support any layer caching between builds. As such, it doesn't bother splitting up steps usefully. Now that we're using GitHub actions (which has a useful [caching primitive](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#registry-cache)) and have a significant native build component, we should put some love into the Docker build.

I'd say there are a few things we probably want to do:
- Switch from the `elgohr/Publish-Docker-Github-Action` action to the official Docker `docker/build-push-action` (I'm not sure why I used the former in the first place)
- Switch from Docker Hub to the [GitHub Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry), for a few reasons: (a) Docker Hub is an increasingly unpleasant place to host things (b) GitHub has native token authentication so we can stop passing a username/password around
- Update the `Dockerfile` to have separate steps with useful caching and use a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds) to ensure that doesn't inflate the final build product. (If we do this right, we should actually be able to make the image _smaller_, since right now we're including build-time dependencies we don't need, like Meteor itself)

ebroder:
Our build configuration was originally written before multi-stage Docker builds and for Docker Hub. Since Docker Hub had no caching of intermediate images and we wanted to generate the smallest images possible, there was no advantage to writing a Dockerfile that was cacheable.

Now that we have a native dependency with a heavyweight build, we would benefit significantly from a build that could cache intermediate steps, and with GitHub Actions and multi-stage builds, we can do that.

Switch to the docker/build-push-action, which among other things, supports the GitHub Actions Cache API, and use that for both test and build. Additionally, switch to the GitHub Container Registry, which allows us to eliminate a shared secret for authentication (since GitHub provides an authentication token as part of the GitHub Actions workflow). Finally, enabling the new docker/setup-buildx-action means that we can lightly parallelize the two stages of the build.

Fixes #494.


ebroder:
To verify that the caching worked correctly, I made a change to just the `.github/workflows/build.yml` file. You can verify the two builds here: [first build](https://github.com/deathandmayhem/jolly-roger/runs/4686164478?check_suite_focus=true), [second build](https://github.com/deathandmayhem/jolly-roger/runs/4686218374?check_suite_focus=true).

Since `.github/workflows` is excluded from the Docker build by `.dockerignore`, I expected the second build to be a complete noop relative to the first. However, it seems that when Docker builds using a [git repository as build context](https://docs.docker.com/engine/reference/commandline/build/#git-repositories), it doesn't apply `.dockerignore` correctly, so I'm going to switch to using a file context.

In any case, it does seem that we're able to take advantage of caching up to the point that we copy the full repository into the container (`Dockerfile:24`), meaning that we start running the lint step about 70 seconds into the build. In other words, we should continue to see lint failures occur relatively quickly, even if the build takes a long time to finish. Much of the delay to that point seems to be fetching the build cache from the `npm ci` step - I suspect GitHub's cache is just not all that fast.

Uploading the cache at the end of the build seems to be quite slow (taking upwards of 5 minutes), and it seems like BuildKit may upload the complete cache on every build (even intermediate images that were fetched from the cache initially).

ebroder:
Worth noting that this does drop linting on the `tests/down/` directory, which has its own `package.json` and eslint rules. I don't feel too bad about that, because I'm somewhat inclined to drop `tests/down/` anyway. I haven't used that to evaluate jolly-roger's load capacity since we were initially rolling it out, and it is a pretty brittle test suite (since it only works if you accurately capture the set of subscriptions that a client makes)

.github/workflows/build.yml
```diff
@@ -1,18 +1,51 @@
...
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - name: Build and test Docker image
+      id: build
+      uses: docker/build-push-action@v2
       with:
-        name: deathandmayhem/jolly-roger
-        username: ${{ secrets.DOCKER_USERNAME }}
-        password: ${{ secrets.DOCKER_PASSWORD }}
+        context: .
+        push: ${{ github.ref == 'refs/heads/main' }} # only push on main builds
```
ebroder, line 45: Note that this only runs on `main` so it's possible this will break and we'll need to fix it later.

.github/workflows/build.yml
```diff
@@ -1,18 +1,51 @@
-name: Build Docker image
+name: Build and test Docker image
 
+# Run builds on main commits and when PRs are opened or updated
 on:
   push:
     branches:
-      - main
+    - main
+  pull_request:
```
ebroder, line 8: This is a slight change in behavior relative to the old configuration. We'll run the build when a PR is opened and when new commits are pushed to the branch after the PR is opened, but we won't run builds on a branch that doesn't have a corresponding PR.

I think this is desirable because we no longer run every build twice once you have opened a PR, and it should be easy enough to adapt to by just opening draft PRs aggressively.

Dockerfile
```diff
@@ -1,11 +1,79 @@
...
+RUN --mount=type=cache,target=/root/.npm meteor npm install --production
+
+# Production image
+# (Be careful about creating as few layers as possible)
+
+FROM ubuntu:18.04
+
+# Install runtime deps
+RUN <<EOF
+  . /etc/os-release
```
ebroder, line 49: FTR, I would normally use `lsb_release` here, but it's not installed in the Ubuntu docker container.

zarvox: `/etc/os-release` is pretty darn guaranteed to exist at this point: https://www.freedesktop.org/software/systemd/man/os-release.html

ebroder: Ah good. I was wondering if sourcing the file was a bad idea. Sounds like that's an explicit design goal

zarvox (approved):

PR #513, opened 2022-01-03
Generate a latest tag on main

ebroder:


ebroder:
Under the default configuration for `docker/metadata-action@v3`, we generate an image tagged with `main` but not one tagged with `latest`. This adds the latest tag as well.

(It's slightly redundant to have the `github.ref == 'refs/heads/main'` check on both the metadata-action and the build-push-action, but it feels more future-proof.)

zarvox (approved):
Read up on this behavior at https://github.com/docker/metadata-action#flavor-input and it seems reasonable

PR #514, opened 2022-01-03
Enable ICMP echo request/reply and frag needed on ALB

ebroder:
This makes it possible to ping deathandmayhem.com directly.


ebroder:
(Including frag needed on top of echo request/reply based on https://www.daemonology.net/blog/2012-11-28-broken-EC2-firewall.html)

zarvox (approved):

PR #515, opened 2022-01-03
Allow `expectedAnswerCount` to be zero

### LINKED ISSUE #512: Allow `expectedAnswerCount` to be 0 and use it for administrivia
We previously special-cased the `administrivia` tag to hide things that don't make sense for "puzzles" without any expected answers.  Now that we actually track how many answers each puzzle expects, we should

1) allow `expectedAnswerCount` to be 0
2) make the special casing also trigger when `expectedAnswerCount` == 0
3) mark administrivia puzzles as expecting 0 answers
4) (optional, if we update all past administrivia puzzles) remove the special handling for `administrivia` (leave in the pin-to-top behavior)

zarvox:
For puzzles where expected answer count is exactly zero:

* On the puzzle page, don't show the answer submission/guess queue
  submission button or modal
* On the puzzle list page, consider the item eligible to be shown
  alongside unsolved puzzles, since these are usually pinned
  administrivia that we don't want to disappear.
* On the puzzle list page, don't show viewer counts
* On the puzzle list page, style the puzzle row in light purple (as we
  currently do for administrivia)

We retain the sort order behavior of the `administrivia` tag (sorts
above all other groups).

I audited the remaining references to the string `administrivia` as well
as the `expectedAnswerCount` field; they all look good.

Fixes #512.

### DIFF
--- a/imports/client/components/Puzzle.tsx
+++ b/imports/client/components/Puzzle.tsx
@@ -68,13 +68,13 @@ const Puzzle = React.memo((props: PuzzleProps) => {
   const tagIndex = _.indexBy(props.allTags, '_id');
   const shownTags = _.difference(props.puzzle.tags, props.suppressTags || []);
   const ownTags = shownTags.map((tagId) => { return tagIndex[tagId]; }).filter(Boolean);
-  const isAdministrivia = props.puzzle.tags.find((t) => { return tagIndex[t] && tagIndex[t].name === 'administrivia'; });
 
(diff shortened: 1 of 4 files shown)


zarvox:
Note that this will require us to go back through past year's administrivia to update their `expectedAnswerCount` values to be 0.  That didn't seem worth a migration, since it's a mere handful of administrivia items to fix.

ebroder (approved):

PR #518, opened 2022-01-04
Add FirehosePage for tailing/searching all chat messages

zarvox:
This page adds an in-app view of the sort that we historically aimed for
from the firehose in IRC, Slack, or Discord, where we could see all chat
messages on all puzzles in a single interleaved time-series view.

I'm ignoring formatting here for now; this is largely to make searching
over the whole of a hunt's chatlogs easier.

The filter search only looks at the message text; if you wanted to
filter on a single puzzle, just go view the log from the puzzle page.

### DIFF
--- a/imports/client/components/FirehosePage.tsx
+++ b/imports/client/components/FirehosePage.tsx
@@ -1,0 +1,266 @@
+import { Meteor } from 'meteor/meteor';
+import { useTracker } from 'meteor/react-meteor-data';
+import { _ } from 'meteor/underscore';
+import { faEraser } from '@fortawesome/free-solid-svg-icons/faEraser';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import React, {
+  useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState,
+} from 'react';
+import Button from 'react-bootstrap/Button';
+import FormControl, { FormControlProps } from 'react-bootstrap/FormControl';
+import FormGroup from 'react-bootstrap/FormGroup';
+import InputGroup from 'react-bootstrap/InputGroup';
+import { Link, useParams, useSearchParams } from 'react-router-dom';
+import styled from 'styled-components';
+import { shortCalendarTimeFormat } from '../../lib/calendarTimeFormat';
+import ChatMessages from '../../lib/models/chats';
+import Profiles from '../../lib/models/profiles';
+import Puzzles from '../../lib/models/puzzles';
+import { ChatMessageType } from '../../lib/schemas/chat';
+import { PuzzleType } from '../../lib/schemas/puzzle';
+import { useBreadcrumb } from '../hooks/breadcrumb';
+import FixedLayout from './styling/FixedLayout';
+
+const FirehosePageLayout = styled.div`
+  padding: 8px 15px;
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  justify-content: flex-start;
+  height: 100%;
+  > * {
+    width: 100%;
+  }
+`;
+
+interface MessageProps {
+  msg: ChatMessageType;
+  displayNames: Record<string, string>;
+  puzzles: Record<string, PuzzleType>;
+}
+
+const Message = ({ msg, displayNames, puzzles }: MessageProps) => {
+  const displayName = msg.sender ? displayNames[msg.sender] : 'jolly-roger';
+  const ts = shortCalendarTimeFormat(msg.timestamp);
+  return (
+    <div>
+      [
+      {ts}
+      ] [
+      <Link to={`/hunts/${msg.hunt}/puzzles/${msg.puzzle}`}>{puzzles[msg.puzzle].title}</Link>
+      {'] '}
+      {displayName}
+      {': '}
+      {msg.text}
+    </div>
+  );
+};
+
+const MessagesPane = styled.div`
+  overflow-y: scroll;
+  flex: 1;
+
+  &.live {
+    border-bottom: 1px solid black;
+  }
+`;
+
+const FirehosePage = () => {
+  const huntId = useParams<'huntId'>().huntId!;
+  const [searchParams, setSearchParams] = useSearchParams();
+  const searchString = searchParams.get('q') || '';
+
+  useBreadcrumb({ title: 'Firehose', path: `/hunts/${huntId}/firehose` });
+
+  const profilesTracker = useTracker(() => {
+    const profilesHandle = Profiles.subscribeDisplayNames();
+    const ready = profilesHandle.ready();
+    return {
+      ready,
+      displayNames: ready ? Profiles.displayNames() : {},
+    };
+  }, []);
+  const puzzlesTracker = useTracker(() => {
+    const puzzlesHandle = Meteor.subscribe('mongo.puzzles', {
+      hunt: huntId,
+    });
+    const ready = puzzlesHandle.ready();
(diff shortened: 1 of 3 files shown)


imports/client/components/FirehosePage.tsx
```diff
@@ -0,0 +1,267 @@
...
+import { useBreadcrumb } from '../hooks/breadcrumb';
+import FixedLayout from './styling/FixedLayout';
+
+const FirehosePageLayout = styled.div`
+  padding: 8px 15px;
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  justify-content: flex-start;
+  max-height: calc(100vh - 60px);
```
grantaelliott, line 31: This lands inside of a `.FixedLayout` that takes care of keeping clear of the Nav, so I don't see any reason to have this calc here (and eliminating it is an improvement to modularity).

Whether or not `max-height` is specified, I think the chat log should fill the available space vertically.

This suggestion addresses both:
```suggestion
  height: 100%;
```

grantaelliott, line 31: This lands inside of a `.FixedLayout` that takes care of keeping clear of the Nav, so I don't see any reason to have this calc here (and eliminating it is an improvement to modularity).

Whether or not `max-height` is specified, I think the chat log should fill the available space vertically.

This suggestion addresses both:
```suggestion
  height: 100%;
```

imports/client/components/PuzzleListPage.tsx
```diff
@@ -502,6 +503,15 @@ const PuzzleListPage = (props: PuzzleListPageWithRouterParams) => {
             <StyledPuzzleListLinkLabel>Hunters</StyledPuzzleListLinkLabel>
           </StyledPuzzleListLinkAnchor>
         </StyledPuzzleListLink>
+        {/* Show firehose link only to operators */}
+        {tracker.canUpdate && (
+          <StyledPuzzleListLink>
+            <StyledPuzzleListLinkAnchor to={`/hunts/${props.match.params.huntId}/firehose`}>
+              <FontAwesomeIcon icon={faCloudShowersHeavy} />
```
grantaelliott, line 510: I'm fine with this choice, but did you also consider `faucet`?

imports/client/components/FirehosePage.tsx
```diff
@@ -0,0 +1,267 @@
...
+      {displayName}
+      {': '}
+      {msg.text}
+    </div>
+  );
+};
+
+const MessagesPane = styled.div`
+  overflow-y: scroll;
+  max-height: 100vh;
```
grantaelliott, line 62: Is this necessary? I don't think it does anything.
```suggestion
```

imports/client/components/FirehosePage.tsx
```diff
@@ -0,0 +1,267 @@
...
+      const isInteresting = compileMatcher(searchKeysWithEmptyKeysRemoved);
+      interestingChatMessages = chatMessages.filter(isInteresting);
+    }
+
+    return interestingChatMessages;
+  }, [getSearchString, compileMatcher]);
+
+  const [shouldScrollBottom, setShouldScrollBottom] = useState<boolean>(true);
+
+  const saveShouldScroll = useCallback(() => {
```
grantaelliott, line 190: ShouldScrollBottom is not consistently set when a filter change causes the length of the list to increase and overflow, forcing the bottom off the list. (The preferred behavior in this case, since the bottom was completely visible prior, may actually be to preserve the state and advance the scroll.)

imports/client/components/FirehosePage.tsx
```diff
@@ -0,0 +1,267 @@
...
+                key={msg._id}
+                msg={msg}
+                puzzles={puzzlesTracker.puzzles}
+                displayNames={profilesTracker.displayNames}
+              />
+            );
+          })}
+        </MessagesPane>
+        <div>
+          {shouldScrollBottom ? 'Log will update to show new messages as they arrive' : ''}
```
grantaelliott, line 259: This doesn't seem to match actual behavior. When fully scrolled to the bottom, scrolling does not adjust to accommodate a new message.

grantaelliott, line 259: I found it took me a while to understand what this presentation was conveying. In fact, I initially thought the disappearing text was a glitch. I would prefer a persistent display of state, such as "Live scrolling: [On/Off]".

Frankly, since the behavior is pretty intuitive and consistent with puzzle page chat histories, I'm not sure I see the need for this message at all.

grantaelliott (changes requested):
I've mostly just tested and haven't had an opportunity to look at the code in detail yet. I left a couple of inline suggestions.

I really like the push to move out-of-channel workflows back into Jolly Roger.

imports/client/components/FirehosePage.tsx
zarvox: Sure, that seems reasonable.  I originally added it with the intent of making it a little more visible if you weren't actually following the tail of the log (and might have been expecting it to) and then missed something, but I suppose if we do a good enough job with the behavior it's not needed.

zarvox: Yeah, somehow I was missing the actual core logic for this.  😓  Should be fixed now, and should also handle window resize.

imports/client/components/FirehosePage.tsx
zarvox: Yeah, my code here was not right.  I like the point you make: if the end of the scrollback is visible, we should ensure it stays visible.  So:

* if content to display is shorter than the pane, `shouldScrollBottom` should get set to `true`.
* if content to display is larger than the pane, if the bottom of the pane is visible, `shouldScrollBottom` should get set to `true`.
* if content to display is larger than the pane, and the user scrolls off the bottom by more than a tiny bit, we should set `shouldScrollBottom` to `false`
* if the content to be displayed changes, or the space available to the content changes (due to e.g. window resize), apply the above rules again

imports/client/components/FirehosePage.tsx
zarvox: You're right.  I incorporated the patch in the rebase.

imports/client/components/PuzzleListPage.tsx
zarvox: Faucet is good too; I didn't really have much of an opinion here.  Switched to faucet.

imports/client/components/FirehosePage.tsx
zarvox: You're right, this is better.  Incorporated into the latest rev.

imports/client/components/FirehosePage.tsx
```diff
@@ -0,0 +1,267 @@
...
+  }, [ready, maybeForceScrollBottom]);
+
+  if (!ready) {
+    return <div>loading...</div>;
+  }
+
+  return (
+    <FixedLayout>
+      <FirehosePageLayout>
+        <h1>Firehose (all chat messages, expect slowness)</h1>
```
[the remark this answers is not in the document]
zarvox: Done.

zarvox:
Okay I think this actually works as intended now.  r? @grantaelliott

imports/client/components/FirehosePage.tsx
grantaelliott: I can confirm this is fixed. The nature of scroll preservation when the filter is changing and the bottom isn't visible is not transparent, but it seems to behave reasonably, even returning to the same position if the filter is shortened, then restored. I'm not even sure specifically what the desired behavior would be, so I think it's good.

grantaelliott (approved):
This looks great! Thanks for the changes.

PR #520, opened 2022-01-04
Minor build optimizations + test suite

ebroder:


zarvox (approved):
LGTM, but the debathena repo was 503ing when the checks ran, so maybe trigger them again

PR #521, opened 2022-01-05
Rework chat scroll position behavior

### LINKED ISSUE #427: Chat text moves without new messages appearing

### LINKED ISSUE #428: CallSection height does not change on sidebar resize
When its width is changed (by adjusting the sidebar divider), CallSection's height does not change to properly accommodate the grid of viewers and callers.

zarvox:
Most of the time, users are interested in the most recent chunk of chat
history, and think of the backlog as being pinned to the bottom, rather
than the top.  However, in the web APIs, overflow definitely defaults to
considering space from the top-left corner of a rectangle.

To approximate the "fixed distance from bottom of scrollback" behavior,
we save the distance from the bottom that we're aiming for, and have a
function which jumps to the appropriate scrollTop value to maintain that
distance-from-bottom which we call in response to a number of possible
ways that might cause the space available to the chat history pane (or
the space demanded by the contents to be displayed) to change:

* Callers or viewers are collapsed or expanded
* New callers or viewers join or leave, causing their boxes to take up a
  different number of rows
* Other transient RTC warnings are displayed
* The sidebar is resized (via the SplitPanePlus drag)
* The browser window as a whole is resized
* The chat input box grows or shrinks due to multiline input
* A new chat message arrives

There's one noteworthy exception: if you scroll partway back in the chat
backlog, you probably don't actually want the chat jumping around as the
scroll height of the chat history grows in response to new messages
coming in, since that makes it hard to read.  So when the message log
changes, if we were scrolled away from the bottom, we instead update the
target distance from bottom.  (When we send a chat message, we do wish
to show it in the backscroll to avoid user confusion, so we jump to the
bottom.)

We were missing a number of possible "size available has changed" event
sources -- nearly every render change in ChatPeople can cause its height
to change, and we also failed to keep the history pinned on sidebar
resize.  This change attempts to rectify these omissions.  Because we
operate on refs and wish to make DOM mutations, we should make use of
`useLayoutEffect` rather than `useEffect` anywhere we'd want to react
before the browser makes the changes visible to the user, since
otherwise we'd see flicker as `scrollTop` jumps around between frames.

<strike>There is one last situation that I have been unable to reliably fix:
when quickly shrinking the chat sidebar from a size where the chat
history does not yet overflow its container to a size where (after
reflow) it does, sometimes we will wind up failing to scroll far enough
down.  However, since the target remains "at the bottom", when any other
event comes in that affects the sidebar height, we correctly converge to
the expected behavior.  I consider this a small enough wart that I will
accept it to get the rest of the behavioral improvement.</strike> Update: this seems to be fixed after converting one more `useEffect` -> `useLayoutEffect` and splitting out the "changes happening during drag" from "committing change after drag completion" for `SplitPanePlus`.

Fixes #427.
Fixes #428.

### DIFF
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -215,7 +215,7 @@ const ChatHistory = React.forwardRef((props: ChatHistoryProps, forwardedRef: Rea
     if (scrollBottomTarget.current > 10) {
       saveScrollBottomTarget();
     } else {
-      scrollToTarget();
+      snapToBottom();
     }
   }, [props.chatMessages.length, saveScrollBottomTarget, scrollToTarget]);
 
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -8,7 +8,7 @@ import { faPuzzlePiece } from '@fortawesome/free-solid-svg-icons/faPuzzlePiece';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import classnames from 'classnames';
 import React, {
-  useCallback, useEffect, useImperativeHandle, useRef, useState,
+  useCallback, useEffect, useImperativeHandle, useLayoutEffect, useRef, useState,
 } from 'react';
 import Alert from 'react-bootstrap/Alert';
 import Badge from 'react-bootstrap/Badge';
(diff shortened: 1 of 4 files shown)


imports/client/components/PuzzlePage.tsx
```diff
@@ -134,78 +134,93 @@ interface ChatHistoryProps {
...
+    // Whenever we rerender due to new messages arriving, make our
+    // distance-from-bottom match the previous one, if it's larger than some
+    // small fudge factor.  But if the user has actually scrolled into the backlog,
+    // don't move the backlog while they're reading it -- instead, assume they want
+    // to see the same messages in the same position, and adapt the target bottom
+    // distance instead.
+    if (scrollBottomTarget.current > 10) {
+      saveScrollBottomTarget();
+    } else {
+      scrollToTarget();
```
grantaelliott, line 218: I find this behavior a little odd. If you're scrolled to the last line, but it's slightly cut off, it will continue to be slightly cut off as new messages arrive. Since the assumption within this margin is that the user intended to be at the bottom, I think we should snap all the way to the bottom when a new message arrives.
```suggestion
      snapToBottom();
```

imports/client/components/PuzzlePage.tsx
```diff
@@ -1017,12 +1042,28 @@ const PuzzlePage = React.memo((props: PuzzlePageWithRouterParams) => {
 
   const onResize = useCallback(() => {
     setIsDesktop(window.innerWidth >= MinimumDesktopWidth);
+    if (chatSectionRef.current) {
+      chatSectionRef.current.scrollHistoryToTarget();
+    }
   }, []);
 
-  const onChangeSideBarSize = useCallback((newSidebarWidth: number) => {
+  const onCommitSideBarSize = useCallback((newSidebarWidth: number) => {
```
grantaelliott, line 1050: Good change. That's more clear.

imports/client/components/SplitPanePlus.tsx
```diff
@@ -262,11 +266,15 @@ const SplitPanePlusHook = (props: SplitPanePlusProps) => {
...
   const onChange = useCallback((newSize: number) => {
     // Setting dragInProgress in onDragStarted creates a frame of strangeness
+    const nextCollapse = calculateCollapse(newSize);
     setState({
       dragInProgress: true,
-      collapseWarning: calculateCollapse(newSize),
+      collapseWarning: nextCollapse,
     });
-  }, [calculateCollapse]);
+    if (onChanged) {
```
grantaelliott, line 274: Thanks for switching this to use a new handler. It's a good addition too.

grantaelliott (approved):
This looks great! I left a suggestion for a minor behavior change when almost but not quite at the bottom.

The only glitchy behavior I've noticed (testing in Safari) occurs when a drag of the divider causes a chat comment to wrap to a different number of lines. This can (but doesn't always) cause the scroll to flicker for a single frame before landing in the right place. While not a show-stopper, if you have a sense of how to fix this, I think it'd be worth while because long chat longs will have many potential wrap transitions and this case will arise frequently. Interestingly, I believe the flicker does not happen when a resize of the window causes the sidebar to shrink (due to the right side's minimum width) past a wrap transition.

zarvox:
I can reproduce the single-frame oddness in both Firefox and Safari (though curiously not in Chrome, for reasons that I did not bother trying to understand deeply).  I spent several hours trying to figure out what exactly is causing the issue in Firefox and Safari.  While the tools available to me in Safari are limited, I think the answer is probably one of unfortunate ordering and batching between React, the browser, react-split-pane, and us, in a way that I've not yet found a reliable fix for.  Basically: I don't think any of our code is getting to run at exactly the right time to react to react-split-pane changing the `style` prop on the chat `Pane` from something like `width: 293` to `width: 292`.

What happens when we drag the resize handle?

* the mousedown event fires, and `SplitPane`'s `onMouseDown` calls `onTouchStart` which sets `SplitPane`'s `active` to true
* mousemove events fire, and on each, `onMouseMove` calls `onTouchMove` which measures the movement in the appropriate dimension (X, in this case), ultimately calling setState on `position` first, calling the `onChange` callback (provided in props by `SplitPanePlus`), and then calling setState on `pane1Size`, which triggers a rerender of the `Pane` which now has a smaller `width`.

Key thing to note here: we get the onChange callback from `SplitPane` *before* the rerender of `Pane` with a 1-pixel-narrower pane size occurs.  So we call scrollToTarget, but the scroll area hasn't actually gotten narrower (and thus, longer) yet, and we measure the node and it hasn't changed yet.

* Since SplitPane's setState was called above, React queues a rerender of `SplitPane`.  When the rerender completes, its first child `Pane` has a `style` with a smaller `width`.  Upon DOM flush, this causes the browser to recompute layout.  The browser leaves `scrollTop` as is, which causes content to appear to shift relative to the bottom of the chat history.

* Because humans are typically not pixel-perfect machines, the mouse moves another pixel and the whole thing goes again, except this time the `scrollToTarget` calls see that `scrollTop` is not at target and adjusts.  The browser updates, and the content snaps back to the bottom.

* Or alternately, in the presence of very precise movement, eventually the user releases the mouse button, which triggers SplitPanePlus's `onDragFinished` callback, which causes PuzzlePage's `onCommitSidebarSize` to call `setSidebarWidth`, which ultimately triggers the `useLayoutEffect` that calls `scrollHistoryToTarget`.  Either way, we converge to correct appearance quickly.

So the crux of the issue is: we want to run code at the time that a layout effect would run in SplitPane, except that SplitPane is a class-based component (so I guess we'd want to run during `componentDidUpdate`) and also doesn't provide a hook point at the right moment.  We can't win here -- if we were to call `setState` before SplitPane calls setState on `pane1Size`, then we can still trigger the "try to scroll to bottom" too early.  We have no way to reliably execute code after SplitPane has rerendered but before the browser renders the next frame.

This is something where I think we can't just bolt on additional cleverness in SplitPanePlus.  To resolve this, I think we'd have to patch react-split-pane (which wouldn't be too bad, but at the point where we're patching it we almost might as well rewrite it).

I think (given the implications) I'm going to merge this now as-is to make sure we have something mostly workable, and either you or I can take another pass at fixing the one-frame flicker on reflow by reworking SplitPane in a separate PR if we want.

grantaelliott:
Thanks for looking into (and documenting) that - I didn’t intend for you to spend hours on it. I agree with the assessment that it should be merged as is.

zarvox:
Nah, that's my own fault for nerd-sniping myself 😄  -- I really wanted to understand what was causing the behavior, since I didn't feel like a had a deep grasp of why things were behaving the way they were, and now I have a sense for what it would take to fix it.

PR #522, opened 2022-01-05
Use our own breadcrumb implementation

ebroder:
Upgrading Bootstrap caused a bug because they began floating the
separator. This works if you're using display: flex, but we were
overriding with display: block to get better wrapping behavior. But
given that we were overriding Bootstrap's styling so heavily, we weren't
getting any benefit from using the Bootstrap classes in the first place,
so just switch to our own version.

Verified that this seems to do the right thing at small window sizes on Chrome, Firefox, and Safari.


zarvox (approved):
Tested this out locally; looks and feels correct.  Thanks for the quick response!

PR #525, opened 2022-01-06
Enable dependabot non-security updates

ebroder:
We currently get security updates from dependabot automatically. This will additionally start opening PRs for non-security updates to dependencies, which hopefully will reduce the frequency with which I have to do this myself.

(And given our acceptance tests, we have some bare minimum confirmation that the site doesn't totally fail to load)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


ebroder:
(There's also currently an outstanding patch release for `fp-ts`, so we should be able to verify this is working)

zarvox (approved):

PR #527, opened 2022-01-06
Fix tests in CI build

ebroder:
When I split the build and test steps into separate Docker stages, I
inadvertently bypassed running the tests at all, because nothing from
the test stage was copied into the final production stage. Docker was
smart enough to skip those steps entirely.

I want to force them to run without un-parallelizing the test + build
steps, so make sure we use the output of the test step by copying the
scripts directory out of it. It's a bit of a roundabout way to do
things, since we don't actually change the scripts directory at all in
that stage, but it creates the dependency we care about.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,13 @@ ENV DEBIAN_FRONTEND noninteractive
 RUN <<EOF
   set -eux
   apt-get update
-  apt-get install --no-install-recommends -y curl python3 python3-pip python3-dev python3-setuptools python3-wheel build-essential git
+  apt-get install --no-install-recommends -y curl gnupg python3 python3-pip python3-dev python3-setuptools python3-wheel build-essential git
+
+  # Install chromium-browser's dependencies, which should match puppeteer's
+  # dependencies. Note: this will need to be updated when we upgrade to 20.04,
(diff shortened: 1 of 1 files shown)


ebroder:
You can see starting around https://github.com/deathandmayhem/jolly-roger/runs/4732817484?check_suite_focus=true#step:6:573 that (a) the test suite is running (b) it's running in parallel with the build, so I think this has the behavior I want.

Dockerfile
```diff
@@ -83,7 +90,8 @@ RUN <<EOF
 EOF
 
 COPY --from=build /built_app /built_app
-COPY scripts /built_app/scripts
+# Copy from test to force a dependency and ensure that the tests run
```
zarvox, line 93: 😂 

Would it make sense to have the `test` image create some sort of "yeah these tests were passed" manifest to copy in here?

zarvox (approved):

Dockerfile
ebroder: Hmm. It seems like `meteortesting:mocha` can [generate an xunit report](https://github.com/Meteor-Community-Packages/meteor-mocha#write-tests-to-a-file)?

PR #541, opened 2022-01-07
Spacing options for PuzzleAnswer

grantaelliott:
Adds a convenience feature which standardizes spacing in related puzzle answers for spotting patterns. A toggle on the related puzzle popover switches between preserving original spacing and removing all spaces (displayed segmented into groups of five characters for easy indexing). The copy button is aware of the spacing setting and copies answers to the clipboard without spaces as appropriate. (Whether using the copy button or highlighting and copying, the gaps between segments are for display only and do not add space characters in the copied text.)

Historically, hunters and operators have been inconsistent about spaces in answers. Hopefully, display flexibility will encourage us to standardize on preserving natural spaces.

Respace off:
original

Respace on:
segmented

### DIFF
--- a/imports/client/components/PuzzleAnswer.tsx
+++ b/imports/client/components/PuzzleAnswer.tsx
@@ -14,7 +14,18 @@ const PuzzleAnswer = React.memo((props: PuzzleAnswerProps) => {
   const respacedAnswer = respace ? props.answer.replace(/\s+/g, '') : props.answer;
   let formattedAnswer : React.ReactNode = respacedAnswer;
   if (respace && segmentSize > 0) {
-    const segments = respacedAnswer.match(new RegExp(`.{1,${segmentSize}}`, 'g')) || [];
+    // Use Intl.Segmenter (stage 3 proposal) if available to properly segment grapheme clusters
+    // Typescript is unaware of it, so there are a few any casts...
+    let graphemes:string[];
+    if (Intl !== undefined && (Intl as any).Segmenter !== undefined) {
+      const graphemeSegmenter = new (Intl as any).Segmenter('en', { granularity: 'grapheme' });
+      graphemes = Array.from(graphemeSegmenter.segment(respacedAnswer), (s) => (s as any).segment);
+    } else {
+      graphemes = Array.from(respacedAnswer);
+    }
+    const segments = Array.from(new Array(Math.ceil(graphemes.length / segmentSize)), (_x, i) => {
+      return graphemes.slice(i * segmentSize, (i + 1) * segmentSize);
+    });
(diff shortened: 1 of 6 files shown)


imports/client/components/PuzzleAnswer.tsx
```diff
@@ -2,14 +2,31 @@ import React from 'react';
...
 }
 
-const PuzzleAnswer = (props: PuzzleAnswerProps) => {
+const PuzzleAnswer = React.memo((props: PuzzleAnswerProps) => {
+  const respace = props.respace ?? false;
+  const segmentSize = Math.floor(props.segmentSize ?? 5);
+  const respacedAnswer = respace ? props.answer.replace(/\s+/g, '') : props.answer;
+  let formattedAnswer : React.ReactNode = respacedAnswer;
+  if (respace && segmentSize > 0) {
+    const segments = respacedAnswer.match(new RegExp(`.{1,${segmentSize}}`, 'g')) || [];
```
zarvox, line 17: I pray we never have to respace emoji answers that get split in the middle of ZWJ sequences, but honestly at that point you probably aren't trying to do character alignment

zarvox (approved):
This is neat!  Thanks!

imports/client/components/PuzzleAnswer.tsx
grantaelliott: Good catch.

Proper grapheme segmentation turns out to be surprisingly tricky. The proposed Intl.Segmenter is clearly the right answer and already has wide support. It looks like there are polyfills available, but I don't think that's really worth it. This suggestion just falls back to the character level if Segmenter isn't available.

Of course, emoji don't render in fixed width, so this doesn't help much for alignment purposes, but the chunking still makes indexing easier. And at least emoji are rendered correctly when respacing (which might still be useful for other answers in the group) is on.

Unicode is so complicated...

```suggestion
    // Use Intl.Segmenter (stage 3 proposal) if available to properly segment grapheme clusters
    // Typescript is unaware of it, so there are a few any casts...
    let graphemes:string[];
    if (Intl !== undefined && (Intl as any).Segmenter !== undefined) {
      const graphemeSegmenter = new (Intl as any).Segmenter('en', { granularity: 'grapheme' });
      graphemes = Array.from(graphemeSegmenter.segment(respacedAnswer), (s) => (s as any).segment);
    } else {
      graphemes = Array.from(respacedAnswer);
    }
    const segments = Array.from(new Array(Math.ceil(graphemes.length / segmentSize)), (_x, i) => {
      return graphemes.slice(i * segmentSize, (i + 1) * segmentSize);
    });
```

PR #542, opened 2022-01-08
Narrow role/channel listings to selected guild

ebroder:
If the Jolly Roger Discord bot has been added to more than one guild,
channels and roles from all of those guilds will be available in the
Discord cache for selection when creating/editing a hunt, which is not
the behavior we want.

This doesn't come up often in practice because generally Jolly Roger
runs in a single-guild setup, but this gives us future proofing (and
also lets us slightly simplify our role logic).

### DIFF
--- a/imports/client/components/HuntListPage.tsx
+++ b/imports/client/components/HuntListPage.tsx
@@ -1,6 +1,7 @@
 import { Meteor } from 'meteor/meteor';
-import { useTracker } from 'meteor/react-meteor-data';
+import { useSubscribe, useTracker } from 'meteor/react-meteor-data';
 import { faEdit } from '@fortawesome/free-solid-svg-icons/faEdit';
+import { faInfo } from '@fortawesome/free-solid-svg-icons/faInfo';
 import { faMinus } from '@fortawesome/free-solid-svg-icons/faMinus';
 import { faPlus } from '@fortawesome/free-solid-svg-icons/faPlus';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
@@ -21,9 +22,10 @@ import { Link } from 'react-router-dom';
 import Ansible from '../../ansible';
 import DiscordCache from '../../lib/models/discord_cache';
 import Hunts from '../../lib/models/hunts';
+import Settings from '../../lib/models/settings';
 import { userMayCreateHunt, userMayUpdateHunt } from '../../lib/permission_stubs';
 import { HuntType, SavedDiscordObjectType } from '../../lib/schemas/hunt';
-import { DiscordChannelType, DiscordRoleType } from '../discord';
+import { SettingType } from '../../lib/schemas/setting';
 import { useBreadcrumb } from '../hooks/breadcrumb';
 import ModalForm, { ModalFormHandle } from './ModalForm';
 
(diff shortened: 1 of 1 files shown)


imports/client/components/HuntListPage.tsx
```diff
@@ -106,26 +107,26 @@ const DiscordSelector = (props: DiscordSelectorProps) => {
...
 };
 
-const DiscordChannelSelector = (params: DiscordSelectorParams) => {
+const DiscordChannelSelector = (params: DiscordSelectorParams & { guildId?: string }) => {
   const { ready, options } = useTracker(() => {
     const handle = Meteor.subscribe('discord.cache', { type: 'channel' });
-    const discordChannels: SavedDiscordObjectType[] = DiscordCache.find(
+    const discordChannels: SavedDiscordObjectType[] = DiscordCache.find({
+      type: 'channel',
+      'object.guild': params.guildId,
```
zarvox, line 115: Talk me through what you expect to happen if `guildId` is `undefined` here.  Does it omit that field from the filter, resulting in an enumeration of *all* text channels that the bot has access to?  Or does it only match values where `object.guild` is `undefined`?

(I guess while we're here, what do you want the behavior here to be if no guild has been selected?  I feel like guild should maybe be required if we're going to attempt to make this more scoped)

imports/client/components/HuntListPage.tsx
```diff
@@ -135,39 +136,28 @@ const DiscordChannelSelector = (params: DiscordSelectorParams) => {
...
 
-const DiscordRoleSelector = (params: DiscordSelectorParams) => {
+const DiscordRoleSelector = (params: DiscordSelectorParams & { guildId?: string }) => {
   const { ready, options } = useTracker(() => {
     const handle = Meteor.subscribe('discord.cache', { type: 'role' });
-    const discordRoles: SavedDiscordObjectType[] = DiscordCache.find(
-      { type: 'role' },
+    const discordRoles: SavedDiscordObjectType[] = DiscordCache.find({
+      type: 'role',
+      'object.guild': params.guildId,
```
zarvox, line 144: Same question here -- do you want this to enumerate all roles of all guilds if guildId is unset, or do you want to require that guildId be set first?

imports/client/components/HuntListPage.tsx
ebroder: Hmm. I guess I didn't quite think through this deeply enough. In practice, the current behavior is that `undefined` acts like `null` - it matches an object where that field is explicitly set to null or unset, which in turn matches nothing (since all channel and role objects have a `guild` property):

```
Models.DiscordCache.find().count()
36
Models.DiscordCache.find({'object.guild': null}).count()
0
Models.DiscordCache.find({'object.guild': undefined}).count()
0
```

Honestly, though, I think the behavior I actually want is for that dropdown to be disabled when `guildId` is undefined, and then it mostly doesn't matter. I'll update accordingly.

zarvox: If this component isn't going to be useful for anything if `guildId` isn't specified, then we should probably just make the param required?  Unless you're hoping to keep the field usable for *unsetting* the channel if it was previously set and then the guild becomes unconfigured, but then we shouldn't disable the field?

imports/client/components/HuntListPage.tsx
```diff
@@ -462,7 +458,8 @@ const HuntModalForm = React.forwardRef((props: HuntModalFormProps, forwardedRef:
         <Col xs={9}>
           <DiscordChannelSelector
             id={`${idPrefix}puzzle-hooks-discord-channel`}
-            disable={disableForm}
+            guildId={guildId}
+            disable={disableForm || guildId === undefined}
```
zarvox, line 462: If these are disabled due to unset `guildId`, can we display some sort of information scent as to why, and what would need to be done to enable it?

ebroder: Hmm. Maybe the right behavior here is to hide these fields entirely and replace them with an `<Alert>` or something?

zarvox (approved):

PR #544, opened 2022-01-09
Prevent navbar overscroll on iOS

### LINKED ISSUE #543: Navbar can be scrolled on puzzle page in iOS Safari
In iOS Safari, attempts to scroll PuzzlePage cause the navbar to overscroll. This glitch is more than cosmetic, as the chat log subsequently cannot be scrolled down (dragged up) until it is first scrolled up (dragged down).

This was introduced when navbar was switched from fixed to sticky. There are other ways to address it (for instance giving body a fixed position when FixedLayout is used) but reverting navbar to fixed positioning is probably the most straightforward.

grantaelliott:
Suggested fix for iOS navbar overscrolling issue. Switches navbar back to fixed from sticky.

Closes #543.

### DIFF
--- a/imports/client/components/App.tsx
+++ b/imports/client/components/App.tsx
@@ -37,6 +37,10 @@ const Breadcrumb = styled.nav`
   flex: 1;
 `;
 
+const ContentContainer = styled.div`
+  margin-top: ${NavBarHeight};
+`;
+
 const BreadcrumbList = styled.ol`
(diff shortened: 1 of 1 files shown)


zarvox (approved):
Looks good!

Thanks for your constant vigilance to keep things working well on mobile, even as we keep changing things all over the place -- I really appreciate the attention to detail!

Issue #172, opened 2019-01-02
Better surface activity level on puzzles


ebroder:
We've received feedback in the past that it's hard to tell what puzzles are being actively worked on. Today, our information is fairly coarse (mostly the subscription counters, which aren't great - c.f. #107).

This ticket is fairly vague, but the general idea is to come up with some notion of which puzzles are active/inactive and surface that in the puzzle list somehow (some sort of heatmapping?). This activity metric could potentially use as inputs things like subscription counters, last chat message, etc.

PR #556, opened 2022-01-11
Track recent activity in voice chat

ebroder:
Use AudioLevelObserver + some debouncing to keep track of the last time
there was voice activity (with roughly 60 second granularity), and keep
that on a new CallHistories model (which persists across calls being
started/stopped, unlike any of the other Mediasoup models).

This data gets displayed in the call section if you're not currently
joined to a call.

Here are some examples of what that might look like:

Screen Shot 2022-01-10 at 8 22 01 PM

(Note that we'll only ever see "seconds ago" in development mode. Adding the activity note to the puzzle list page is left as an exercise for future-me assuming we're happy with the rest of this.)

### DIFF
--- a/imports/client/components/ChatPeople.tsx
+++ b/imports/client/components/ChatPeople.tsx
@@ -11,10 +11,14 @@ import React, {
 import Button from 'react-bootstrap/Button';
 import OverlayTrigger from 'react-bootstrap/OverlayTrigger';
 import Tooltip from 'react-bootstrap/Tooltip';
+import styled from 'styled-components';
 import Flags from '../../flags';
+import { RECENT_ACTIVITY_TIME_WINDOW_MS } from '../../lib/config/webrtc';
 import { getAvatarCdnUrl } from '../../lib/discord';
+import CallHistories from '../../lib/models/mediasoup/call_histories';
 import Peers from '../../lib/models/mediasoup/peers';
 import Profiles from '../../lib/models/profiles';
+import relativeTimeFormat from '../../lib/relativeTimeFormat';
 import { PeerType } from '../../lib/schemas/mediasoup/peer';
 import useSubscribeAvatars from '../hooks/use-subscribe-avatars';
 import { Subscribers } from '../subscribers';
@@ -62,6 +66,11 @@ const ViewerPersonBox = ({
     </OverlayTrigger>
   );
 };
+
+const PeopleListHeader = styled.header`
+  padding-left: 1rem;
+  text-indent: -1rem;
+`;
 
 interface ChatPeopleProps {
   huntId: string;
@@ -153,6 +162,29 @@ const ChatPeople = (props: ChatPeopleProps) => {
   // anything in the UI (which prevents clients from subbing to 'mediasoup:join'
   // or doing signalling).
   const rtcDisabled = useTracker(() => Flags.active('disable.webrtc'), []);
+
+  const recentVoiceActivity = useTracker(() => (
+    CallHistories.findOne({ call: puzzleId })?.lastActivity
+  ), [puzzleId]);
+  const [voiceActivityRelative, setVoiceActivityRelative] = useState<string>();
+  useEffect(() => {
+    let interval: number | undefined;
+    if (recentVoiceActivity) {
+      const formatter = () => relativeTimeFormat(recentVoiceActivity, {
+        complete: false,
+        minimumUnit: Meteor.isDevelopment ? 'second' : 'minute',
+      });
+      setVoiceActivityRelative(formatter());
+      interval = Meteor.setInterval(() => {
+        setVoiceActivityRelative(formatter());
+      }, RECENT_ACTIVITY_TIME_WINDOW_MS);
+    }
+    return () => {
+      if (interval) {
+        Meteor.clearInterval(interval);
+      }
+    };
+  }, [recentVoiceActivity]);
 
   const {
     unknown,
@@ -409,7 +441,13 @@ const ChatPeople = (props: ChatPeopleProps) => {
     // * when joining the audiocall
     onHeightChange();
   }, [
-    onHeightChange, rtcViewers.length, viewers.length, callersExpanded, viewersExpanded, callState,
+    onHeightChange,
+    rtcViewers.length,
+    viewers.length,
+    callersExpanded,
+    viewersExpanded,
+    callState,
+    voiceActivityRelative,
   ]);
 
   if (loading) {
@@ -429,10 +467,17 @@ const ChatPeople = (props: ChatPeopleProps) => {
               <Button variant="primary" size="sm" block onClick={joinCall}>{joinLabel}</Button>
             </div>
             <div className="chatter-subsection av-chatters">
-              <header onClick={toggleCallersExpanded}>
+              <PeopleListHeader onClick={toggleCallersExpanded}>
                 <FontAwesomeIcon fixedWidth icon={callersHeaderIcon} />
                 {`${rtcViewers.length} caller${rtcViewers.length !== 1 ? 's' : ''}`}
-              </header>
+                {voiceActivityRelative && (
+                  <>
+                    {' (last voice activity: '}
+                    {voiceActivityRelative}
+                    )
+                  </>
+                )}
+              </PeopleListHeader>
               <div className={classnames('people-list', { collapsed: !callersExpanded })}>
                 {rtcViewers.map((viewer) => <ViewerPersonBox key={`person-${viewer.user}-${viewer.tab}`} {...viewer} />)}
               </div>
@@ -476,10 +521,10 @@ const ChatPeople = (props: ChatPeopleProps) => {
     <section className="chatter-section">
       {!rtcDisabled && callersSubsection}
       <div className="chatter-subsection non-av-viewers">
-        <header onClick={toggleViewersExpanded}>
+        <PeopleListHeader onClick={toggleViewersExpanded}>
           <FontAwesomeIcon fixedWidth icon={viewersHeaderIcon} />
           {`${totalViewers} viewer${totalViewers !== 1 ? 's' : ''}`}
-        </header>
+        </PeopleListHeader>
         <div className={classnames('people-list', { collapsed: !viewersExpanded })}>
           {viewers.map((viewer) => <ViewerPersonBox key={`person-${viewer.user}`} {...viewer} />)}
         </div>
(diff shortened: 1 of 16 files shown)


grantaelliott:
This is a really good idea. It feels slightly strange that only recent voice activity is called out, but I think I see why you prefer it. Is your logic that recent text chat is already timestamped and available, so only voice activity requires a separate notice? Especially since we don't have last spreadsheet activity, a top level "last activity" could be misleading as well.

ebroder:
I think that in my mind the activity indicator is localized to, like, where that activity is happening? Like, if you want to see if chat is active, look in the chat area; if you want to see if the sheet is active (to the extent that you can tell), look in the sheet area; if you want to tell if voice...etc.

I think that mostly justifies having a voice-only UI element in that location for now, but in the long term I'd definitely like to have a much more sophisticated view of "puzzle activity" than we have today. (I opened #172 a long time ago to track that, though I've made literally 0 progress on the problem.) In the interim, this allows us to surface the one piece of information that we have and isn't otherwise already inspectable.

zarvox:
I feel like the Alert on the PuzzlePage here takes up a lot of space and since it's a different color, it also draws the eye a bit more than the information it's intending to convey deserves.  I'm somewhat inclined to instead put the message under the collapsible "N callers" heading, where it's adjacent to other information about the call while also packing the data into less vertical space.  My screenshots below used:

```diff
diff --git a/imports/client/components/ChatPeople.tsx b/imports/client/components/ChatPeople.tsx
index e356617..bf4257f 100644
--- a/imports/client/components/ChatPeople.tsx
+++ b/imports/client/components/ChatPeople.tsx
@@ -8,10 +8,10 @@ import classnames from 'classnames';
 import React, {
   ReactChild, useCallback, useEffect, useLayoutEffect, useState,
 } from 'react';
-import Alert from 'react-bootstrap/Alert';
 import Button from 'react-bootstrap/Button';
 import OverlayTrigger from 'react-bootstrap/OverlayTrigger';
 import Tooltip from 'react-bootstrap/Tooltip';
+import styled from 'styled-components';
 import Flags from '../../flags';
 import { RECENT_ACTIVITY_TIME_WINDOW_MS } from '../../lib/config/webrtc';
 import { getAvatarCdnUrl } from '../../lib/discord';
@@ -122,6 +122,10 @@ function participantState(explicitlyMuted: boolean, deafened: boolean) {
   }
 }
 
+const LastVoiceActivityDiv = styled.div`
+  margin-left: 15px;
+`;
+
 // ChatPeople is the component that deals with all user presence and
 // WebRTC call subscriptions, state, and visualization.
 const ChatPeople = (props: ChatPeopleProps) => {
@@ -455,17 +459,17 @@ const ChatPeople = (props: ChatPeopleProps) => {
             <div className="av-actions">
               <Button variant="primary" size="sm" block onClick={joinCall}>{joinLabel}</Button>
             </div>
-            {voiceActivityRelative && (
-              <Alert variant="info">
-                Last voice activity:
-                {' '}
-                {voiceActivityRelative}
-              </Alert>
-            )}
             <div className="chatter-subsection av-chatters">
               <header onClick={toggleCallersExpanded}>
                 <FontAwesomeIcon fixedWidth icon={callersHeaderIcon} />
                 {`${rtcViewers.length} caller${rtcViewers.length !== 1 ? 's' : ''}`}
+                {callersExpanded && voiceActivityRelative && (
+                  <LastVoiceActivityDiv>
+                    Last voice activity:
+                    {' '}
+                    {voiceActivityRelative}
+                  </LastVoiceActivityDiv>
+                )}
               </header>
               <div className={classnames('people-list', { collapsed: !callersExpanded })}>
                 {rtcViewers.map((viewer) => <ViewerPersonBox key={`person-${viewer.user}-${viewer.tab}`} {...viewer} />)}
```

---
In local testing, I'm finding that the backend code isn't reliably updating the timestamp in the collection when I generate audio at a reasonable (or even at unreasonable) level:

image

I looked in the DB, and indeed, the record in CallHistories isn't getting updated.  Some `console.log` debugging suggests that the function inside `Meteor.bindEnvironment()` is not actually running, despite `updateLastActivity` getting called by mediasoup.  This is on my M1 MacBook Air.  Any idea what's going on here?

grantaelliott:
@zarvox My preference would be to keep the notice outside the collapsible caller list, as this information is useful even if you aren't interested in who is on the call.

ebroder:
> In local testing, I'm finding that the backend code isn't reliably updating the timestamp in the collection when I generate audio at a reasonable (or even at unreasonable) level:

Hmm. I did have problems with that earlier, but I thought I addressed by wrapping the update function. Does it work if you move the `Meteor.bindEnvironment` outside of the `_.debounce`? That's the only situation I can see where we might not be bound.

(I think what's happening, for the record, is that mediasoup is swallowing exceptions thrown out of its event handlers)

ebroder:
> @zarvox My preference would be to keep the notice outside the collapsible caller list, as this information is useful even if you aren't interested in who is on the call.

What about putting it in the expansion line itself?

"▶️ 1 caller (last activity: 2 minutes ago)"

grantaelliott:
> > @zarvox My preference would be to keep the notice outside the collapsible caller list, as this information is useful even if you aren't interested in who is on the call.
> 
> What about putting it in the expansion line itself?
> 
> "▶️ 1 caller (last activity: 2 minutes ago)"

I would support that (with the caveat that it wraps kind of strangely, but that shouldn't be hard to fix). Maybe right align it?

ebroder:
> > In local testing, I'm finding that the backend code isn't reliably updating the timestamp in the collection when I generate audio at a reasonable (or even at unreasonable) level:
> 
> Hmm. I did have problems with that earlier, but I thought I addressed by wrapping the update function. Does it work if you move the `Meteor.bindEnvironment` outside of the `_.debounce`? That's the only situation I can see where we might not be bound.
> 
> (I think what's happening, for the record, is that mediasoup is swallowing exceptions thrown out of its event handlers)

Ah I think I've got a repro, let me poke around some

ebroder:
Pushed a fix. `_.debounce` prevents the function from being called until `wait` time has passed without it ever being called. I wanted `_.throttle`

zarvox:
I confirm `_.throttle` fixes the server bits; thanks for digging in.

Now, I observe that the levels appear to be such that any time a participant is sending audio, the activity timestamp gets updated.  This only tells me the last time anyone was present in the call unmuted, rather than anything about "were they talking above a certain DB level" or even "was there more than one call participant at a time".  Could we try either picking a slightly higher noise floor, or requiring there be at least one stream client on the call to consider marking the call "active" based on activity on that stream?

ebroder:
Hmm. I didn't see that in my testing, but if I turn off my noise cancellation app I can repro. `router.createAudioLevelObserver` takes a `threshold` parameter (in dBvo from -127 to 0) that defaults to -80, which seems quite low. In my testing, active speech registered at about -30 or so. I set the threshold at -50 or so to account for variance in people's mic gain, but let me know if that seems to work for you (especially because Firefox seemed to set the gain quieter than Chrome does)

I'll work on the UI changes in the meantime.

ebroder:
Instead of floating the time indicator to the right, I think it looks OK if I just add a hanging indent to the heading:
Screen Shot 2022-01-12 at 12 12 40 PM

zarvox:
> @zarvox My preference would be to keep the notice outside the collapsible caller list, as this information is useful even if you aren't interested in who is on the call.

Sure, we can show the text all the time you're not in a call.

> What about putting it in the expansion line itself?
> 
> "▶️ 1 caller (last activity: 2 minutes ago)"

I'd be happy with the note in the expansion line itself, as long as when the text has to wrap the left edge stays block aligned (e.g. remove the font icon from the flow and margin-left all the text) instead of wrapping underneath the arrow.

I tried a couple things and I think I prefer not right-aligned for this text; it's especially awkward once you need to wrap text.

I would also entertain but am not attached to the idea of making shorter relative time strings like "last activity: 1y363d1h2m ago" (where each time unit is represented by a single letter, and they are joined with no whitespace) if we think that the text is too long.

zarvox:
> Hmm. I didn't see that in my testing, but if I turn off my noise cancellation app I can repro. router.createAudioLevelObserver takes a threshold parameter (in dBvo from -127 to 0) that defaults to -80, which seems quite low. In my testing, active speech registered at about -30 or so. I set the threshold at -50 or so to account for variance in people's mic gain, but let me know if that seems to work for you (especially because Firefox seemed to set the gain quieter than Chrome does)

That sounds pretty reasonable.  On the profile page we give the rough guidance:

>  *   When you aren't speaking, all bars should be below -70dBFS. If they aren't, consider moving to a quieter location if available.
>  *  When you are speaking: bars on the left should peak over -60dBFS or you'll probably be somewhat hard to hear, so you may need to speak up.
>  *  If you get louder than -35dBFS or so, you will probably come across as quite loud, and it might be polite to either speak in a softer tone or reduce your microphone volume.

I'm okay with the levels being slightly harder to trigger than what we tell people to target, but also I think most folks will *probably* land in the -50 to -30 dBFS range when speaking and maybe we should just update that guidance to match.

----

I think the hanging indent you made looks good.

ebroder:
Cool. I think action is to you on the review, then

imports/client/components/ChatPeople.tsx
```diff
@@ -154,6 +163,29 @@ const ChatPeople = (props: ChatPeopleProps) => {
   // or doing signalling).
   const rtcDisabled = useTracker(() => Flags.active('disable.webrtc'), []);
 
+  const recentVoiceActivity = useTracker(() => (
+    CallHistories.findOne({ call: puzzleId })?.lastActivity
+  ), [puzzleId]);
+  const [voiceActivityRelative, setVoiceActivityRelative] = useState<string>();
```
zarvox, line 169: This value needs to be in the deps list of the `useLayoutEffect()` below (line 437), since changes in the text here can change wrapping, which can change height, which can cause scroll position changes.

imports/server/migrations/36-call-history-indexes.ts
```diff
@@ -0,0 +1,10 @@
+import { Migrations } from 'meteor/percolate:migrations';
+import CallHistories from '../../lib/models/mediasoup/call_histories';
+
+Migrations.add({
+  version: 36,
+  name: 'Add indexes to CallHistory collection',
+  up() {
+    CallHistories.createIndex({ call: 1 }, { unique: true });
```
zarvox, line 8: Can I request a (non-unique) index on `hunt` as well?  I'm hoping to add some form of this data to the PuzzleListPage.

Totally fine if you prefer I do that in a separate migration later though

ebroder:
PTAL?

zarvox (approved):

PR #564, opened 2022-01-11
Promisify locking logic

ebroder:
In the process, split the observer monitoring for removal and the timeout monitoring for preemption into separate promises (rather than having them both resolve a single future), to avoid having them both potentially resolve the same promise.

In addition to using slightly more standardized interfaces, this should resolve some errors we saw in testing the SFU code, which exercises locks much more heavily than we did in the past, such as:

```
Exception in setTimeout callback: Error: Future resolved more than once
    at Future.return (/built_app/bundle/programs/server/node_modules/fibers/future.js:261:10)
    at monitorTimeout (imports/server/models/lock.ts:88:13)
    at imports/server/models/lock.ts:98:53
```
```
Exception in defer callback: Error: Future resolved more than once
    at Future.return (/built_app/bundle/programs/server/node_modules/fibers/future.js:261:10)
    at monitorTimeout (imports/server/models/lock.ts:96:13)
    at Object.withLock (imports/server/models/lock.ts:101:9)
    at imports/server/discord-client-refresher.ts:93:9
```

I believe that both of these exceptions were indicative of the lock getting deleted after we attempted to acquire the lock but before we fetched the lock in `monitorTimeout`, resulting in both the observer and the timeout function attempting to resolve the Future (in an undefined order, which is why there were two different stack traces)

I tested this by running `CLUSTER_WORKERS_COUNT=auto meteor` to spin up multiple workers, since those bugs in the locking code do require concurrency to exercise. I then opened 10 different tabs and joined a call, verified that some of the clients were mapped to different servers, then closed all of them at the same time, to try and maximize the number of workers simultaneously attempting to acquire the lock. I was able to reproduce the "Future resolved more than once" error on main, but not on this branch.


zarvox (approved):
I think this is safe.

PR #567, opened 2022-01-11
PuzzlePage interface style polish

grantaelliott:
- Move chat controls and call interface away from edges. The nonstandard controls are more trouble than they're worth. Adding padding makes buttons more clear, improves the focus effect, looks more natural when a bottom inset is present, and reduces amount of non-standard styling.
- Extend the fixed layout to the bottom, when bottom inset is non-zero. Resizer makes more sense stretched to full height and chat input needn't waste vertical padding. This is a first step in better using the areas outside the inset.
- Style improvements for people in chatter
  - Add gap between. Remove disused connection status bar.
  - Remove inconsistently visible background shading on deafen.
  - Fix mute and deafen icon overflow.
  - Get higher resolution discord icons (for hiDPI displays).
- Capitalize voice chat buttons (for consistency) and fix overflow on narrow chat panes
- Increase chat and metadata padding to match left-pane interface
- Make chat highlights unbroken
- Remove or lighten borders where possible

Open to suggestions.

Screenshot

### DIFF
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -275,7 +275,8 @@ $puzzle-page-padding: 8px;
 
     .btn {
       flex: 1;
-      padding: 0 auto;
+      padding-right: 2px;
+      padding-left: 2px;
       display: flex;
       align-items: center;
       justify-content: center;
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -420,8 +420,8 @@ $comm-pane-padding: 8px;
 // Puzzle metadata
 .puzzle-metadata {
   flex: none;
-  padding: 2px 4px;
-  border-bottom: 1px solid black;
+  padding: 2px 8px;
+  border-bottom: 1px solid #dadce0;
 }
 
 .puzzle-metadata-row {
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -383,8 +383,7 @@ $puzzle-page-padding: 8px;
 }
 
 .chat-message, .chat-placeholder {
-  padding: 0 $puzzle-page-padding;
-  margin-bottom: 2px;
+  padding: 0px $puzzle-page-padding 2px;
   word-wrap: break-word;
   font-size: 14px;
   &.system-message {
--- a/client/stylesheets/puzzle.scss
+++ b/client/stylesheets/puzzle.scss
@@ -319,77 +319,50 @@ $puzzle-page-padding: 8px;
     display: flex;
     align-items: center;
     justify-content: center;
-    text-transform: uppercase;
-    font-weight: bold;
-    font-size: 24px;
-    color: black;
-    line-height: 24px;
     position: relative;
     margin: 4px 4px 0 0;
 
     .icon {
       font-size: 12px;
       width: 16px;
-      text-align: center;
+      height: 16px;
+      display: flex;
+      align-items: center;
+      justify-content: center;
       color: $danger;
-      height: 50%;
-    }
-
-    $mutedTop: -4px;
-    $mutedRight: -2px;
-    $deafenDiff: 2px;
-
-    &.muted .muted-icon {
       position: absolute;
-      top: $mutedTop;
-      right: $mutedRight;
+      right: 0px;
+
+      &.muted-icon {
+        top: 0px;
+      }
+
+      &.deafened-icon {
+        bottom: 0px;
+      }
     }
 
-    &.deafened {
-      color: $secondary;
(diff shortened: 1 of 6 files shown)


client/stylesheets/puzzle.scss
```diff
@@ -257,65 +255,47 @@ table.puzzle-list {
...
   header {
-    margin: 4px 2px;
+    margin: 4px 0;
     cursor: pointer;
   }
 
   .people-list {
     display: flex;
     flex-wrap: wrap;
     justify-content: flex-end;
```
grantaelliott, line 297: I was tempted to change this, so chatting people flow left to right, but I didn't want to step on toes. @zarvox Why did you choose right to left?
```suggestion
    justify-content: flex-start;
```

zarvox: I think that was @aldeka's design originally.  Karen, do you remember your reasons for having additional person-boxes fill space right-to-left?

I can appreciate that it feels like it makes better use of the horizontal space, and that in small calls the collapse controls are easier to read when not cramped by the people-boxes, but I haven't tried the opposite yet

zarvox: Yeah, Karen confirms it's primarily for visual balance -- when there are few viewers, the left-right balance is much better, and when there are many viewers, it looks about the same either way.

zarvox:
I was initially change-resistant, but after poking at the UI for a little bit the additional padding is growing on me.  I worried about loss of visible sidebar history, but I think that winds up being at most a line, and readability is likely improved.

I'm encountering a couple issues that I'd like to find solutions for:

### When muted or deafened, buttons overflow the container

image

This appears to have been possible before when shrinking the sidebar to its minimum width, but it was only noticeable when deafened.  Now, it appears when muted as well and is much more noticeable since the margins fail to match as soon as we overflow.  Maybe we make the sidebar min-width large enough to accommodate (my measurements say 190px)?  Maybe we make the buttons text-wrap break-word, or give them smaller internal horizontal padding?

### The padding at top doesn't match between the metadata section and the call buttons

image

Dunno if you'd rather these both be 8px (like the left) or 2px (like the right) but it feels like if we are going to pad things in the chat sidebar with the same general button style, winding up in similar rows, they should probably line up.

grantaelliott:
I was back and forth on changing the input styles as well, because I liked the tightly spaced chat bar. But edge cases on the chat bar keep emerging and the voice chat buttons that imitate its style aren't obviously buttons. This style is also closer to that of most modern chat apps. I've convinced myself that this is the better option, but it's obviously a minor point and I'm definitely open to objections.

I hadn't noticed the button overflow. That has to be fixed. You're right; it's an existing problem just made a little worse. I'll also fix the microphone slash, which overflows the person box.

I agree on the metadata button misalignment. I held off on changing it for the sake of vertical space for the sheet, but you noticed it too and it's only a few pixels. The buttons are actually different heights, but I think that's really an argument for switching the chat buttons to the small variant (and get back some vertical space for chat)

Thanks @zarvox. I'll fix all those over lunch tomorrow.

zarvox:
> The buttons are actually different heights, but I think that's really an argument for switching the chat buttons to the small variant (and get back some vertical space for chat)

They both look 31px tall in the inspector on initial pageload to me; are we talking about the same thing?  I was trying to say that `div.puzzle-metadata` has 2px vertical padding (and then the `button.puzzle-metadata-action-row` has another 2px of vertical margin) but `section.chatter-section` has 8px vertical padding, which makes the buttons slightly unaligned on the page.

I guess once you shrink the sidebar if you're in the call the 3 buttons wind up wrapping text, and then they're different heights, but that seems kinda unavoidable.  I'm at peace with that winding up with the bottom line inconsistent -- if people are shrinking the sidebar below initial split size, they are probably trying to focus on the sheet anyway, and care less about seeing the chat as it stands.

grantaelliott:
I think that's everything. I updated the PR to reflect the complete set of changes.

Notably, I made a few larger than expected changes to the people in chatter:
- We were fetching discord icons at 40px, so they appeared fuzzy on hiDPI displays. I just changed the default to the nominal 128px and adding styling. (Alternatively, we could explicitly request 40 * Window.devicePixelRatio, but that's close enough to Discord's default and the file sizes are very small, so it seemed better to simply let the scaling decisions be made locally.)
- At one time, there was a background that changed color to further highlight when someone was deafened (and therefore not really in the chat). This is now covered up by avatar images, so it only displays for users without avatars. Since it was inconsistently displayed and conveys the same information as the deafened icon, I simply removed it. @aldeka, please let me know if you'd prefer that be reinstated (with some accommodation for avatars).
- Per conversation with @zarvox I removed the disused connection status bar.

zarvox (approved):

PR #568, opened 2022-01-12
Rework Google Drive organization and permissions

ebroder:
This reorganizes Google documents and spreadsheets to be stored in per-hunt folders, instead of spewed out into the root of the Drive of the account that we're using (ugh). It makes an earnest attempt to re-organize existing documents so that we have the organization scheme going forward.

Then, with documents in folders, we can rework how we grant permissions. Instead of granting each user access to each individual Google Drive document, instead grant viewer access to the per-hunt folder. Read-only access means that users can't move files around or rename them or anything (that's Jolly Roger's job). However, having explicit viewer access on the folder combined with the implicit "anyone with the link can edit" access on individual documents is sufficient to prevent them from showing up as an Anonymous Animal.

This also reduces the number of API calls we have to make to Google from one of `O(puzzles * hunters)` to `O(hunters)`, which hopefully means we're less likely to get rate-limited (especially because we can spread that `O(hunters)` out over an extended period leading up to Hunt, rather than bursting through when the first puzzles drop).

I did some local testing and things seemed to work, and I think this is valuable enough to _maybe_ be worth rushing in before Friday, but I'm also pretty apprehensive about changing something this core so late in the game. I think that as long as the basic functionality doesn't explode, this shouldn't get in the way of people solving puzzles (they just might be Anonymous Animals), but I'm also very eager to hear any suggestions of flows to test.

(Note that the migration number here conflicts with #556, so one of them will need to get fixed on rebase)


imports/server/migrations/35-reorganize-google-drive.ts
```diff
@@ -0,0 +1,22 @@
...
+import Hunts from '../../lib/models/hunts';
+import Puzzles from '../../lib/models/puzzles';
+import { ensureDocument, ensureHuntFolder } from '../gdrive';
+
+Migrations.add({
+  version: 35,
+  name: 'Migrate existing Google Drive files into new folder organization structure',
+  up() {
+    Hunts.find().forEach((hunt) => {
+      ensureHuntFolder(hunt);
```
zarvox, line 14: This approach seems fragile -- if Google integration is not configured but Hunts exist and have not been migrated, won't this throw when it reaches `checkClientOk();` and then the server won't come up, and you'll have to manually reset the migration lock.

Similarly, if the Google integration is disabled when this migration is applied, we'll never actually migrate the existing files into the different org structure.

Maybe this is better off either:

1. being reachable via an admin-only Meteor method that you call from the JS console in prod, and migrate manually, or
2. running on every startup with plenty of additional try-catch

I'm leaning towards 1 though?

imports/client/components/SetupPage.tsx
```diff
@@ -75,8 +75,9 @@ enum SubmitState {
 
 const googleCompletenessStrings = [
   'Unconfigured',
-  '1/3 complete',
-  '2/3 complete',
+  '1/4 complete',
+  '2/4 complete',
+  '3/4 complete',
```
zarvox, line 80: Do we intend to nudge people towards making a root folder, or is it more of an optional thing?  I guess we nudge them towards having a template doc, but I feel like that's more impactful to user experience than making sure that you're putting data in a subdirectory instead of your drive root.

I guess I'm wondering if gdrive.root this should be an optional setting?  If I really do want to use the drive root (e.g. because we have a whole account for running jolly-roger as), it looks like the only way to achieve that is to leave gdrive.root unconfigured, but then we'll display this config as "incomplete", which seems confusing.

imports/server/setup.ts
```diff
@@ -105,6 +107,32 @@ Meteor.methods({
...
+      throw new Meteor.Error(401, 'Must be admin to configure gdrive');
+    }
+
+    if (root) {
+      Settings.upsert(
+        { name: 'gdrive.root' },
+        { $set: { value: { id: root } } }
+      );
+
+      Meteor.defer(() => {
```
zarvox, line 125: I think that maybe this should maybe become a separate Meteor method, reachable via an optional "also migrate any of my past hunt data" that we could show in the same configuration section?  It's definitely non-obvious from the UI that filling out this setting will cause immediate side-effects on the Drive side.

imports/server/migrations/35-reorganize-google-drive.ts
zarvox: or

3. being reachable via a button on the setup page which also covers the "go do side effects until you've converged on the new directory layout under the new root folder" part of changing the setting?

imports/client/components/SetupPage.tsx
ebroder: Yeah that's reasonable. I was riffing off of the behavior of the template documents, but I can treat this as optional

imports/server/migrations/35-reorganize-google-drive.ts
ebroder: Yeah, OK, I think I'm convinced that this is a bad idea. I added a "reorganize google drive button" to the setup page, which I think should address both this and https://github.com/deathandmayhem/jolly-roger/pull/568#discussion_r783007491

Leaving this unrebased for the time being for ease of review

zarvox (approved):
Looks good

(I'm not super familiar with how the !fixup and !amend magic comments interact with GitHub but I'd be sad if the !amend clobbered the !fixup, so carefully flatten these yourself if that's a risk)

ebroder:
The magic commands aren't a GitHub thing, they're a [git thing](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash). I'll do a rebase before merging (which was already going to need some babysitting, since the commits will step on each other a bit) to make them disappear.

ebroder:
(Verified locally that `git diff d2a987d..d8196eb` only shows the `@types/styled-components` upgrade that got picked up - can't show it on GitHub because it only supports `...` diff, not `..`, but in any case the rebase didn't affect the set of changes to code)

Issue #109, opened 2018-10-14
Look into a sentry or bugsnag integration so we can get some error tracking


zarvox:
With the removal of Kadira, we have less visibility into if the server is throwing exceptions and whatnot.  Might be nice to have some more observability.

PR #572, opened 2022-01-13
Add an error boundary to the main App component

ebroder:
This gives us a slightly better UX than having the entire app unmount
and leave the user with a blank screen. The "reset" button is admittedly
not likely to solve anyone's problems, but it at least gives them some
agency and an opportunity to go somewhere else

Here's an example of what it looks like:

Screen Shot 2022-01-12 at 5 20 20 PM

I was hoping to include the stack from Mobile Firefox, but sadly I don't seem to be able to repro on my local dev environment.
.

### DIFF
--- a/imports/client/components/App.tsx
+++ b/imports/client/components/App.tsx
@@ -1,8 +1,10 @@
 import { Meteor } from 'meteor/meteor';
 import { useSubscribe, useTracker } from 'meteor/react-meteor-data';
+import { faExclamationTriangle } from '@fortawesome/free-solid-svg-icons/faExclamationTriangle';
 import { faUser } from '@fortawesome/free-solid-svg-icons/faUser';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import React, { useCallback, useMemo } from 'react';
+import Alert from 'react-bootstrap/Alert';
 import Dropdown from 'react-bootstrap/Dropdown';
 import DropdownItem from 'react-bootstrap/DropdownItem';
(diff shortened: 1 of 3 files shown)


ebroder:
Worth noting that I'm not sure how well this will work with minified JS, although it is showing something reasonable (`meteor://:computer:app`) in dev. If it doesn't work, we can probably use https://www.stacktracejs.com/ to de-minify.

ebroder:
Hmm. Alright, well, `stacktrace-js` won't be an answer out of the box, because `stacktrace-gps`, which looks up sourcemaps`, doesn't support the `X-Sourcemap` HTTP header that Meteor uses (it only respects the `//# sourceMappingURL` magic commment)

ebroder:
Oh yeah - it's also worth noting that error boundaries do not catch all errors:

> Error boundaries do not catch errors for:
> 
> * Event handlers
> * Asynchronous code (e.g. setTimeout or requestAnimationFrame callbacks)
> * Server side rendering
> * Errors thrown in the error boundary itself (rather than its children)

The good news is that none of those cause React to unmount the DOM either, so don't think there's as much urgency in catching them. `react-error-boundary` does offer a [hook](https://www.npmjs.com/package/react-error-boundary#user-content-useerrorhandlererror-unknown) that you can use to "throw" errors into the error boundary, but all else being equal I think the current behavior of having them show up in the console is incrementally preferable. (Though at some point we should follow up on #109 and include some client-side integration to go with it.)

zarvox (approved):

PR #573, opened 2022-01-13
Limit size of Discord avatars

grantaelliott:
Fixes a bug introduced in 8f81153.

Most Discord avatar images were unstyled, so fetching higher resolution images caused formatting issues.


ebroder (approved):
Approving so we can get the fix out, but I've had good luck elsewhere using `srcset` for this, with the bonus that anyone without a retina screen saves a little bandwidth

grantaelliott:
Makes sense. Looking through this, there's a lot of copy-paste code behind these that could easily be factored out and make it easy to use srcset (but in the short term, I just wanted to fix the error...).

PR #574, opened 2022-01-13
Adjust to height change after loading ChatPeople

ebroder:


grantaelliott (approved):
Can't reproduce the issue locally, but this makes sense.

PR #575, opened 2022-01-13
Run errors through StackTrace.JS before rendering them

ebroder:
This is a little frivolous, but probably worth merging. In particular, this gives us some indirection around differences in browser behavior. Chrome includes the exception message in `Error.stack`; Firefox does not. I also thought Firefox's stacktraces were kind of ugly, but that's really just because of the minification in prod. This doesn't do anything about minification yet, because `stacktrace-gps` only supports the `//# sourceMappingURL` magic comments and not the `X-SourceMap` header that Meteor uses.

Before:

87FE4A11-145C-42BA-8646-E8F84A6B405B

After:

A140F83C-74FE-4CD4-91D3-F91D656DE74C

(The rendering on Chrome is about the same before and after)


zarvox (approved):
LGTM

I'm curious if this impacts bundle size; debugging tools can often be kinda large but hopefully this is relatively simple?  (Maybe some day we want to add some automation to track code size over time...)

PR #579, opened 2022-01-13
Add a puzzle delete button

### LINKED ISSUE #425: We should have a functional "delete puzzle" button
Hide it behind like three layers of indirection for safety and make it only available to operators, but "oops two people created a puzzle entry at once" or "we already had a stub for that meta, and then made another one" should not require admin-in-js-console to rectify.

ebroder:
Operators are now allowed to delete puzzles. When they do, they can
optionally specify another puzzle that hunters should treat as its
replacement. When a puzzle is deleted, chat and guesses are disabled,
all edit permissions are revoked from the shared Google document, and a
modal pops up advising hunters that the puzzle is deleted and now
read-only.

Deleted puzzles show up in the operators' puzzle list view at the very
bottom in an independent group, and the modal on the individual puzzle
page has a "undelete" button in case of error.

Fixes #425.

Here are some screenshots

image
image
image
image

### DIFF
--- a/imports/client/components/PuzzleListPage.tsx
+++ b/imports/client/components/PuzzleListPage.tsx
@@ -90,6 +90,17 @@ const PuzzleListView = (props: PuzzleListViewProps) => {
   const allPuzzles = useTracker(() => Puzzles.find({ hunt: props.huntId }).fetch(), [props.huntId]);
   const allTags = useTracker(() => Tags.find({ hunt: props.huntId }).fetch(), [props.huntId]);
 
+  const deletedPuzzlesLoading = useSubscribe(
+    props.canUpdate ? 'mongo.puzzles.deleted' : undefined,
+    { hunt: props.huntId }
+  );
+  const deletedLoading = deletedPuzzlesLoading();
+  const deletedPuzzles = useTracker(() => (
+    !props.canUpdate || deletedLoading ?
+      undefined :
+      Puzzles.findDeleted({ hunt: props.huntId }).fetch()
+  ), [props.canUpdate, props.huntId, deletedLoading]);
+
   const [searchParams, setSearchParams] = useSearchParams();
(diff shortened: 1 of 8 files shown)


imports/client/components/PuzzlePage.tsx
```diff
@@ -948,6 +961,75 @@ const PuzzlePageMultiplayerDocument = React.memo((props: PuzzlePageMultiplayerDo
...
 
+const PuzzleDeletedModal = ({
+  puzzleId, huntId, replacedBy,
+}: { puzzleId: string, huntId: string, replacedBy?: string }) => {
+  const canUpdate = useTracker(() => userMayWritePuzzlesForHunt(Meteor.userId(), huntId), [huntId]);
+
+  const replacementLoading = useSubscribe('mongo.puzzles.allowingDeleted', { _id: replacedBy });
+  const loading = replacementLoading();
+
+  const replacement = useTracker(() => Puzzles.findOne(replacedBy), [replacedBy]);
```
zarvox, line 972: This is a `findOne`, but you subscribed with `allowingDeleted` above -- did you want `findOneAllowingDeleted` here?

imports/client/components/PuzzleListPage.tsx
```diff
@@ -286,9 +297,21 @@ const PuzzleListView = (props: PuzzleListViewProps) => {
       <div>
         {maybeMatchWarning}
         {listComponent}
+        {deletedPuzzles && deletedPuzzles.length > 0 && (
+          <RelatedPuzzleGroup
+            key="deleted"
+            group={{ puzzles: deletedPuzzles, subgroups: [] }}
+            noSharedTagLabel="Deleted puzzles (operator only)"
```
zarvox, line 304: Oh that's cute :)

zarvox (approved):

PR #580, opened 2022-01-13
Defer layout effect callback that we can't fire due to null refs

zarvox:
There's a couple places where we need height changes from other children
to trigger scrolling in another child of a component.  We had been
dispatching these events via callbacks that accessed refs, but it turns
out that under certain render batching and ordering, we would fail to
deliver the callback to the pane that needed to be scrolled.

What's more, when we miss delivering such an event, the browser
generates a scroll event on the pane to be scrolled which we must
interpret as the user scrolling the pane to that target, breaking
the previous scroll offset, and causing us to no longer keep the
chat backlog sticky at the bottom.

This patchset attempts to fix the misbehavior by adding a ref (because
we must act within *this* render batch, not later due to a setState) for
keeping track of if we need to deliver that callback once the ref is
nonnull, along with a useLayoutEffect in the parent which will check the
refs and call the callback as appropriate before yielding back to the
browser to render the changes to the DOM.

### DIFF
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -341,6 +341,7 @@ const ChatInput = React.memo((props: ChatInputProps) => {
 
   const onHeightChangeCb = useCallback((newHeight: number) => {
     if (onHeightChange) {
+      trace('ChatInput onHeightChange', { newHeight });
       onHeightChange(newHeight);
     }
   }, [onHeightChange]);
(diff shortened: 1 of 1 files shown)


ebroder (approved):
This feels a little bit sloppy to me - surely there's a more direct way to sequence side-effects in the way that we want (strawman: use a callback ref?) - but I'm supportive of making priority 1 fixing this before tomorrow. Let's try and come back and fix it up in the future?

PR #586, opened 2022-01-14
Show puzzle age & recent voice activity next to viewer count

### LINKED ISSUE #299: Include time since unlock on puzzle list page
This was requested during Hunt as a way to see which puzzles were freshest and might have low-hanging fruit vs. stale and likely abandoned. It's possible that there's some alternative indicator for puzzle progress we could lean on (more aggressive tagging? better summaries on individual puzzles?)

ebroder:
In the interest of showing more context about each puzzle on the puzzle
list page, for unsolved puzzles, next to the subscriber count, we've added:

* time since unlock, which is nice to peruse to see what puzzles are
  older and how long they've been there
* an icon indicating time since last voice detected in the audio call

These generally aren't the most important data on the page, though, so
I've made them slightly smaller font and muted in color, so as to reduce
the space they take up and the amount they draw the eye.

Fixes #299.

Competitor to #583 🙂


ebroder:
image

image

zarvox (approved):
Yeah this seems slightly better than mine -- thanks for the tweaks!

grantaelliott:
I like these tweaks as well and this, in general, is a good idea.

Unfortunately, in my quick tests, I believe it's breaking the layout on mobile. Below is a screenshot from iPhone 12 Pro simulator, showing line wrapping. I'm fine with merging this quickly in light of the time, but it'd be great to get this fixed quickly if we can.

Simulator Screen Shot - iPhone 12 Pro - 2022-01-14 at 11 25 08

Issue #618, opened 2022-01-17
Rework permissions to be scoped per-hunt


ebroder:
Right now, the operator permission is a global permission. This means that anyone who has ever been an operator on a Jolly Roger instance has those permissions in perpetuity, and it's difficult to use a single instance of Jolly Roger among different groups (who may not all use the same set of operators).

We should fix this by making roles that are scoped to an individual hunt. These would not replace global roles - at the very least, the `admin` role must remain global (it doesn't make sense to only have the admin role on a single hunt)

## Data model

Currently roles are tracked as a property of the `Meteor.User` type:

```typescript
  roles?: string[];
```

Instead, we change them to have an additional scope, tracked as a record type:

```typescript
type Scope = string;
const GLOBAL_SCOPE = '__GLOBAL__';

type Permission = string;
[...]
  roles?: Record<Scope, Permission[]>
```

In practice, the `admin` role would be the only global role, and the `operator` role would be the only hunt-scoped role, but nothing structurally prevents that. For the data migration, anyone who has the currently global `operator` role will have that role migrated onto all hunts of which they are currently members.

## "Active operator"

Today `"inactiveOperator"` is a separate role from `"operator"`. An `inactiveOperator` only has permissions to make themselves an operator (i.e. they can add themselves to the `operator` role). This was originally done when we retrofitted the ability to enable/disable operating mode, because all of the operator UI elements conditionalize on having a permission, not on having a particular role, and taking away the `operator` mode was the most expedient way to make all of them disappear.

As part of this change, make operator active status a client-side property, stored in `localStorage` and scoped per-hunt. Operators always have the operator role, and at each permission check we look at `localStorage` to decide whether to show the UI elements. To make it easier to toggle activeness, make a UI element on the puzzle list page, next to the add puzzle button.

Given that the UI element would be on an individual hunt page and not global, my preference is to make the activeness a per-hunt property.

This would address both #599 and #212.

## Future work

In the long term, it would be great if anyone with a Jolly Roger account could create a new hunt, of which they would be the sole operator. However, that is explicitly out of scope for this issue. One challenge to remember in supporting this is risk of a confused deputy with external integrations. Jolly Roger is an admin on Discord servers it inhabits by default, which is more role power than we want.

[...]

Issue #599, opened 2022-01-16
Make "is active operator" state client-driven rather than server-driven


zarvox:
@quentinmit writes:

> I'd really like operator state to be per-browser or even per-tab
> I like to put the main page up on a TV but the operator UI clutters that up, but I can't turn it off because then all my browsers lose it

Today, UI is shown based on whether you have privilege to take an action, and so to mark yourself as not operating, we drop the role that grants the privilege to do all operator-y things besides "can promote people to operator".  This should probably get reworked to just hiding the UI that the operator doesn't want to see clientside (and dropping the related subscriptions).

This might additionally make it possible for admins to not be operating, which I think is currently impossible based on the purely permission-driven approach.

I think the particular UI items that we might want to hide for operators who are not actively operating are limited to those around creating, updating, and deleting puzzles, as well as the answer queue popups for pending guesses in the notification center.

Issue #230, opened 2019-01-20
Add Puzzle button is directly below notification queue


ebroder:
...which makes it pretty hard to dispatch new puzzles when you have pending answers.

ebroder:
Wes also noted that if you accidentally set overly aggressive dingwords, the notifications can cover up the link to your profile page to turn them off.

ebroder:
@aldeka suggested having a bell in the menu bar that pops out a menu on the right, which seems like a good idea to me. We can potentially try to have a counter badge (which FontAwesome has some helpful [utility classes](https://fontawesome.com/v5.15/how-to-use/on-the-web/styling/layering) for), or just a new notifications dot.

zarvox:
Hmm.  I suspect we **do** want the drawer to cover other content, rather than displace it.  Here's why: the vast majority of flows that involve notifications (dingwords, guess queue) also involve attending to them pretty immediately, and substantially all of the other flows that you might be doing in the app (solving puzzles, scanning the puzzle list) instead benefit substantially from not giving up screen real estate to an empty, unactionable notifications drawer.  If you really want to have the operator queue visible at all times, consider leaving a tab open to the GuessQueuePage instead?

I think that making it not require an additional user action to read notifications is important.  The primary value of notifications in jolly-roger is in being immediately visible and difficult to miss.  I do not want to make an important announcement in jolly-roger and have people not read it because we never convinced them to click the bell icon in their navbar.  And when operating, I do not want to have to click in the navbar to see what the newest guess queue item that has come in is, either.  My next action, ~100% of the time, is going to be addressing it, so why make me work to see it?

To that end: I'm favorable toward the "show notifications automatically when notification count goes from zero to nonzero" and then have a click on the notification bell temporarily hide notifications again, until clicked again or until the notification count goes to zero (after which there's nothing to show, but if a new notification comes in, we should show it immediately, otherwise it's too easy to get into a state where notifications are hidden and forgotten).

I'd want the "hide the notifications for now please" button to only affect the current tab, and to not be stored anywhere durable.

PR #619, opened 2022-01-18
Per-hunt permissions

### LINKED ISSUE #212: Consistent admin UI
Many (all?) UI behaviors do not respect the "Operating" setting for accounts with admin privileges, though the control remains available on the profile page.  For example, the "Add a puzzle" button on the puzzle page is always present, even when "Operating" is unchecked.

This has been discussed in Slack and all current admins are aware of the behavior.  While ideally the behavior would be changed to reflect the operating setting, the easiest solution in the short term is to disable the "Operating" checkbox for admins with a note that this account has admin privileges and is therefore always operating.

This is the same issue originally submitted as #208.

ebroder:
Implements the design in #618.

Fixes #212, #599, #618.

Biggest UI change is to the profile list page:

Screen Shot 2022-01-18 at 12 03 46 AM
Screen Shot 2022-01-18 at 12 04 09 AM
Screen Shot 2022-01-18 at 12 04 20 AM

The page also supports searching by role (if you have access to role data, which requires you be an operator/admin):

Screen Shot 2022-01-18 at 12 08 54 AM

There's also a new button on the puzzle list page to toggle operator UI status:

Screen Shot 2022-01-17 at 11 53 03 PM
Screen Shot 2022-01-17 at 11 53 15 PM

This doesn't address #230, which can make toggling the operator UI potentially difficult, but it does peek out just enough that you can grab it.x

Screen Shot 2022-01-17 at 11 54 42 PM

### DIFF
--- a/imports/client/components/ProfileList.tsx
+++ b/imports/client/components/ProfileList.tsx
@@ -1,11 +1,13 @@
 import { Meteor } from 'meteor/meteor';
+import { useTracker } from 'meteor/react-meteor-data';
 import { faEraser } from '@fortawesome/free-solid-svg-icons/faEraser';
 import { faPlus } from '@fortawesome/free-solid-svg-icons/faPlus';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import React, {
-  useCallback, useEffect, useMemo, useRef, useState,
+  MouseEvent, useCallback, useEffect, useImperativeHandle, useMemo, useRef, useState,
 } from 'react';
 import Alert from 'react-bootstrap/Alert';
+import Badge from 'react-bootstrap/Badge';
 import Button from 'react-bootstrap/Button';
 import FormControl, { FormControlProps } from 'react-bootstrap/FormControl';
 import FormGroup from 'react-bootstrap/FormGroup';
@@ -14,17 +16,19 @@ import FormText from 'react-bootstrap/FormText';
 import InputGroup from 'react-bootstrap/InputGroup';
 import ListGroup from 'react-bootstrap/ListGroup';
 import ListGroupItem from 'react-bootstrap/ListGroupItem';
-import * as RRBS from 'react-router-bootstrap';
+import Modal from 'react-bootstrap/Modal';
+import { Link } from 'react-router-dom';
 import styled from 'styled-components';
 import { getAvatarCdnUrl } from '../../lib/discord';
+import { userIdIsAdmin } from '../../lib/is-admin';
+import { userIsOperatorForHunt } from '../../lib/permission_stubs';
 import { ProfileType } from '../../lib/schemas/profile';
 
 const ProfilesSummary = styled.div`
   text-align: right;
 `;
 
-const StyledListGroupItem = styled(ListGroupItem)`
-  padding: .25rem;
+const ListItemContainer = styled.div`
   width: 100%;
 
   display: flex;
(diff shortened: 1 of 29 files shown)


ebroder:
OK. I think this is ready for review

imports/server/migrations/37-role-reorganization.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    Meteor.users.find({ roles: 'inactiveOperator' } as any).forEach((user) => {
+      Meteor.users.update(user._id, {
+        $push: { roles: 'operator' },
+      } as any, {
+        validate: false,
+        filter: false,
+      } as any);
+      Meteor.users.update(user._id, {
+        $pull: { roles: 'inactiveOperator' },
```
zarvox, line 19: I take it you can't do both the `$push` and the `$pull` on the same target field in the same update?

imports/server/guesses.ts
```diff
@@ -74,17 +72,15 @@ class PendingGuessWatcher {
...
-  constructor(sub: Subscription) {
+  constructor(sub: Subscription, huntId: string) {
     this.sub = sub;
 
-    const user = MeteorUsers.findOne(sub.userId!)!;
-
-    this.guessCursor = Guesses.find({ state: 'pending', hunt: { $in: user.hunts } });
+    const guessCursor = Guesses.find({ state: 'pending', hunt: huntId });
     this.guesses = {};
     this.huntRefCounter = new RefCountedObserverMap(sub, Hunts);
```
zarvox, line 80: I was going to comment that it's generally better to try to share these by passing them into the constructor so that you don't double-call e.g. `added()` with the same object multiple times on the same sub, but in practice the Hunts and Puzzles for each one of these `HuntPendingGuessWatcher`s will be orthogonal, so it doesn't matter in this case.

If you were doing something where data didn't necessarily fall completely orthogonally in the same way, like reactively publishing display names for users that might overlap between hunts, then you would want to reuse the same RefCountedObserverMap for each `HuntPendingGuessWatcher`.  Which...I guess we always subscribe to all display names everywhere?  In the fullness of time maybe we'll want to make that more granular or scoped.  For now, we need the display names on each page we might show the pending guess queue, which is all pages.  We don't have to change that now, but it feels like a potential win on the fetching-data-as-needed side -- if I remember correctly, in prod I think a substantial fraction of our initial data loading is profiles/display names.

imports/server/api/resources/users.ts
```diff
@@ -37,14 +36,12 @@ const ACTIVE_THRESHOLD = 365 * 24 * 60 * 60 * 1000;
...
   const active = user.lastLogin &&
           Date.now() - user.lastLogin.getTime() < ACTIVE_THRESHOLD;
-  const operator = deprecatedUserMayMakeOperator(user._id);
 
   return {
     _id: user._id,
     primaryEmail: user.emails && user.emails[0].address,
     googleAccount: profile.googleAccount,
     active,
-    operator,
```
zarvox, line 47: Confirmed safe to remove at https://github.com/deathandmayhem/jolly-roger/pull/463#discussion_r779212334 (and we know of no other API clients)

imports/client/components/ProfileList.tsx
```diff
@@ -14,17 +16,19 @@ import FormText from 'react-bootstrap/FormText';
...
+import { userIsOperatorForHunt } from '../../lib/permission_stubs';
 import { ProfileType } from '../../lib/schemas/profile';
 
 const ProfilesSummary = styled.div`
   text-align: right;
 `;
 
-const StyledListGroupItem = styled(ListGroupItem)`
-  padding: .25rem;
+const StyledListGroupItem = styled.div`
```
zarvox, line 31: nit: if this is no longer actually a `ListGroupItem`, maybe rename this?

imports/client/components/ProfileList.tsx
```diff
@@ -44,13 +48,170 @@ const ImageBlock = styled.div`
...
+  { profile, huntId }: { profile: ProfileType, huntId: string },
+  forwardedRef: React.Ref<OperatorModalHandle>,
+) => {
+  const [visible, setVisible] = useState(true);
+  const show = useCallback(() => setVisible(true), []);
+  const hide = useCallback(() => setVisible(false), []);
+  useImperativeHandle(forwardedRef, () => ({ show }), [show]);
+
+  const demote = useCallback(() => {
+    Meteor.call('makeOperatorForHunt', profile._id, huntId);
```
zarvox, line 73: I'd prefer we waited for the successful reply from the server before dismissing the dialog, and displayed the failure message to the user on failure

imports/client/components/ProfileList.tsx
```diff
@@ -44,13 +48,170 @@ const ImageBlock = styled.div`
...
+const PromoteOperatorModal = React.forwardRef((
+  { profile, huntId }: { profile: ProfileType, huntId: string },
+  forwardedRef: React.Ref<OperatorModalHandle>,
+) => {
+  const [visible, setVisible] = useState(true);
+  const show = useCallback(() => setVisible(true), []);
+  const hide = useCallback(() => setVisible(false), []);
+  useImperativeHandle(forwardedRef, () => ({ show }), [show]);
+
+  const demote = useCallback(() => {
```
zarvox, line 72: Unusual name for a function that promotes an operator

imports/client/components/ProfileList.tsx
```diff
@@ -44,13 +48,170 @@ const ImageBlock = styled.div`
...
+          {' '}
+          an operator?
+        </p>
+      </Modal.Body>
+      <Modal.Footer>
+        <Button variant="secondary" onClick={hide}>
+          Cancel
+        </Button>
+        <Button variant="danger" onClick={demote}>
+          Demote
```
zarvox, line 96: This button label says the opposite of what the button does

imports/lib/schemas/user.ts
```diff
@@ -12,7 +12,7 @@ const UserCodec = t.type({
   createdAt: date,
   lastLogin: t.union([date, t.undefined]),
   services: t.union([t.object, t.undefined]),
-  roles: t.union([t.array(t.string), t.undefined]),
+  roles: t.union([t.object, t.undefined]),
```
zarvox, line 15: Hmmm, does io-ts give us the ability to express that this object is a `Record<string, string[]>`?

ebroder: io-ts does, but I don't think SimplSchema does. As far as I can tell, it can only represent an object with a fixed set of keys.

We don't actually use the io-ts generated codec anywhere, because there's no way to attach it to the `Meteor.users` collection that already exists. Instead, we add our modifications to `Meteor.users` with an [ambient type declaration](https://github.com/deathandmayhem/jolly-roger/blob/2f7d6c973f67855afc1f60fb24bf51f7643c9c0d/types/user.d.ts). Honestly, we should probably not use io-ts here at all, and instead just create a SimplSchema directly.

imports/server/guesses.ts
ebroder: Yeah we could be handling display names a lot better than we do now.

imports/server/migrations/37-role-reorganization.ts
ebroder: Yeah, Mongo seems to throw an error if you try to perform two operations on the same field. (I assume because the ordering would be underspecified, even though it's not in this specific case)

ebroder:
OK I think I've addressed everything you called out. PTAL?

zarvox (approved):

zarvox:
Looks like there was a GitHub Actions blip -- might need to poke checks to rerun

PR #633, opened 2022-01-21
Convert much of puzzle.scss to locally-styled components

zarvox:
This change untangles and flattens some pretty tricky contextual CSS
from the PuzzleListPage and PuzzlePage to try to make things read more
straightforwardly from the perspective of wherever a particular
requirement for a style comes from, rather than having lots of nested
selectors.

Of note: the `grid` and `table` RelatedPuzzle layouts wound up sharing
exceedingly little in practice, but they were rather coupled to certain
CSS classes, so I decided that they would be better off split out as
separate components so they could be changed and evolve more
independently. `RelatedPuzzleTable` and `PuzzleTable` are separate
components filling the role of the previous `table` layout used for the
overlay when rendering Tags in the metadata section of PuzzlePage.

I also took this opportunity to scrutinize and adjust the styling of the
answer-removal buttons next to each answer in "no operator queue" mode,
so they should be aligned better now.  Other than that, there shouldn't
be any visible changes resulting from this patchset.

There's still more to convert from puzzle.scss, but given that this is
already rather large and so much of the styling heavily interconnected,
I figured this was probably a good chunk to carve off for now.  Future
changes will hopefully be smaller changes and easier to review.


imports/client/components/Puzzle.tsx
```diff
@@ -21,14 +21,109 @@ import PuzzleAnswer from './PuzzleAnswer';
...
+import {
+  SolvedPuzzleBackgroundColor, UnsolvedPuzzleBackgroundColor,
+  ExpectsNoAnswersPuzzleBackgroundColor,
+} from './styling/constants';
+import { mediaBreakpointDown } from './styling/responsive';
+
+const PuzzleDiv = styled.div<{
+  expectsNoAnswers: boolean;
+  isUnsolved: boolean;
+  isSolved: boolean;
```
ebroder, line 33: These are mutually exclusive, right? Would we be happier if it was a single enum?

imports/client/components/PuzzlePage.tsx
```diff
@@ -545,37 +680,36 @@ const PuzzlePageMetadata = ({
...
-    <span className="puzzle-metadata-answers">
+    <PuzzleMetadataAnswers>
       {
         correctGuesses.map((guess) => (
-          <span key={`answer-${guess._id}`} className="answer tag-like">
+          <PuzzleMetadataAnswer key={`answer-${guess._id}`}>
             <span>{guess.guess}</span>
             {!hasGuessQueue && (
-              <Button className="answer-remove-button" variant="success" onClick={() => onRemoveAnswer(guess._id)}>&#10006;</Button>
+              <AnswerRemoveButton variant="success" onClick={() => onRemoveAnswer(guess._id)}>
```
ebroder, line 689: Not in scope, but this should really be a `useCallback`, right?

imports/client/components/PuzzleTable.tsx
```diff
@@ -0,0 +1,91 @@
...
+  width: 100%;
+  max-width: 100%;
+  border-collapse: separate;
+  border-spacing: 0 4px;
+`;
+
+const PuzzleTableTr = styled.tr<{
+  expectsNoAnswers: boolean;
+  isSolved: boolean;
+  isUnsolved: boolean;
```
ebroder, line 22: (Same question re: enum)

ebroder (approved):
👏🏻 I did a, like, moderately thorough read-through and didn't find anything broken, although I won't swear that (e.g.) there isn't removed styling from puzzle.scss that was still in use elsewhere. I think that's OK - we should go ahead and merge this and worse case we can clean up the bugs.

imports/client/components/PuzzlePage.tsx
zarvox: Yeah, probably the right answer is to have an actual function component for each guess, instead of defining them inline, and then yes, that component would call `useCallback`.

imports/client/components/Puzzle.tsx
zarvox: Gave it another shot and I was able to extract this in a way that I liked well enough

imports/client/components/PuzzleTable.tsx
zarvox: Used the same fix as the other here

ebroder (approved):

PR #636, opened 2022-01-22
Convert more styles from scss to styled-components

zarvox:
This covers the rest of the chat + call participants sections, leaving
pretty much just the `SplitPanePlus`-related styles defined in
`puzzle.scss`.

This change should have no visible differences.

### DIFF
--- a/imports/client/components/styling/constants.tsx
+++ b/imports/client/components/styling/constants.tsx
@@ -1,8 +1,10 @@
-import { Solvedness } from '../../../lib/solvedness';
+import type { Solvedness } from '../../../lib/solvedness';
 
 export const NavBarHeight = '50px';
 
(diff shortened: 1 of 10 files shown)


imports/client/components/styling/constants.tsx
```diff
@@ -1,9 +1,11 @@
-import { Solvedness } from '../../../lib/solvedness';
+import type { Solvedness } from '../../../lib/solvedness';
```
ebroder, line 1: No objection, but I'm mildly curious why this changed. (When Typescript added `import type`, I thought about tracking down all the places we were importing types and making them explicit, but couldn't figure out a good reason to)

ebroder (approved):

imports/client/components/styling/constants.tsx
zarvox: No particularly reason, just me being slightly sloppy.

I was observing vscode complaining about the following, which is totally not a real problem:

image

So I tried switching it to just importing the type instead.  But that didn't appear to have the effect I was hoping for either.  Restarting vscode appears to have cleared up its confusion, so I bet it just missed dispatching the file creation event in some cache somewhere.

PR #638, opened 2022-01-23
Merge profiles onto the Meteor.User object

### LINKED ISSUE #628: Merge profiles collection into users
We previously avoided storing publicly-visible data about users on Meteor's users collection. There were a few specific concerns:

* We might accidentally expose other data that should remain private (especially things like `users.services`)
* The `profile` field on users is default writeable from the client, meaning that we couldn't enforce schema

However, there are significant drawbacks to having a separate profiles collection. There are several queries we'd want to write that require joining against users (e.g. "find the profile for all users that are in hunt X"), and we have to contemplate the possibility that the user has no profile (c.f. #407). These drawbacks become more significant as (e.g.) we consider allowing anyone to create a hunt and thus want to be more respecting of privacy for non-overlapping users.

I'm not sure there's a good technical mitigation for the first concern, though certainly care and inspection where we publish user data will be important. The second is much easier to handle - the Meteor documentation specifically [recommends](https://docs.meteor.com/api/accounts.html#:~:text=To%20forbid%20users%20from%20making%20any%20modifications%20to%20their%20user%20document%3A) creating a new `deny` rule to overrule the default `allow` rule that allows writing to the `profile` field.

Given that, we should merge profiles into users. I propose specifically merging it into the `profile` field, rather than creating something new. I would specifically propose _not_ merging `hunts` and `roles` onto the `profile` field. My mental model for the new `profile` field is that we may not be willing to publish the profile to all users, but if we're willing to publish a profile to a given user, that user should receive the entirety of the field. For `hunts` and `roles`, I think we would want to filter them further (e.g. two users can see which hunts they overlap on, but not the other's hunts that they weren't involved with).

I think there are three main things to be careful with when this is implemented:

* We must make sure to add the `deny` rule to block edits to `profile`
* Our current logic publishes all profiles to any logged-in user; we'll likely need to continue to do so, but need to be careful about which fields we publish (and how that interacts with our normal strategy of letting the client specify `fields` and other query options
* When migrating the data over, we should make sure to explicitly nuke anything that has been manually set on the existing `users.profile` field, since we've historically run with the default allow rule.

ebroder:
We historically avoided using the profile field on Meteor.Users due to
its relatively lax default permissioning scheme. But those defaults are
not difficult to change, and having user information spread across two
different collections makes a lot of query patterns difficult (since
they effectively require joins).

So instead, migrate the old profiles model onto the Meteor.User object.
This is intended to be a fairly 1-for-1 migration in order to make
review easier, so it does not do things like add additional constraints
to data visibility relative to what we have historically done; that's
reserved as a future task.

Fixes #628.


imports/server/users.ts
```diff
@@ -29,6 +29,31 @@ Meteor.publish('huntMembers', function (huntId: string) {
...
+
+Meteor.publish('profiles', function () {
+  if (!this.userId) {
+    return [];
+  }
+
+  // For now, all user profiles are public, including email address
+  return MeteorUsers.find({}, { fields: { 'emails.address': 1, profile: 1 } });
+});
+
```
ebroder, line 56: Given the sensitivity around the users model, I decided that life would probably be easier if we had a more more-limited publishes rather than the singular "pass your own query" funtimes from the old `mongo.profiles` publish. (I'm not yet proposing an official style change yet, but in general I'm a little over the "pass your own query" style of publishes.). Long-term, my plan is to make these take a `huntId` parameter and then do some :sparkles:query and projection magic:sparkles:, but reserving that for future work. For now, this just preserves exactly the same functionality as we had before.

imports/client/components/AllProfileListPage.tsx
```diff
@@ -1,22 +1,22 @@
...
 const AllProfileListPage = () => {
   useBreadcrumb({ title: 'Users', path: '/users' });
-  const profilesLoading = useSubscribe('mongo.profiles');
+  const profilesLoading = useSubscribe('profiles');
   const loading = profilesLoading();
 
-  const profiles = useTracker(() => {
-    return loading ? [] : Profiles.find({}, { sort: { displayName: 1 } }).fetch();
+  const users = useTracker(() => {
+    return loading ? [] : Meteor.users.find({}, { sort: { 'profile.displayName': 1 } }).fetch();
```
zarvox, line 13: Any reason this uses `Meteor.users` directly instead of `MeteorUsers`?

imports/client/components/App.tsx
```diff
@@ -156,16 +155,8 @@ const ErrorFallback = ({ error, resetErrorBoundary }: FallbackProps) => {
 
 const AppNavbar = () => {
   const userId = useTracker(() => Meteor.userId()!, []);
-  const profileLoading = useSubscribe('mongo.profiles', { _id: userId });
```
zarvox, line 159: Oh nice this saves a roundtrip!

imports/client/components/CallSection.tsx
```diff
@@ -175,10 +174,10 @@ const ProducerBox = ({
 }) => {
   const spectraDisabled = useTracker(() => Flags.active('disable.spectra'));
   const { initial, discordAvatarUrl } = useTracker(() => {
-    const profile = Profiles.findOne(Meteor.userId()!);
+    const user = Meteor.user()!;
     return {
-      initial: profile ? profile.displayName.slice(0, 1) : 'U', // get it?  it's you
-      discordAvatarUrl: getAvatarCdnUrl(profile?.discordAccount),
+      initial: user.profile?.displayName ?? 'U', // get it?  it's you
```
zarvox, line 179:
```suggestion
      initial: user.profile?.displayName?.slice(0, 1) ?? 'U', // get it?  it's you
```

imports/client/components/ChatPeople.tsx
```diff
@@ -217,24 +216,23 @@ const ChatPeople = ({
         self = p;
       }
 
-      const user = p.createdBy;
-      const profile = Profiles.findOne(user);
-      if (!profile || !profile.displayName) {
+      const user = Meteor.users.findOne(p.createdBy);
```
zarvox, line 219: `MeteorUsers` instead of `Meteor.users`?

imports/server/profile.ts
```diff
@@ -22,23 +21,16 @@ Meteor.methods({
       muteApplause: Boolean,
       dingwords: [String],
     });
-    const user = MeteorUsers.findOne(this.userId)!;
-    const primaryEmail = user.emails && user.emails[0].address;
-
     Ansible.log('Updating profile for user', { user: this.userId });
-    Profiles.update({
+    Meteor.users.update({
```
zarvox, line 25: Got some more uses of `Meteor.users` in this file -- is this just inconsistency, or is there a set of situations where we need to use `Meteor.users` directly instead of `MeteorUsers`?

imports/server/migrations/38-consolidate-profiles.ts
```diff
@@ -0,0 +1,28 @@
...
+Migrations.add({
+  version: 38,
+  name: 'Consolidate profiles onto Meteor.users',
+  up() {
+    Meteor.users.find({ profile: { $ne: null as any } }).forEach((u) => {
+      Meteor.users.update(u._id, { $unset: { profile: 1 } }, { validate: false } as any);
+    });
+
+    Profiles.find({}).forEach((profile) => {
+      Meteor.users.update(profile._id, { $set: { profile } });
```
zarvox, line 20: If I'm not mistaken: this sets a field `_id` on the nested profile object?  Maybe pluck just the fields we actually expect here instead?

zarvox:
A couple higher-level things:

It's not clear to me if you're still intending to avoid using `Meteor.users` directly (so we're guaranteed that we pull in the deny rule and simpl-schema attachment early).  I think we likely still want to; if so, you should go back through this and do the bulk replace.

I'm somewhat skeptical the `defaultValue` on `displayName` from `imports/lib/schemas/user.ts` is actually helping much -- any time that we'd not have a name for a user, we should indicate the absence instead of just using the empty string.  It'd be good to test the situation where a user gets into a situation where they have no display name (missed their initial invitation email and then used password reset to recover their account but as a result never did the profile onboarding) and see what result that produces.  We should also make sure that `ProfileMissingMessage` in `NotificationCenter` still guides the user to populate a real profile (and perhaps even a non-empty `displayName` if the defaultValue here is automatically populating it with empty-string).

ebroder:
Alright - I went ahead and switched everything to use `MeteorUsers`, and I manually copied the fields I wanted in the migration.

For `displayName`, I dropped the default value. I think the schema now expresses that "you may not have a profile at all, but if you have a profile, it must have a `displayName` field", which seems like a fine semantic to have. I also verified that if I unset (`db.users.update({_id: '...'}, {$unset: {profile: 1}})`) the profile field or set it to an empty object (`db.uses.update({_id: '...'}, {$set: {profile: {}}})`, which should theoretically be impossible), I get the missing profile notification.

zarvox (approved):

PR #643, opened 2022-01-24
Rename files with default exports

ebroder:
For files with default exports, the recommended style is for the filename to match the name of the export. This makes (e.g.) autocompletion of default exports better. This is [cited](https://github.com/airbnb/javascript#naming--filename-matches-export) in the Airbnb style guide, which we mostly follow, although the Airbnb eslint config doesn't have a rule to enforce it. Instead I pulled in `eslint-plugin-filenames`.

Beyond the aesthetics, I think this is a useful (if noisy) refactor, because it provides better hints to autocomplete about what constant to use when importing a default export. I ran into this in #638 because VS Code kept trying to `meteor_users.ts` with the constant name `meteor_users` instead of `MeteorUsers`.

The diff is chunderous, but the manual work was mostly (a) renaming files (b) adding the eslint configuration (c) in a few cases (e.g. `imports/client/components/DeepLink.tsx`) renaming the constant in the file to satisfy eslint-plugin-filenames (d) using `eslint --fix` to correct import order.


ebroder:
Oh no. I think I'm getting screwed over by case-insensitivity on OS X

ebroder:
OK. I think I have this fixed now - the problem was that I was renaming the files in VS Code, which didn't actually execute a `git mv`, and with case-insensitivity on OS X, git didn't detect them as a move. I think I've gone through and actually committed all of the file moves now.

zarvox (approved):

PR #650, opened 2022-01-25
Move profile fields to top-level of Meteor.User

ebroder:
Meteor's support for merging different fields on the same object across
multiple subscriptions doesn't work for nested objects, meaning that a
subscription with `fields: {'profile.displayName': 1}` and a
subscription with `fields: {'profile.discordAccount': 1}` will clash,
only returning one of the two fields.

Given that this is a somewhat fundamental restriction on the current
Meteor DDP protocol, the easiest solution is to promote the fields from
the user profile to top-level themselves.

### DIFF
--- a/imports/client/components/AllProfileListPage.tsx
+++ b/imports/client/components/AllProfileListPage.tsx
@@ -12,7 +12,7 @@ const AllProfileListPage = () => {
   const users = useTracker(() => {
     return (loading ?
       [] :
-      MeteorUsers.find({ profile: { $ne: undefined } }, { sort: { 'profile.displayName': 1 } }).fetch()
+      MeteorUsers.find({ displayName: { $ne: undefined } }, { sort: { displayName: 1 } }).fetch()
     );
   }, [loading]);
 
--- a/imports/client/components/HuntProfileListPage.tsx
+++ b/imports/client/components/HuntProfileListPage.tsx
@@ -25,8 +25,8 @@ const HuntProfileListPage = () => {
     loading ?
       [] :
       MeteorUsers.find(
-        { hunts: huntId, profile: { $ne: undefined } },
-        { sort: { 'profile.displayName': 1 } },
+        { hunts: huntId, displayName: { $ne: undefined } },
+        { sort: { displayName: 1 } },
       ).fetch()
   ), [huntId, loading]);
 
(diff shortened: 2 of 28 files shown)


imports/server/users.ts
```diff
@@ -1,8 +1,25 @@
...
+const profileFields: Record<ProfileFields, 1> = {
+  displayName: 1,
+  googleAccount: 1,
+  discordAccount: 1,
+  phoneNumber: 1,
+  muteApplause: 1,
+  dingwords: 1,
+};
+
+Accounts.setDefaultPublishFields({
```
zarvox, line 18: Maybe worth a comment here that this only changes what fields are autopublished to the `Meteor.user()` object for the logged-in user.

imports/client/components/AllProfileListPage.tsx
```diff
@@ -6,13 +6,14 @@ import ProfileList from './ProfileList';
 
 const AllProfileListPage = () => {
   useBreadcrumb({ title: 'Users', path: '/users' });
-  const profilesLoading = useSubscribe('profiles');
```
zarvox, line 9: I notice that this change means that we're no longer subscribing to the roles, emails, or phoneNumber fields for a user, which means that when we pass the user list down to `ProfileList`, it will no longer be able to filter-match on the values of those fields.

imports/client/components/ProfileList.tsx
```diff
@@ -277,12 +277,12 @@ const ProfileList = ({
     const isInteresting = (user: Meteor.User) => {
       for (let i = 0; i < toMatch.length; i++) {
         const searchKey = toMatch[i];
-        if (user.profile?.displayName?.toLowerCase().indexOf(searchKey) === -1 &&
+        if (user.displayName?.toLowerCase().indexOf(searchKey) === -1 &&
           user.emails?.every((e) => (
```
zarvox, line 281: Once we switch to subscribing to only `avatars` and `displayNames`; user.emails will always be empty, as will user.phoneNumber (and maybe `user.roles` too?).  Being unable to search over those fields feels like a functional regression?

imports/client/components/HuntProfileListPage.tsx
```diff
@@ -18,15 +18,16 @@ const HuntProfileListPage = () => {
 
   const usersLoading = useSubscribe('huntMembers', huntId);
   const userRolesLoading = useSubscribe('huntUserInfo', huntId);
-  const profilesLoading = useSubscribe('profiles');
```
zarvox, line 21: Same thing here as in `AllProfileListPage` -- we're missing fields that `ProfileList` would benefit from having.

ebroder:
Yeah good callouts. PTAL?

zarvox (approved):

PR #654, opened 2022-01-27
Add a basic smoke test that all routes work

ebroder:
I'm not entirely sure how useful this will prove to be, but it feels just a little bit better than having nothing at all. This introduces a test which enumerates all routes and attempts to load each of them (using fixture data to populate the substitutions). If loading the page throws an exception, that will cause the test to fail (even though our error boundary will display the exception, it still gets rethrown as an uncaught exception which mocha picks up on).

In order to support this, I needed to get our routes into a format that I could manipulate, so I first migrated all of our routes into the `Routes.tsx` file (which is relatively straightforward using react-router's concept of [outlets](https://reactrouter.com/docs/en/v6/api#outlet) for rendering nested routes); and I switched to using react-router's [`useRoutes`](https://reactrouter.com/docs/en/v6/api#useroutes), which takes route data in a structured format. I used this as an opportunity to cleanup some of the duplicated `<AuthenticatedRoute />` tags from the JSX configuration.

### DIFF
--- a/imports/client/components/Routes.tsx
+++ b/imports/client/components/Routes.tsx
@@ -1,17 +1,24 @@
 import React, { Suspense } from 'react';
-import { Route, Routes as ReactRouterRoutes } from 'react-router-dom';
+import { Navigate, Route, Routes as ReactRouterRoutes } from 'react-router-dom';
 import { BreadcrumbsProvider } from '../hooks/breadcrumb';
 import useDocumentTitle from '../hooks/useDocumentTitle';
 import AllProfileListPage from './AllProfileListPage';
+import AnnouncementsPage from './AnnouncementsPage';
 import EnrollForm from './EnrollForm';
+import FirehosePage from './FirehosePage';
 import FirstUserForm from './FirstUserForm';
+import GuessQueuePage from './GuessQueuePage';
 import HuntApp from './HuntApp';
 import HuntListPage from './HuntListPage';
+import HuntProfileListPage from './HuntProfileListPage';
 import Loading from './Loading';
 import LoginForm from './LoginForm';
 import PasswordResetForm from './PasswordResetForm';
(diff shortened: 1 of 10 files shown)


imports/client/components/Routes.tsx
```diff
@@ -1,46 +1,89 @@
...
+  { path: '/hunts', element: <HuntListPage /> },
+  { path: '/users/:userId', element: <ProfilePage /> },
+  { path: '/users', element: <AllProfileListPage /> },
+  { path: '/setup', element: <SetupPage /> },
+  { path: '/rtcdebug', element: <RTCDebugPage /> },
+].map((r) => {
+  return {
+    ...r,
+    element: <AuthenticatedPage>{r.element}</AuthenticatedPage>,
+  };
```
ebroder, line 52: Note that it's intentional that this will only act on top-level routes; it's not a tree walk. That matches the current behavior, since once we have authentication on the top-level route, we don't need it on the sub-routes.

tests/acceptance/smoke.tsx
```diff
@@ -0,0 +1,142 @@
...
+    );
+  };
+
+  const fixtureHunt = Object.keys(fixtures)[0];
+  const fixturePuzzle = fixtures[fixtureHunt].puzzles[0]._id;
+
+  describe('routes', function () {
+    before(async function () {
+      await Meteor.callPromise('test.resetDatabase');
+      await Meteor.callPromise('test.authentication.createUser');
```
zarvox, line 61: Given that we have a method for provisioning the first (admin) user from an empty database -- would it make sense to eliminate this method and replace it with a call to `provisionFirstUser`?  Then you don't have to mess with `test.authentication.addRole` either, and you're also getting some coverage benefit from exercising the actual first-user-setup flow.

zarvox (approved):

PR #656, opened 2022-01-27
Use existing methods in tests

ebroder:
Because we're resetting the database, we can take advantage of
`provisionFirstUser` to make ourselves an admin account, and then use
those admin privileges to make ourselves an operator.

(Following up on a suggestion from @zarvox in #654)


zarvox (approved):

zarvox:
Looks like the linter wants you to remove some unused imports in tests/acceptance/lib.ts (Accounts, check, addUserToRole) and tests/acceptance/smoke.tsx (GLOBAL_SCOPE)

PR #657, opened 2022-01-28
Use eslint-formatter-gha in testing

ebroder:
eslint-formatter-gha detects if its running inside of a GitHub Action, and if so uses GitHub's [workflow commands](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions) to format the output in a way that results in annotating the diff. I introduced a trivial lint failure to verify that it threads through correctly, which I'll remove before merging, obviously.

Sadly, I wasn't able to find a similar plugin for `tsc`. It's rather difficult to search for anything involving "typescript output formatter" because inevitably you either find information about `tsc` outputting transpiled Javascript or people wanting to run `prettier` or similar on their TypeScript code. However, my conclusion on a quick skim of TypeScript's plugin architecture was that it doesn't actually allow you to modify the output format, so I think we may end up needing to, like, `sed` our way to success if we want to also incorporate this for type errors.

In any case, it seems likely to be enough work to defer it from this PR and reserve as a future improvement.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,3 +51,5 @@ jobs:
         cache-to: type=gha,mode=max # mode=max means cache intermediate images
         build-args: |
           GIT_REVISION=${{ github.sha }}
+          CI=true
+          GITHUB_ACTIONS=true
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,9 @@ EOF
 
 WORKDIR /app
 
+ARG CI=true
+ARG GITHUB_ACTIONS=
+
 # Install Meteor
 COPY .meteor/release /app/.meteor/release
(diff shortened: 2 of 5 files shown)


ebroder:
Ugh OK this may be doomed with the current build pipeline, because `docker buildx` prefixes all of the output with the stage number + time (in seconds) since the build started:

```
#25 75.44 ::group::ESLint Annotations
#25 75.44 ::error file=/app/imports/client/main.tsx,line=42,col=1::Too many blank lines at the end of file. Max of 0 allowed.
#25 75.44 ::endgroup::
```

and that keeps GitHub from recognizing the workflow commands. I think if we wanted to make this work, we might need to, like, output an intermediate docker image and `docker run` the tests instead of having them happen as part of the container build.

Issue #605, opened 2022-01-16
Fix width of puzzle activity indicators individually


grantaelliott:
Currently the puzzle activity indicators are fixed size as a group, but this causes them to slightly misalign with each other. (Unrelatedly, the current size is a little too small.)

I suggest each (of currently three) indicator be given its own fixed width. This comes at the cost of a few horizontal pixels, but improves legibility and reinstates vertical alignment for each indicator.

PR #658, opened 2022-01-28
Tweak puzzle activity monitor

aldeka:
Increased the spacing between activity monitor items a little bit, and adjusted the column width so it wouldn't overflow.

Moved the call activity icon to the third position, and hid it if there's never been a call.

In the audio icon's previous (second) position, it was a choice of either having the margin between the items look unbalanced (as the differing volume icons change width quite a bit) or having activity elements move as values change. Neither of those seemed like a good idea. By moving it third, we only have to worry about the icon's left margin relative to the other icons and can reserve a fixed width for the icon to appear in if it is absent. Plus it makes it easier to differentiate puzzles with calls from puzzles without at a glance!

### DIFF
--- a/imports/client/components/Puzzle.tsx
+++ b/imports/client/components/Puzzle.tsx
@@ -71,8 +71,12 @@ const PuzzleTitleColumn = styled(PuzzleColumn)`
 `;
 
 const PuzzleActivityColumn = styled(PuzzleColumn)`
-  width: 120px;
+  width: 11rem;
   text-align: right;
+  // Push to take whole row in narrow views
+  ${mediaBreakpointDown('xs')`
+    flex: 0 0 100%;
+  `}
 `;
 
 const PuzzleLinkColumn = styled(PuzzleColumn)`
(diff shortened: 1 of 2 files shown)


aldeka:
Screen Shot 2022-01-29 at 15 18 20
Screen Shot 2022-01-29 at 15 18 40
Screen Shot 2022-01-29 at 15 19 17

ebroder:
I like this overall, but it does take up a fair bit more space (looks like about half the screen?) on mobile. Does it make sense to (say) have a media breakpoint that compacts things a bit more at small screen sizes, potentially at the cost of inter-row alignment?

grantaelliott (approved):
I agree with @ebroder - this looks really nice but takes up a lot of space on mobile. I suggest we give PuzzleActivityColumn its own row on XS viewports, as we already do for TagListColumn and PuzzleAnswerColumn. (My preference would be to keep PuzzleLinkColumn in the first row, but I suppose it could move to the new row as well.)

Even with that caveat, I think this is an improvement, so if you'd prefer to open an issue and deal with that later, I'd be ok with merging this as is.

It's a little strange that there's a popover for nothing, but I do see the advantage of dropping the icon instead of using the empty speaker. I lean towards removing the popover as well in that case (but leaving the space-filler).

This closes #605. Thanks!

PR #674, opened 2022-02-01
Use eslint-formatter-gha to annotate diff with lint errors

ebroder:


Dockerfile
```diff
@@ -45,12 +48,15 @@ COPY . /app
...
-  export METEOR_ALLOW_SUPERUSER=1
-  meteor npm run lint
-  meteor npm run test
+COPY <<-EOF /test.sh
+	#!/bin/bash
+	set -eux
+	set -o pipefail
+	export METEOR_ALLOW_SUPERUSER=1
+	meteor npm run lint | sed -e "s,/app/,\${PATH_PREFIX:+\${PATH_PREFIX}/},g"
+	meteor npm run test
```
ebroder, line 57: FTR, the switch from spaces to tabs here was due to relying on the space-stripping behavior of `<<-EOF` heredocs, which only strip tabs (not spaces). I figured if I was going to use tabs here, I should use them everywhere.

zarvox (approved):

PR #676, opened 2022-02-03
Take advantage of autopublish for some cleanup

ebroder:
We're already auto-publishing fields on user, so we might as well skip
the explicit publishes we were using before. For feature flags, Meteor
will auto-publish to the client any subscription with a null name, so we
don't need an unconditional subscribe (because Meteor will do it for
us).


ebroder:
Similarly I think this is non-controversial so going to force-merge.

PR #677, opened 2022-02-06
Only publish profile information for users with a common hunt

### LINKED ISSUE #629: Limit profile publishing based on hunt overlap
Currently, profiles are published to all logged in users. If we allow people to start creating their own hunts, we need to contemplate the possibility that someone wouldn't want their data (which includes sensitive stuff like phone number) shared with other users they have no connection to.

This should be moderately straightforward once #628 is complete, since we can add a `{ hunts: { $in: user.hunts } }` term to the query against users (because of how Mongo handles queries against arrays, the `$in` will match against both the array in its entirety and - more importantly for our purposes - each of the individual array entries).

ebroder:
This moves away from publishing all profiles to all users, and instead limits profiles to the set of users that have a common hunt with you. There are two main chunks of work to support this:

* Historically, we relied on subscribing to display names for all users, which we used to render pending guesses, dingword chat notifications, and announcements. All three of these are rendered globally (as part of the `NotificationCenter`), and not scoped to a particular hunt. Two of the three already had a special publish that joined together multiple collections, so I opted to add a special publish for `pendingAnnouncements` and include the relevant display name as part of those publishes. To make it easier to manage these sorts of publishes, I added a new `JoinPublisher` which takes a tree of foreign key relationships - basically a generalized version of what we were previously doing with `RefCountedObserverMap` that also supports arbitrary depth joins. All other cases where we subscribe to display names or avatars are easily scoped down to an individual hunt.
* Historically, if you can see a user's profile you can see all of it, including their membership in hunts that you aren't a part of (and for operators, you could see all roles, not just the hunt for which you are an operator). Instead, we now slice the list of hunts to only include the ones you're a member of. In order to ensure that this renders correctly when hunt membership changes, all user publishes are now reactive to the subscriber's hunt membership.

From a performance perspective, we should still be able to mostly take advantage of Meteor's coalescing of identical queries; the only problematic queries are the `"allProfiles"` and `"profiles"` publishes, which use an `$in` query. However, since I don't generally expect lots of people to be simultaneously viewing those pages to begin with, it seems relatively harmless.

This should mostly be reviewable commit-by-commit.

Fixes #629.


imports/server/JoinPublisher.ts
```diff
@@ -0,0 +1,266 @@
...
+          this.observers.get(join.model._name)!.decref(val);
+        });
+
+        // finally update this.values
+        foreignKeys?.forEach(({ field }) => {
+          const val = fields[field] as unknown as string;
+          if (!val) {
+            return;
+          }
+          fkValues.set(field, val);
```
zarvox, line 160: This is inconsistent with the comment, and from what I can tell `fkValues` is not read again after writing it here.  Is this supposed to be updating something that persists across the `added()` `changed()` `removed()` calls?  maybe `this.values`?

zarvox: Oh, never mind -- `fkValues` is a reference to the value stored in `this.values` for `id`, so updating that Map is exactly what we want here.  I've been in static single assignment land for too long

zarvox (approved):

PR #679, opened 2022-02-06
Store all puzzle list state in localStorage

### LINKED ISSUE #595: Save group collapse state durably
By late hunt, there's often a bunch of tag groups that you're not that interested in seeing all of, and you might want for them to stay collapsed until further notice, as you open and close dozens of tabs.

I think there's some subtle behavior here vis-a-vis interactions with filter-search, where while you're searching I think we want to default to expanding everything (lest your collapse hide something of interest), but otherwise we might track which tags are collapsed and save that in localStorage somewhere?

ebroder:
This includes the collapse/uncollapsed state of individual groups as
well as the group/unlock preference. In order to prevent proliferation
of localStorage keys, everything is switched to use-persisted-state and
grouped under a single key as a data structure.

To avoid confusion when searching, we ignore the collapse state of all
groups when showing any search results. And in order to avoid needing to
do a bunch of clicks, there's a new "expand all" button that will expand
any collapsed groups.

Fixes #595.

### DIFF
--- a/imports/client/components/PuzzleListPage.tsx
+++ b/imports/client/components/PuzzleListPage.tsx
@@ -10,7 +10,7 @@ import { faUserCog } from '@fortawesome/free-solid-svg-icons/faUserCog';
 import { faUsers } from '@fortawesome/free-solid-svg-icons/faUsers';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import React, {
-  useCallback, useEffect, useRef, useState,
+  useCallback, useEffect, useRef,
 } from 'react';
 import Alert from 'react-bootstrap/Alert';
 import Button from 'react-bootstrap/Button';
@@ -31,7 +31,12 @@ import Puzzles from '../../lib/models/Puzzles';
 import Tags from '../../lib/models/Tags';
 import { userMayWritePuzzlesForHunt } from '../../lib/permission_stubs';
 import { PuzzleType } from '../../lib/schemas/Puzzle';
-import { useOperatorActionsHiddenForHunt } from '../hooks/persisted-state';
+import {
+  useHuntPuzzleListCollapseGroups,
+  useHuntPuzzleListDisplayMode,
+  useHuntPuzzleListShowSolved,
+  useOperatorActionsHiddenForHunt,
+} from '../hooks/persisted-state';
 import PuzzleList from './PuzzleList';
 import PuzzleModalForm, {
   PuzzleModalFormHandle, PuzzleModalFormSubmitPayload,
@@ -81,10 +86,6 @@ const FilterToolbarInputGroup = styled(InputGroup)`
   }
 `;
 
-function showSolvedStorageKey(huntId: string): string {
-  return `showsolved-${huntId}`;
-}
-
 const PuzzleListView = ({
   huntId, canAdd, canUpdate,
 }: {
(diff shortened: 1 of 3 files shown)


imports/client/components/PuzzleListPage.tsx
```diff
@@ -368,10 +358,11 @@ const PuzzleListView = ({
           <ViewControlsSection>
             <FormLabel>View puzzles by:</FormLabel>
             <ButtonToolbar className="puzzle-view-buttons">
-              <ToggleButtonGroup type="radio" className="mr-2" name="puzzle-view" defaultValue="group" value={displayMode} onChange={switchView}>
+              <ToggleButtonGroup type="radio" className="mr-2" name="puzzle-view" defaultValue="group" value={displayMode} onChange={setDisplayMode}>
                 <ToggleButton variant="outline-info" value="group">Group</ToggleButton>
                 <ToggleButton variant="outline-info" value="unlock">Unlock</ToggleButton>
               </ToggleButtonGroup>
+              <Button variant="info" className="mr-2" onClick={expandAllGroups}>Expand all</Button>
```
zarvox, line 365: It's sorta weird that we show this when the display mode is "unlock" since it literally does not impact that view at all?

Maybe we could show it only if the current displayMode is "group" (and maybe put it after `showSolved`)?

ebroder: Hmm. It feels a little jumpy to me to have it show/vanish:

https://user-images.githubusercontent.com/28167/155198194-3955e8a7-8725-429b-bb72-d93d6ea999d2.mov

Thoughts on disabling it instead?

https://user-images.githubusercontent.com/28167/155198370-97a8cb88-b2d0-40f6-94c6-b9108ab6fc8e.mov

(No objections to moving it to a different spot; I don't have a strong feeling)

imports/client/components/RelatedPuzzleGroup.tsx
```diff
@@ -65,16 +69,18 @@ const RelatedPuzzleGroup = ({
   }
   return (
     <PuzzleGroupDiv>
-      <PuzzleGroupHeader onClick={toggleCollapse}>
-        <FontAwesomeIcon fixedWidth icon={collapsed ? faCaretRight : faCaretDown} />
+      <PuzzleGroupHeader onClick={forceExpand ? undefined : toggleCollapse}>
+        {!forceExpand && (
```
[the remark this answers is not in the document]
ebroder: Yeah, I think that's reasonable. I think I implemented roughly what you described, although structured slightly differently. The behavior should be:

- if there's no search, collapse state is tracked in localStorage
- if there is a search, collapse state is tracked in a non-localStorage `useState`
- when we transition between search/no search, reset non-localStorage collapsed to false (it's not strictly necessary to reset on both transitions, but one of them is a noop and it was slightly simpler to structure)

imports/client/components/PuzzleListPage.tsx
zarvox: I think we're getting a lot of saturated-color buttons in that row which are all trying to draw the user's attention, and I wish we could make them feel a bit lighter or make them take up less space, since they're really not the most important content on the page, but alas.

I tried putting the "expand all" button on the far right (after the operator button set) and there were things I liked about it (weaker blob of a bunch of saturated-color buttons all together, and the churn seemed pretty minimal when removing the expand button when toggling view mode) but there were also things I disliked about it (it makes the wrapping on mobile super goofy, and then visually it feels like the Expand all button is logically a part of the filter search bar, which also doesn't feel *quite* right ontologically).

I'm not really sure what to do with positioning here.  Nothing we've tried yet feels *right*, really.  Part of me wonders if it's something we should make "expand all" (and maybe collapse all?) into a button that appears on hover (or on touchscreen devices: active focus/click) at the far right of the collapsible rows instead -- it'd keep the feature closer to what people are already going to reach for to interact with the thing by default anyway?  Hmmm, that might be a little weird since maybe it would feel contextually like the button only affects the children of the group that you're clicking on.

I dunno.  I can live with the current button location and keeping it there disabled when the display mode means that the button won't do anything, but I will ask that we also disable the button if all groups are already expanded (making clicking the button a no-op).

Maybe we can take a UI-rework pass over this section of the puzzle list page at some later point and try to file down some of the rough edges here.

imports/client/components/RelatedPuzzleGroup.tsx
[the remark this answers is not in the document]
zarvox: 👍🏻 sounds reasonable

imports/client/components/PuzzleListPage.tsx
ebroder: OK, made a few more changes: I pulled the "expand all" button down into its own row that disappears when you're in unlock mode. I find it less jarring because that section of the UI is already going to churn when you switch modes. I also disabled it when all groups are expanded, and switched it to secondary colors which are a little quieter:

https://user-images.githubusercontent.com/28167/155387495-b185a645-72c2-41aa-83c7-dbb6a0d7c5ab.mov

(I did consider a "collapse all" button for parity, but I think there are both practical and philosophical reasons to not. Practically, it's kind of annoying to tell if any groups are uncollapsed because it's kind of annoying to enumerate all of the groups. Philosophically, I'm not sure it makes sense to look at JR with all groups collapsed. Like, what's the point of having the puzzle list open at all at that point?)

zarvox: I like it!  Thanks for the iteration! :)

zarvox (approved):

PR #690, opened 2022-02-08
Dependency updates

ebroder:
This drops a few dependencies that weren't in use:

* `lru-cache` - Dropped after switching from `adornis:typescript` to Meteor-native TypeScript support in #306
* `md5` and `@types/md5` - Dropped after switching away from Gravatar in #555
* `url` - Originally added for `googleapis`, but it's also exposed via `meteor-node-stubs` and the Google Drive integration seems to still work without it
* `@types/fibers` - This is no longer necessary after reworking locks to use promises in #564
* `assert` - This is needed for `bundle-visualizer` but is now exposed via meteor-node-stubs.

It also updates the fontawesome dependencies. There are no changes required for those updates to work, but they depend on each other so have to be taken together (which is why dependabot's PRs didn't work)


ebroder:
Straightforward cleanups, so going to force-merge

PR #691, opened 2022-02-09
Factor out avatar

grantaelliott:
Factor out avatar for maintainability and add a few improvements:
- Fall back from Discord to initial avatars everywhere in the app
- Add support for hiDPI displays to Discord avatars
- Constrain getAvatarCdnUrl to supported resolutions
- Use consistent fallback initial for users with no display name
- Add globally consistent colored background to initial avatars to improve differentiation

It would be great (and straightforward) to add Google avatar as a fallback between Discord and initial, but I think that would require storing slightly more when linking accounts.

Example of both supported avatars in chat:

All color variants for the updated initial avatar (Based on https://sashamaps.net/docs/resources/20-colors/ with black substituted for beige, with is hard to see against a white background):


imports/lib/discord.ts
```diff
@@ -3,9 +3,13 @@ import { DiscordAccountType } from './schemas/DiscordAccount';
...
 const API_BASE = 'https://discord.com/api/v8';
 
+// If size is not a supported resolution, returns the next supported size up (or the max size)
 function getAvatarCdnUrl(da?: DiscordAccountType, size: number = 128): string | undefined {
   if (da?.avatar) {
-    return `https://cdn.discordapp.com/avatars/${da.id}/${da.avatar}.png?size=${size}`;
+    const supportedSizes = [16, 20, 32, 40, 60, 64, 80, 100, 128, 160, 256, 320, 512];
+    const maxSize = supportedSizes[supportedSizes.length - 1];
+    const requestSize = supportedSizes.find((s) => s >= size) ?? maxSize;
+    return `https://cdn.discordapp.com/avatars/${da.id}/${da.avatar}.png?size=${requestSize}`;
```
ebroder, line 12: Are there cases where we actually want to render an arbitrary-sized avatar, or does it make sense to require passing an exact size instead? Something like

```ts
export type VALID_AVATAR_SIZE = 16 | 20 | 32 | 40 | ...;

function getAvatarCdnUrl(da?: DiscordAccountType, size: VALID_AVATAR_SIZE = 128: string | undefined {
...
}
```

would make it a type error to pass in an unsupported size (and you could use `VALID_AVATAR_SIZE` elsewhere that a size was being thrown around)

imports/client/components/Avatar.tsx
```diff
@@ -0,0 +1,68 @@
...
+      ['#fabed4', '#000000'], ['#469990', '#ffffff'], ['#dcbeff', '#000000'],
+      ['#9A6324', '#ffffff'], ['#000000', '#ffffff'], ['#800000', '#ffffff'],
+      ['#aaffc3', '#000000'], ['#808000', '#ffffff'], ['#ffd8b1', '#000000'],
+      ['#000075', '#ffffff'], ['#a9a9a9', '#ffffff'],
+    ];
+    const initial = displayName ? displayName.trim().slice(0, 1).toUpperCase() : '?';
+    const idSum = Array.from(_id ?? '').reduce((t, c) => t + c.codePointAt(0)!, 0);
+    const [circleColor, initialColor] = palette[idSum % palette.length];
+    const style = { backgroundColor: circleColor, color: initialColor };
+    return <AvatarInitial style={style}>{initial}</AvatarInitial>;
```
ebroder, line 60: Can we do these as props on `AvatarInitial` instead of injecting CSS?

imports/client/components/Avatar.tsx
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  const content = makeDiscordAvatar() || makeInitialAvatar();
+  const inlineStyle = inline ? { display: 'inline-block' } : { };
+  return <AvatarContainer size={size} style={inlineStyle}>{content}</AvatarContainer>;
```
ebroder, line 65: (Same here)

imports/client/components/Avatar.tsx
```diff
@@ -0,0 +1,68 @@
...
+      ['#9A6324', '#ffffff'], ['#000000', '#ffffff'], ['#800000', '#ffffff'],
+      ['#aaffc3', '#000000'], ['#808000', '#ffffff'], ['#ffd8b1', '#000000'],
+      ['#000075', '#ffffff'], ['#a9a9a9', '#ffffff'],
+    ];
+    const initial = displayName ? displayName.trim().slice(0, 1).toUpperCase() : '?';
+    const idSum = Array.from(_id ?? '').reduce((t, c) => t + c.codePointAt(0)!, 0);
+    const [circleColor, initialColor] = palette[idSum % palette.length];
+    const style = { backgroundColor: circleColor, color: initialColor };
+    return <AvatarInitial style={style}>{initial}</AvatarInitial>;
+  };
```
ebroder, line 61: It's a little weird to have functions that are declared inside of a component. I don't think it's a performance problem here, but it can be a code smell for creating new objects on every render in a way that can break React's diff detection. Can we pull these into separate components instead?

grantaelliott: Sure. I think the fallback probably has to be more explicit than null coalescence if they're separate components, but that's not really a problem.

imports/client/components/Avatar.tsx
grantaelliott: I know that's a frequent preference, but I'd argue that for a style that only affects appearance, preserving modularity of the generated classes has real value. If the colors are props of the styled div, separate classes are generated for each color combination and testing style changes in-browser is impractical without a way to affect all instances at once. (The inline style below actually represents a different usage and I have no objection to just making that a prop.)

That said, it's not a huge matter either way and if we prefer avoiding css injection as a matter of convention, I'm fine with it.

imports/lib/discord.ts
grantaelliott: There actually is a big case for requesting an arbitrary size: hiDPI assets. For instance, in the chat, we request a 40px avatar, which makes calls for 40px, 80px, and 120px, the last of which isn't an available size and gets bumped to 128px.

I'm happy to add `VALID_AVATAR_SIZE` as you suggest. but I'd argue we still need to add a companion utility method to get the next available size. (Since `Avatar` is now the only thing that calls `getAvatarCdnUrl`, that utility method would always be invoked and it wouldn't change much, but I see the argument that this improves the discord library.) Assuming we went that way, would you prefer the utility method be a stand-alone size calculator or wrap `getAvatarCdnUrl`?

imports/client/components/Avatar.tsx
ebroder: Hmm. Yeah, OK I'm fine with that

imports/lib/discord.ts
ebroder: Ah OK I hadn't considered hiDPI. I'm OK leaving this as-is given that.

ebroder (approved):

PR #693, opened 2022-02-09
Make root redirects replacements

ebroder:
This avoids breaking the back button.


ebroder:
I think this is pretty non-controversial so going to go ahead and force-merge this.

PR #694, opened 2022-02-09
Upgrade to Meteor 2.6

ebroder:
Meteor 2.6 has been released, but hasn't yet been marked as the "recommended release", so has to be manually selected with `meteor update --release 2.6`.

The primary change here is support for MongoDB 5.0, including upgrading to a new major version of the node mongo driver, which is somewhat pressing as MongoDB Atlas will be imminently upgrading to 5.0 (within the next week or two). So long as we're using Meteor's wrappers around mongo, this should be a transparent upgrade; the main consideration is that underlying node library has changed pretty significantly, which potentially affects anything using `rawCollection`.

Fortunately the one place that we use `rawCollection` (our idempotent re-implementation of `dropIndex`) still works with the new API. We do need to worry about any meteor packages that we've included that are also using `rawCollection` (or `rawDatabase`). Fortunately, I think that's just `xolvio:cleaner`, and while it is using a deprecated method ([`Collection.remove`](https://mongodb.github.io/node-mongodb-native/4.3/classes/Collection.html#remove)), it is only deprecated not broken.

The MongoDB driver did switch the name of the projection option from `fields` to `projection`. I opted not to migrate in this PR as (a) Meteor is maintaining a compatibility layer (b) the type declarations haven't been updated yet which would make it annoying, but we should eventually come back and do that.


ebroder:
(The full migration guide is available at https://guide.meteor.com/2.6-migration.html)

ebroder:
@zarvox There's definitely no rush to you to review this, but just as an FYI, I will likely force-merge if Atlas upgrades us to 5.0 before you have a chance to look at it.

ebroder:
Our Mongo Atlas cluster was upgraded to 5.0, and I'm seeing some weird behavior with multiple subscriptions that deliver the same record, so I'm going to go ahead and force-merge this since I suspect there's some behavior change here.

PR #725, opened 2022-02-26
Use eslint-formatter-gha in testing

ebroder:
This is another take on #674. The first attempt broke because it didn't copy the scripts in correctly, so the app in the prod container wasn't able to startup. This should fix that.


.github/workflows/build.yml
```diff
@@ -39,8 +39,23 @@ jobs:
...
+        load: true
+        cache-from: type=gha
+        cache-to: type=gha,mode=max # mode=max means cache intermediate images
+        build-args: |
+          GIT_REVISION=${{ github.sha }}
+          CI=true
+          GITHUB_ACTIONS=true
+    - name: Run tests
+      run: |
+        docker run -e PATH_PREFIX=${{ github.workspace }} -e CI=true -e GITHUB_ACTIONS=true --rm ${{ env.TEST_TAG }}
```
zarvox, line 57: `env.TEST_TAG` doesn't appear to be set, which is why the actions run failed when it got to here and docker run had no actual container image argument passed to it

ebroder: Yeah looks like the cherry-pick didn't work the way I thought it would. I'll clean this up so that it actually has all of the changes from #674.

ebroder:
Here's a comparison of #674 vs. this commit:

```diff
evan@mathias jolly-roger (evan/eslint-formatting-take-2) % interdiff <(git diff ':/Merge pull request #674^-') <(git show)
diff -u b/Dockerfile b/Dockerfile
--- b/Dockerfile
+++ b/Dockerfile
@@ -99,11 +99,12 @@
 EOF

 COPY --from=build /built_app /built_app
+COPY scripts /built_app/scripts

 ARG GIT_REVISION
 RUN <<EOF
 	set -eux
-	if [ -n "$GIT_REVISION" ]; then
+	if [ -n "${GIT_REVISION:-}" ]; then
 		echo $GIT_REVISION > /built_app/GIT_REVISION
 	fi
 EOF
diff -u b/package-lock.json b/package-lock.json
--- b/package-lock.json
+++ b/package-lock.json
@@ -7083,12 +7083,6 @@
         "tslib": "^1.8.1"
       }
     },
-    "tunnel": {
-      "version": "0.0.6",
-      "resolved": "https://registry.npmjs.org/tunnel/-/tunnel-0.0.6.tgz",
-      "integrity": "sha512-1h/Lnq9yajKY2PEbBadPXj3VxsDDu844OnaAo52UVmIzIvwwtBPIuNvkjuzBlTWpfJyUbG3ez0KSBibQkj4ojg==",
-      "dev": true
-    },
     "tweetnacl": {
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/tweetnacl/-/tweetnacl-1.0.3.tgz",
@@ -7191,6 +7185,12 @@
         "tslib": "^1.8.1"
       }
     },
+    "tunnel": {
+      "version": "0.0.6",
+      "resolved": "https://registry.npmjs.org/tunnel/-/tunnel-0.0.6.tgz",
+      "integrity": "sha512-1h/Lnq9yajKY2PEbBadPXj3VxsDDu844OnaAo52UVmIzIvwwtBPIuNvkjuzBlTWpfJyUbG3ez0KSBibQkj4ojg==",
+      "dev": true
+    },
     "tweetnacl": {
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/tweetnacl/-/tweetnacl-1.0.3.tgz",
```

Issue #630, opened 2022-01-20
Migrate external integration configuration from global to per-hunt


ebroder:
Right now, external integration configurations are global. If we allow anyone to create a hunt (or even if we expanded hunt edit permissions from admins to operators), this creates a confused deputy risk. Jolly Roger is (e.g.) an admin on the servers it inhabits, which is quite a lot of role-assignment power.

The best way to handle this is to make external integrations a property of the individual hunt. This is most pressing for Discord, and secondarily pressing for mailing list integration (though #495 may have some bearing here).

I'll copy what @zarvox said when I originally proposed this as part of #618:

> We could also take this opportunity and overall goal as an opportunity to rethink the nature of how some of our external integrations are configured. For instance, instead of having server-global settings for doc templates, Drive folder and server account, we could have an approach where when configuring a hunt, the hunt-creating operator configures credentials for the backend to use for that hunt. Then, instead of having a single gdrive client refresher, we'd have one-per-hunt, and all serverside actions on drive would need to look up the hunt and use the credentials from that Hunt. This would definitely increase the surface area of credentials that we'd need to store and retrieve, and perhaps we'd want to store these values separately from the actual Hunt objects themselves to ensure we don't accidentally publish them to clients, but I see no fundamental reason why we couldn't take this alternate approach, and in our case we could continue to use our current google account in production as the "data owner" that provides the Drive folder and the ability to create documents within that folder to the hunt config. Splitting out the server-wide "identity/code owner" from the per-hunt "data/access owner" feels appropriate.
>
> Discord guild configuration feels somewhat similar. I could imagine an additional step in the Hunt-provisioning flow where the user creating the Hunt would create an OAuth delegation to the bot (probably for a smaller set of permissions than admin, as you've noted), and then the UI for hunt configuration could allow you to select from among [no guild, the guild currently configured, the list of guilds that your user account has granted the bot access]. There will be implications on the backend around DiscordCache data and its visibility; I think based on the structure of Discord's bot integrations we'd still need only one bot account & connection to Discord itself, but we'd have to be much more careful about what data we publish where.
>
[...]

PR #764, opened 2022-03-18
Move hunt creation/edit to a separate page

ebroder:
As we add more hunt configuration, we're rapidly outgrowing what's
reasonable to put into a relatively small modal. Given that, push it to
a separate page.

This is mostly pure copy-paste of the old modal content, but with a few
changes - namely handling breadcrumbs and URL parameters (instead of
having a hunt passed in), and slight changes to the post-submit flow.
Instead of closing the modal, we take different behavior depending on
whether we're creating a new hunt or editing an existing one. For
creation, we redirect to the newly created hunt. For edits, we scroll to
the bottom of the page and display a banner confirming that the edit was
successful, with a back button to take you back to the hunt list.

### DIFF
--- a/imports/client/components/HuntEditPage.tsx
+++ b/imports/client/components/HuntEditPage.tsx
@@ -1,0 +1,560 @@
+import { Meteor } from 'meteor/meteor';
+import { useSubscribe, useTracker } from 'meteor/react-meteor-data';
+import { faInfo } from '@fortawesome/free-solid-svg-icons/faInfo';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import React, { useCallback, useRef, useState } from 'react';
+import Alert from 'react-bootstrap/Alert';
+import Button from 'react-bootstrap/Button';
+import Col from 'react-bootstrap/Col';
+import Container from 'react-bootstrap/Container';
+import Form, { FormProps } from 'react-bootstrap/Form';
+import FormCheck from 'react-bootstrap/FormCheck';
+import FormControl, { FormControlProps } from 'react-bootstrap/FormControl';
+import FormGroup from 'react-bootstrap/FormGroup';
+import FormLabel from 'react-bootstrap/FormLabel';
+import FormText from 'react-bootstrap/FormText';
+import Row from 'react-bootstrap/Row';
+import { useNavigate, useParams } from 'react-router-dom';
+import Ansible from '../../Ansible';
+import DiscordCache from '../../lib/models/DiscordCache';
+import Hunts from '../../lib/models/Hunts';
+import Settings from '../../lib/models/Settings';
+import { BaseType } from '../../lib/schemas/Base';
+import { HuntType, SavedDiscordObjectType } from '../../lib/schemas/Hunt';
+import { SettingType } from '../../lib/schemas/Setting';
+import { useBreadcrumb } from '../hooks/breadcrumb';
+import ActionButtonRow from './ActionButtonRow';
+
+enum SubmitState {
+  IDLE = 'idle',
+  SUBMITTING = 'submitting',
+  SUCCESS = 'success',
+  FAILED = 'failed',
+}
+
+type HuntSubmit = Omit<HuntType, keyof BaseType>;
+
+const splitLists = function (lists: string): string[] {
+  const strippedLists = lists.trim();
+  if (strippedLists === '') {
+    return [];
+  }
+
+  return strippedLists.split(/[, ]+/);
+};
+
+interface DiscordSelectorParams {
+  disable: boolean;
+  id: string;
+  value: SavedDiscordObjectType | undefined;
+  onChange: (next: SavedDiscordObjectType | undefined) => void;
+}
+
+interface DiscordSelectorProps extends DiscordSelectorParams {
+  loading: boolean;
+  options: SavedDiscordObjectType[];
+}
+
+const DiscordSelector = ({
+  disable, id: formId, value, onChange, loading, options,
+}: DiscordSelectorProps) => {
+  const onValueChanged: FormControlProps['onChange'] = useCallback((e) => {
+    if (e.currentTarget.value === 'empty') {
+      onChange(undefined);
+    } else {
+      const match = options.find((obj) => { return obj.id === e.currentTarget.value; });
+      if (match) {
+        onChange(match);
+      }
+    }
+  }, [onChange, options]);
+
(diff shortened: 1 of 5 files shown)


ebroder:
For the record, I don't know that I'll want to keep the "single save button" design that we currently have as I start working through #630, since I think we may want to save each of the integrations separately. I opted to keep the behavior similar for now though and revisit later.

imports/client/components/HuntEditPage.tsx
```diff
@@ -0,0 +1,558 @@
...
+
+        <div ref={footer}>
+          {submitState === SubmitState.FAILED && <Alert variant="danger">{errorMessage}</Alert>}
+          {submitState === SubmitState.SUCCESS && (
+            <Alert variant="success" dismissible onClose={onSuccessDismiss}>
+              Hunt information successfully updated
+            </Alert>
+          )}
+
+          <Button variant="secondary" type="button" onClick={goBack}>Go back</Button>
```
zarvox, line 549: While the previous modal form needed a way to dismiss it, since getting to this page now involves a navigation, I feel like between the browser back button and the breadcrumbs we don't need an additional way to do this.  This is no longer a modal form; it no longer needs its own "dismiss" button.  Can we take it out?

imports/client/components/HuntEditPage.tsx
```diff
@@ -0,0 +1,558 @@
...
+          {' '}
+          <Button variant="primary" type="submit" disabled={disableForm}>Save</Button>
```
zarvox, line 551: Can we make this button align to the right edge of the form?  I think "bottom right" is where I see a majority of "submit" buttons in forms in general.

(And if there were multiple buttons, the default "do a useful thing" action should be the one aligned with the edge of the form, with other actions growing out towards the left as needed)

zarvox: On reflection: I notice we're pretty inconsistent with "forms have save buttons on the right edge", and probably a big part of that is that the button is easy to lose on the far right if the page is very wide.  As a thought for the future: we might be well-served by picking a reasonable `max-width` to apply to some of our other forms and then moving the buttons to the right edge for them too.  It'd certainly help a bit with the visual balance; we tend to be very left-heavy at present (I'm thinking especially of OwnProfilePage and SetupPage).

imports/client/components/HuntEditPage.tsx
ebroder: Haha. Honestly, I added the back button because I expected you to want the explicit navigation/nudge for post-save action. I don't mind dropping it

imports/client/components/HuntEditPage.tsx
ebroder: Yeah, +1 - I'll wait until we land #767 and then incorporate its action button class.

ebroder:
PTAL?

imports/client/components/HuntEditPage.tsx
```diff
@@ -0,0 +1,560 @@
...
+      {...rest}
+    />
+  );
+};
+
+const HuntEditPage = () => {
+  const huntId = useParams<{ huntId: string }>().huntId;
+  const hunt = useTracker(() => (huntId ? Hunts.findOne(huntId) : null), [huntId]);
+
+  useBreadcrumb({ title: huntId ? 'Edit Hunt' : 'Create Hunt', path: `/hunts/${huntId ? `${huntId}/edit` : 'new'}` });
```
zarvox, line 180: As written, this results in there being no clickable breadcrumb for `/hunts/new`, which feels a little odd to me.

For the case where we're creating a new hunt, I think I'd lightly prefer if we had two breadcrumbs here -- one "Hunts", linking to `/hunts`, and then maybe "Create" (at `/hunts/new`).  Thoughts?

imports/client/components/HuntEditPage.tsx
```diff
@@ -0,0 +1,560 @@
...
+
+  const onFormCallback = useCallback((error?: Error, newHuntId?: string) => {
+    if (error) {
+      setErrorMessage(error.message);
+      setSubmitState(SubmitState.FAILED);
+    } else {
+      setSubmitState(SubmitState.SUCCESS);
+      setErrorMessage('');
+
+      // If there's a result, that means we created a new hunt - redirect to it.
```
zarvox, line 260: Nice workflow optimization here :)

imports/client/components/HuntEditPage.tsx
```diff
@@ -0,0 +1,560 @@
...
+          {submitState === SubmitState.FAILED && <Alert variant="danger">{errorMessage}</Alert>}
+          {submitState === SubmitState.SUCCESS && (
+            <Alert variant="success" dismissible onClose={onSuccessDismiss}>
+              Hunt information successfully updated
+            </Alert>
+          )}
+
+          <ActionButtonRow>
+            <FormGroup>
+              <Button variant="primary" type="submit" disabled={disableForm}>Save</Button>
```
zarvox, line 551: Maybe label this button "Create" if we're in create mode and "Save" if we're in edit mode?

imports/client/components/HuntListPage.tsx
```diff
@@ -733,13 +207,9 @@ const HuntListPage = () => {
...
     <div id="jr-hunts">
       <h1>Hunts</h1>
-      <HuntModalForm
-        ref={addModalRef}
-        onSubmit={onAdd}
-      />
       {canAdd && (
         <>
-          <Button onClick={showAddModal} variant="success" size="sm" title="Add new hunt...">
+          <Button as={Link} to="/hunts/new" variant="success" size="sm" title="Add new hunt...">
```
zarvox, line 212: I wonder if this would be better as a button with text "Create new hunt..." instead of an icon.

zarvox (approved):
Thanks, this is good!  My only comments are a few tiny nits that I'm not particularly attached to, so I'll approve and leave you the option of merging as-is or tweaking and submitting.

imports/client/components/HuntEditPage.tsx
ebroder: I'll poke at this in a bit. The reason I didn't add a second breadcrumb initially is that you can't call hooks conditionally. I can try to modify `useBreadcrumb` to take `object | null` and not render the breadcrumb if null is passed in (similar to how `useSubscribe` works), but I'm not sure if that will actually work with `useBreadcrumb`'s weird state tracking.

ebroder: OK managed to get this working by just rearranging the route map a bit. I'll push in a sec

PR #767, opened 2022-03-22
Rework forms to limit width and put buttons right

zarvox:
Save buttons feel like they should always be at the bottom right of the
form they're modifying.  To keep them from getting lost, we should limit
the max-width of forms.

Large amounts of text (like that on the SetupPage) is also easier to
read when slightly narrower, and there's less of an argument for
information density on that page vs the PuzzleListPage, so I limited the
width of the content of each section..

I also took the opportunity to slightly tweak the design on SetupPage,
making the section headers fill the full width, making the toggle button
pairs into a reusable component, and adjusting some padding to avoid
buttons right on the edge of their containing element.

### DIFF
--- a/imports/client/components/OwnProfilePage.tsx
+++ b/imports/client/components/OwnProfilePage.tsx
@@ -6,6 +6,7 @@ import { ServiceConfiguration } from 'meteor/service-configuration';
 import React, { useCallback, useMemo, useState } from 'react';
 import Alert from 'react-bootstrap/Alert';
 import Button from 'react-bootstrap/Button';
+import Container from 'react-bootstrap/Container';
 import FormCheck from 'react-bootstrap/FormCheck';
 import FormControl, { FormControlProps } from 'react-bootstrap/FormControl';
 import FormGroup from 'react-bootstrap/FormGroup';
@@ -14,6 +15,7 @@ import FormText from 'react-bootstrap/FormText';
 import Flags from '../../Flags';
 import TeamName from '../TeamName';
 import { requestDiscordCredential } from '../discord';
+import ActionButtonRow from './ActionButtonRow';
 import AudioConfig from './AudioConfig';
 import Avatar from './Avatar';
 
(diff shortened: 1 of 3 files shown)


imports/client/components/OwnProfilePage.tsx
```diff
@@ -280,6 +282,10 @@ enum OwnProfilePageSubmitState {
   ERROR = 'error',
 }
 
+const PageContainer = styled.div`
+  max-width: 800px;
+`;
+
```
ebroder, line 288: Instead of defining our own class here and in `SectionContent`, thoughts on using Bootstrap's/`react-bootstrap`'s `Container`, which serves a similar function?

imports/client/components/SetupPage.tsx
```diff
@@ -48,14 +54,17 @@ const SectionHeaderButtons = styled.span`
...
   button {
     margin-left: 8px;
-    margin-bottom: 8px;
   }
 `;
 
+const SectionContent = styled.div`
+  max-width: 800px;
+`;
+
```
ebroder, line 63: I'm definitely not wild about the section content being max-width'd but the headers stretching all the way to the end of the screen - especially with the feature flag buttons, it looks kind of imbalanced on my screen:

Screen Shot 2022-03-23 at 8 34 44 PM

I think I'd rather just max-width the whole page?

ebroder (approved):
I'd mildly prefer that we lean towards Bootstrap's paradigms on pages where (as you point out) we're not optimizing for information density, and I'm pretty skeptical of the full-width section headers on the setup page, but I also don't feel super strongly about it so if you do I'm fine merging this (but will give you the ✅ and you can make the call)

imports/client/components/SetupPage.tsx
zarvox: Fair.  My first experiment with that approach felt like it was hard to pick out the sections at a glance, but 1) that was before I had worked out negative margins to extend the section headers into the gutter, and 2) I agree this looks sorta ridiculous at very large widths.

Also, in the fullness of time, it might be nice to have a miniature navbar/scrollspy thing at large screen sizes since we'll have the space for it and the setup page is super long.

imports/client/components/OwnProfilePage.tsx
zarvox: Hmmm.  I think the centering behavior might make sense in some contexts, and I could probably use it here.

For SetupPage specifically I think I want the max-width to be narrower than react-bootstrap's largest breakpoint of 960px.

PR #781, opened 2022-03-29
Replace `percolate:migrations` with our own mini library

zarvox:
`percolate:migrations` has a few impedence mismatches with how we think
about migrations, along with a significant bug that caused data
corruption alongside another correctness bug that makes me not really
want to rely on the library any more.

## The impedence mismatches:

* We don't believe in down() migrations.  Nobody wants to write them
  (empirically, we currently have zero down() implementations for our 39
  migrations), nobody tests them, and in production nobody is around to
  run them anyway.
* We don't believe in running migrations manually.  The operational cost
  is comparatively high as is the complexity of maintaining code that is
  able to correctly handle both the old and new forms of data.  We are
  willing to accept some brief transient inconsistency if we have
  backends running different versions of code with different
  expectations of data shapes to save ourselves the trouble of having to
  keep code both forward and backward compatible.  As a result, we allow
  the code to apply migrations and assume that data will very promptly
  converge to the shape the author expects.
* Relatedly, we *do* care about operators not having to manually touch
  databases.  We want jolly-roger to not require any particular database
  expertise nor familiarity with the codebase to operate a server
  successfully.
* To that end, we care a lot about our migration framework keeping data
  as consistent as possible.

## The data-corrupting bug:

When applying multiple migrations in one
`Migrations.migrateTo('latest')` call, if anything goes wrong with a
migration none of the migrations (including ones that completed
successfully!) will be recorded as having been completed.

This means any bug that ever causes a migration to fail not only fails
the migration, but leaves the database in an inconsistent state, such
that migrations that have run to completion (and other migrations that
may have made assumptions about lower-versioned migrations never running
again) will be run again the next time `migrateTo()` is called.

This is not a hypothetical bug; the fallout from the migrations bug
fixed by #659 and #661 trashed both my and @aldeka's personal servers,
and could only be recovered by editing their production databases by
hand (first to unlock the migrations control collection, and later to
skip over migrations that had already been run to completion that would
now fail given the current contents of the DB).

## The less-significant correctness bug:

Insertion of the control record in the migrations table for the first
time is done with an upsert and can race against other servers,
clobbering `version`.  Since `jolly-roger` aims to run in a clustered
environment, this also produces the potential for situations where a
particular migration is executed against the DB more than once, and
after higher-versioned migrations have run.

## This PR

So, this adds `MigrationRegistry`, a ~drop-in replacement library for
our subset of usage of `percolate:migrations` which:

* skips most of the runtime type-checking in favor of using TS natively
* only includes the API that we actually made use of
* avoids the bugs described above
* has some tests


imports/server/MigrationRegistry.ts
```diff
@@ -0,0 +1,200 @@
...
+  private ensureControlCreated() {
+    this.log('Migrations: creating control record at version 0');
+    // We use insert rather than upsert here
+    ignoringDuplicateKeyErrors(() => {
+      this.collection.insert({
+        _id: 'control',
+        version: 0,
+        locked: false,
+      });
+    });
```
ebroder, line 172: Huh, I'm surprised that this works. I thought that conflicts on `_id` threw a different error code than conflicts on secondary indexes, and I didn't think that I wrote `ignoringDuplicateKeyErrors` to catch both. Maybe Mongo finally fixed that.

imports/server/MigrationRegistry.ts
```diff
@@ -0,0 +1,200 @@
...
+    } else if (observedVersion > target) {
+      this.log(`Migrations: not migrating, observed version ${observedVersion} exceeds target ${target}`);
+      return false;
+    }
+    // by elimination, observedVersion < target.  We have migrations to run.
+    // Acquire the lock.  We might race with another backend that is also
+    // trying to apply this same migration.
+    if (!this.lock()) {
+      this.log('Not migrating, control is locked');
+      return false;
```
ebroder, line 107: If we're optimizing for not requiring administrative intervention, is it worth having a prevention mechanism in case the lock was acquired in the distant past? (Say, more than an hour ago)

imports/server/MigrationRegistry.ts
zarvox: It's entirely possible that it doesn't!  I should test this.

imports/server/MigrationRegistry.ts
ebroder: Yeah I don't think Mongo has anything of that shape. Historically, Mongo has shied away from semantics that are tied to connection state because it makes it more difficult to behave correctly in the event of a failover. I think there's a bit more these days (especially around the causal consistency features) but still nothing like that.

zarvox: Updated behavior here to preempt if the lock is >10 minutes old (and added a test that preemption works as expected).

tests/main.ts
```diff
@@ -1,6 +1,8 @@
 import './unit/imports/lib/calendarTimeFormat';
 import './unit/imports/lib/relativeTimeFormat';
-import './unit/imports/server/MigrationRegistry';
+if (Meteor.isServer) {
+  import './unit/imports/server/MigrationRegistry';
+}
```
ebroder, line 9: Do we need both this and the `if` in the test file?

zarvox: No.  I already removed the condition from the test file in 793b7236e7e11e283563af6865692eaf9e68f7fd though.

ebroder (approved):

PR #862, opened 2022-05-25
Upgrade to TypeScript 4.7

ebroder:
This is almost completely uneventful, but is newly strict about indexed
access to an empty object, so tweak a cast slightly.

~(Also requires taking some updates to the TypeScript/ESLint integration to pick up versions that are compatible with TS 4.7)~ Bah I thought my working copy was up to date but apparently it was not.

Supersedes #861

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1688,7 +1688,7 @@
     "buffer-crc32": {
       "version": "0.2.13",
       "resolved": "https://registry.npmjs.org/buffer-crc32/-/buffer-crc32-0.2.13.tgz",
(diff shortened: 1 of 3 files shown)


zarvox (approved):

package-lock.json
```diff
@@ -1688,7 +1688,7 @@
     "buffer-crc32": {
       "version": "0.2.13",
       "resolved": "https://registry.npmjs.org/buffer-crc32/-/buffer-crc32-0.2.13.tgz",
-      "integrity": "sha512-VO9Ht/+p3SN7SKWqcrgEzjGbRSJYTx+Q1pTQC0wrWqHx0vpJraQ6GtHx8tvcg1rlK1byhU5gccxgOgj7B0TDkQ==",
+      "integrity": "sha1-DTM+PwDqxQqhRUq9MO+MKl2ackI=",
```
zarvox, line 1691: weird that this switched back to sha1, but 🤷🏻‍♂️

PR #892, opened 2022-06-21
Add some additional files to .gitignore

ebroder:
The new Meteor Toolbox extension for VS Code seems to generate some
additional detritus.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 node_modules
 npm-debug.log
+jsconfig.json
+.vscode/launch.json


ebroder:
(Note that we do have a list of recommended extensions for VS Code in `.vscode/extensions.json`. For the time being, I'm not adding the Meteor Toolbox extension to that list because I'm not entirely convinced of its utility, but I do have it installed and thus have to deal with the detritus)

zarvox (approved):

PR #893, opened 2022-06-21
Enable @typescript-eslint typed linters

ebroder:
...and then turn several of them back off. A lot of the linters are
intended to tamp down use of `any`, which is admirable, but also
difficult to avoid in most of the places where we use it. (Although I
did go through and clean up some of the places where we didn't need to
rely on `any` first)

Most of the substantive changes were to account for the
`no-misused-promises` and `no-floating-promises` rules. This is because
in some cases we would invoke a promise as a way to fire off a
background task, without waiting for the result. For those cases, I
mostly used `void` casts, sometimes combined with IIFEs where necessary.


ebroder:
FTR let me know if you'd prefer I wait to tee this up until after #781 lands.

imports/client/components/PuzzleAnswer.tsx
```diff
@@ -38,9 +38,9 @@ const PuzzleAnswer = React.memo(({
     // Use Intl.Segmenter (stage 3 proposal) if available to properly segment grapheme clusters
     // Typescript is unaware of it, so there are a few any casts...
     let graphemes:string[];
-    if (Intl !== undefined && (Intl as any).Segmenter !== undefined) {
```
zarvox, line 41: Are these just places where upstream's type annotations got better, obviating our any casts?

imports/client/components/HuntEditPage.tsx
```diff
@@ -79,7 +79,7 @@ const DiscordSelector = ({
 
     if (value) {
       if (!options.find((opt) => {
-        return opt.id === value!.id;
```
zarvox, line 82: looks like inference got smarter, nice

imports/client/components/SetupPage.tsx
```diff
@@ -4,7 +4,6 @@ import { Meteor } from 'meteor/meteor';
 import { OAuth } from 'meteor/oauth';
 import { useSubscribe, useTracker } from 'meteor/react-meteor-data';
 import { ServiceConfiguration } from 'meteor/service-configuration';
-import { _ } from 'meteor/underscore';
```
zarvox, line 7: Heh, looks like that's been unused since da6579de9cfa172308bde273bdf9f7c6519d75c6.  Do we have systematic linting for unused imports now, or was this something that VSCode flagged for you that you happened to clean up while you were here?

imports/client/components/SetupPage.tsx
```diff
@@ -459,7 +458,7 @@ const FeatureToggle = ({ enabled, onToggleEnabled }: {
 
 const GoogleIntegrationSection = () => {
   const enabled = useTracker(() => !Flags.active('disable.google'), []);
-  const oauthSettings = useTracker(() => ServiceConfiguration.configurations.findOne({ service: 'google' }) as any, []);
+  const oauthSettings = useTracker(() => ServiceConfiguration.configurations.findOne({ service: 'google' }) as unknown as { clientId: string }, []);
```
zarvox, line 461: Should this be `as unknown as undefined | { clientId: string }`?  There's no guarantee we've got a record in the DB to retrieve for `service: 'google'`, so we shouldn't assert that it's always going to return an object.

imports/lib/models/Base.ts
```diff
@@ -53,11 +53,10 @@ class Base<T extends BaseType> extends Mongo.Collection<T> {
     });
   }
 
-  // @ts-ignore Because the Mongo.Collection doesn't know about SimpleSchema
```
zarvox, line 56: Huh, did this get smarter?

imports/lib/schemas/typedSchemas.ts
```diff
@@ -249,12 +249,12 @@ export const inheritSchema = function <
   childSchemaCodec: t.InterfaceType<CP, CT>,
   parentOverrides: Overrides<PT>,
   childOverrides: Overrides<CT>
-): [t.TypeC<PP & CP>, Overrides<PT & CP>] {
+): [t.TypeC<PP & CP>, Overrides<PT & CT>] {
```
zarvox, line 252: Nice catch

imports/server/api_keys.ts
```diff
@@ -21,7 +21,7 @@ Meteor.methods({
     check(this.userId, String);
     check(forUser, Match.Optional(String));
 
-    const user = userForKeyOperation(<string> this.userId, forUser);
+    const user = userForKeyOperation(this.userId, forUser);
```
zarvox, line 24: Oh nice -- I guess the `check()` above indicates that if `this.userId` is falsey that we diverge, so `this.userId` is always `string`?

imports/server/discordClientRefresher.ts
```diff
@@ -113,7 +113,7 @@ class DiscordClientRefresher {
             // If we get the lock, we're responsible for opening the websocket
             // gateway connection
             const ready = new Promise<void>((r) => client.on('ready', r));
-            client.login(this.token);
+            Promise.await(client.login(this.token));
```
zarvox, line 116: That's another correctness fix, nice

imports/server/hooks/DiscordHooks.ts
```diff
@@ -133,10 +133,7 @@ const DiscordHooks: Hookset = {
         },
       };
 
-      Meteor.defer(() => {
-        // send actual message
-        bot.postMessageToChannel(channel, msg);
-      });
+      void bot.postMessageToChannel(channel, msg);
```
zarvox, line 136: Oh, I guess `bot.postMessageToChannel` was already async, so we didn't need to Meteor.defer it, just kick off the Promise and then ignore it (with the `void` here)?

package-lock.json
```diff
@@ -1688,13 +1688,13 @@
     "buffer-crc32": {
       "version": "0.2.13",
       "resolved": "https://registry.npmjs.org/buffer-crc32/-/buffer-crc32-0.2.13.tgz",
-      "integrity": "sha512-VO9Ht/+p3SN7SKWqcrgEzjGbRSJYTx+Q1pTQC0wrWqHx0vpJraQ6GtHx8tvcg1rlK1byhU5gccxgOgj7B0TDkQ==",
+      "integrity": "sha1-DTM+PwDqxQqhRUq9MO+MKl2ackI=",
```
zarvox, line 1691: Not that I think it matters much in practice, but: why, in the year of our Linux on the Desktop 2022 are these going back to sha1?

Maybe try nuking your npm cache via the steps at https://dnasir.com/2018/02/12/solving-the-package-lock-json-integrity-discrepancy/ ?

tests/acceptance/smoke.tsx
```diff
@@ -126,7 +126,7 @@ if (Meteor.isClient) {
 
     describe('which are unauthenticated', function () {
       before(async function () {
-        await Meteor.wrapPromise(Meteor.logout);
+        await Meteor.wrapPromise(Meteor.logout)();
```
zarvox, line 129: Huh, what was the impact before this change here?  Were we never actually logging out here?  (I vaguely recall we avoid testing routes with a token in them like account-enrollment and password reset, so only `/login` and `/create-first-user` were even candidates for running here)

zarvox (commented):
This is mostly me marvelling at how the typechecker has gotten smarter over time and how we can clean up a bunch of things that were previously too complicated, but there's also a couple of questions for my edification and one plausible bug.

I missed reading through this until after I'd landed #781 already (sorry!), so this'll need a rebase, but I'm overall bullish on the changes.  Thanks for tracking down my bug and thanks for adding ever-more guardrails!

imports/client/components/PuzzleAnswer.tsx
ebroder: I suspect that we upgraded Typescript since this code was written, and the later versions of TS came with native declarations for the newer proposal where the older versions did not.

imports/client/components/SetupPage.tsx
ebroder: This got picked up my `@typescript-eslint/no-unused-vars`, which is just a version of the standard `no-unused-vars` that works better with TS.

imports/lib/models/Base.ts
ebroder: I'm not entirely sure why we ever needed it, since we're already casting to `any` all over the place

imports/server/api_keys.ts
ebroder: Yeah. I augmented the type declarations for `check` a while back so it acts as a type assertion: DefinitelyTyped/DefinitelyTyped#41414, which means that TS knows enough to narrow the type. I think this is another case where the typecast predates that change and has been unnecessary for a while.

imports/server/discordClientRefresher.ts
ebroder: The `await` on `ready` on the next line will effectively incorporate what this is `await`ing on, so I think it's technically not a correctness fix, but I figured it was safer to use the `void` cast throwaway sparingly when it wasn't necessary.

imports/server/hooks/DiscordHooks.ts
ebroder: Yeah it fires off an HTTP request and then returns a promise - though that might be new since the Meteor `HTTP` library was deprecated in favor of using `fetch`.

tests/acceptance/smoke.tsx
ebroder: Yeah we would have never actually logged out

ebroder:
PTAL? (No worries about landing #781 first - the rebase was pretty straightforward)

zarvox (approved):

zarvox:
The tests are hitting the DDP rate limiter as a result of doing repeated logins.  I observed that occasionally locally when I was doing the migrations testing, but considered it orthogonal.   https://github.com/meteor/meteor/blob/05562e5a6005ff7bd355cb9696474ad5c6bf9968/packages/accounts-base/accounts_server.js#L1367-L1380 is the source for why.

PR #911, opened 2022-07-02
Experiment with using ValidatedMethod

ebroder:
@zarvox I'm curious if you think this is a good idea. If so I'll work on converting the rest of our methods.

mdg:validated-method gives us a handful of affordances that are missing in our current method declaration setup, and I think they're probably worth adopting. Most useful for us is that methods can be invoked as functions with named arguments and types on arguments and return values instead of as strings with unknown positional arguments.

There are some drawbacks - we lose some typechecker-enforced assurances about argument validation. Our previous house style declared the types of all method arguments as `unknown` and relied on `check` functions to make type assertions. There's no similar construct here - we can still declare the arguments to `validate` as type `unknown`, and we can rely on `audit-argument-checks` to make sure we're not using parameters without some form of validation, but nothing connects the `check` functions we call in `validate` against the types we declare in `run`, so it's possible we could get the types wrong.

(We could call our `check`s in the `run` function, but then the declared type when calling the function would be `unknown`, and I think that's worse.)

In practice I think the overall utility is high enough to be worth that risk.

### DIFF
--- a/imports/client/components/AnnouncementsPage.tsx
+++ b/imports/client/components/AnnouncementsPage.tsx
@@ -11,6 +11,7 @@ import Announcements from '../../lib/models/Announcements';
 import { indexedDisplayNames } from '../../lib/models/MeteorUsers';
 import { userMayAddAnnouncementToHunt } from '../../lib/permission_stubs';
 import { AnnouncementType } from '../../lib/schemas/Announcement';
+import postAnnouncement from '../../methods/postAnnouncement';
 import { useBreadcrumb } from '../hooks/breadcrumb';
 import useSubscribeDisplayNames from '../hooks/useSubscribeDisplayNames';
 import markdown from '../markdown';
@@ -50,10 +51,10 @@ const AnnouncementForm = ({ huntId }: { huntId: string }) => {
     setMessage(event.target.value);
   }, []);
 
-  const postAnnouncement = useCallback(() => {
+  const postAnnouncementCb = useCallback(() => {
     if (message) {
       setSubmitState(AnnouncementFormSubmitState.SUBMITTING);
-      Meteor.call('postAnnouncement', huntId, message, (error?: Error) => {
+      postAnnouncement.call({ huntId, message }, (error) => {
         if (error) {
           setErrorMessage(error.message);
           setSubmitState(AnnouncementFormSubmitState.FAILED);
@@ -78,7 +79,7 @@ const AnnouncementForm = ({ huntId }: { huntId: string }) => {
       <div className="button-row">
         <Button
           variant="primary"
-          onClick={postAnnouncement}
+          onClick={postAnnouncementCb}
           disabled={submitState === 'submitting'}
         >
           Send
--- a/imports/methods/TypedMethod.ts
+++ b/imports/methods/TypedMethod.ts
@@ -1,0 +1,75 @@
+import { check } from 'meteor/check';
+import { EJSONable, EJSONableProperty } from 'meteor/ejson';
+import { Meteor } from 'meteor/meteor';
+
+type TypedMethodParam = EJSONable | EJSONableProperty
+type TypedMethodArgs = Record<string, TypedMethodParam>;
+type TypedMethodValidator<Arg extends TypedMethodArgs> =
+  (this: Meteor.MethodThisType, arg0: unknown) =>
+    Arg;
+type TypedMethodRun<Arg extends TypedMethodArgs, Return extends TypedMethodParam | void> =
+  (this: Meteor.MethodThisType, arg0: Arg) => Return;
+
+export default class TypedMethod<
+  Args extends TypedMethodArgs,
+  Return extends TypedMethodParam | void,
+> {
+  name: string;
+
+  definition?: {
+    validate: TypedMethodValidator<Args>,
+    run: TypedMethodRun<Args, Return>,
+  };
+
+  constructor(name: string) {
+    check(name, String);
+
+    this.name = name;
+  }
+
+  define({ validate, run }: {
+    validate: TypedMethodValidator<Args>,
+    run: TypedMethodRun<Args, Return>,
+  }) {
+    if (this.definition) {
+      throw new Error(`TypedMethod ${this.name} has already been defined`);
+    }
+
+    this.definition = { validate, run };
+
+    /* eslint-disable-next-line @typescript-eslint/no-this-alias -- we need to
+       capture both the TypedMethod object (so we can find the implementation)
+       and the method invocation context (for things like userId) */
+    const self = this;
+    Meteor.methods({
+      /* eslint-disable-next-line meteor/audit-argument-checks -- because we're
+        wrapping the method call, the eslint check can't tell what's going on,
+        but the actual Meteor audit-argument-checks package will still fire if
+        we don't use check methods. */
+      [this.name](arg0: Args) {
+        self.execute(this, arg0);
+      },
+    });
+  }
+
+  execute(context: Meteor.MethodThisType, arg0: Args): Return {
+    if (!this.definition) {
+      throw new Error(`TypedMethod ${this.name} has not been defined`);
+    }
+
+    const validatedArgs = this.definition.validate.bind(context)(arg0);
+    const result = this.definition.run.bind(context)(validatedArgs);
+    return result;
+  }
+
+  call(
+    arg0: Args,
+    callback?: (error: Meteor.Error, result: Return) => void,
+  ): void {
+    return Meteor.call(this.name, arg0, callback);
+  }
+
+  callPromise(arg0: Args): Promise<Return> {
+    return Meteor.callPromise(this.name, arg0);
+  }
+}
--- a/imports/server/announcements.ts
+++ b/imports/server/announcements.ts
@@ -1,37 +1,11 @@
 import { check } from 'meteor/check';
 import { Meteor } from 'meteor/meteor';
-import Ansible from '../Ansible';
 import Announcements from '../lib/models/Announcements';
 import MeteorUsers from '../lib/models/MeteorUsers';
 import PendingAnnouncements from '../lib/models/PendingAnnouncements';
-import { userMayAddAnnouncementToHunt } from '../lib/permission_stubs';
 import JoinPublisher from './JoinPublisher';
 
 Meteor.methods({
-  postAnnouncement(huntId: unknown, message: unknown) {
-    check(this.userId, String);
-    check(huntId, String);
-    check(message, String);
-
-    if (!userMayAddAnnouncementToHunt(this.userId, huntId)) {
-      throw new Meteor.Error(401, `User ${this.userId} may not create annoucements for hunt ${huntId}`);
-    }
-
-    Ansible.log('Creating an announcement', { user: this.userId, hunt: huntId, message });
-    const id = Announcements.insert({
-      hunt: huntId,
-      message,
-    });
-
-    MeteorUsers.find({ hunts: huntId }).forEach((user) => {
-      PendingAnnouncements.insert({
-        hunt: huntId,
-        announcement: id,
-        user: user._id,
-      });
-    });
-  },
-
   dismissPendingAnnouncement(pendingAnnouncementId: unknown) {
     check(this.userId, String);
     check(pendingAnnouncementId, String);
--- a/imports/methods/postAnnouncement.ts
+++ b/imports/methods/postAnnouncement.ts
@@ -1,0 +1,5 @@
+import TypedMethod from './TypedMethod';
+
+export default new TypedMethod<{ huntId: string, message: string }, void>(
+  'Announcements.methods.post'
+);
--- a/imports/server/methods/postAnnouncement.ts
+++ b/imports/server/methods/postAnnouncement.ts
@@ -1,0 +1,40 @@
+import { check } from 'meteor/check';
+import { Meteor } from 'meteor/meteor';
+import Ansible from '../../Ansible';
+import Announcements from '../../lib/models/Announcements';
+import MeteorUsers from '../../lib/models/MeteorUsers';
+import PendingAnnouncements from '../../lib/models/PendingAnnouncements';
+import { userMayAddAnnouncementToHunt } from '../../lib/permission_stubs';
+import postAnnouncement from '../../methods/postAnnouncement';
+
+postAnnouncement.define({
+  validate(arg: unknown) {
+    check(this.userId, String);
+    check(arg, {
+      huntId: String,
+      message: String,
+    });
+
+    return arg;
+  },
+
+  run({ huntId, message }) {
+    if (!userMayAddAnnouncementToHunt(this.userId, huntId)) {
+      throw new Meteor.Error(401, `User ${this.userId} may not create announcements for hunt ${huntId}`);
+    }
+
+    Ansible.log('Creating an announcement', { user: this.userId, hunt: huntId, message });
+    const id = Announcements.insert({
+      hunt: huntId,
+      message,
+    });
+
+    MeteorUsers.find({ hunts: huntId }).forEach((user) => {
+      PendingAnnouncements.insert({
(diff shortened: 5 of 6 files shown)


imports/methods/announcements/post.ts
```diff
@@ -0,0 +1,38 @@
...
+import { ValidatedMethod } from 'meteor/mdg:validated-method';
+import { Meteor } from 'meteor/meteor';
+import Ansible from '../../Ansible';
+import Announcements from '../../lib/models/Announcements';
+import MeteorUsers from '../../lib/models/MeteorUsers';
+import PendingAnnouncements from '../../lib/models/PendingAnnouncements';
+import { userMayAddAnnouncementToHunt } from '../../lib/permission_stubs';
+
+export default new ValidatedMethod({
+  name: 'Announcements.methods.post',
```
ebroder, line 11: Note that I've changed the name of the method here. This format matches the style of the [module documentation](https://github.com/meteor/validated-method#using-simpleschema) and feels a little cleaner to me. Their example also names the function (or in my case, module) after the last component of the string name.

tsconfig.json
```diff
@@ -24,7 +24,7 @@
     "baseUrl": ".",
     "moduleResolution": "node",
     "resolveJsonModule": true,
-    "types": ["node", "mocha"],
+    "types": ["node", "mocha", "@types/meteor-mdg-validated-method"],
```
ebroder, line 27: TypeScript seems unable to find this automatically because it's in `node_modules/@types/meteor-mdg-validated-method` instead of, like, `node_modules/@types/meteor/mdg:validated-method` or something.

imports/server/assets.ts
```diff
@@ -14,14 +14,6 @@ import UploadTokens from './models/UploadTokens';
...
 import { BlobType } from './schemas/Blob';
 
-// Meteor has no way to import Assets under ES6 modules; it's just always a
-// global that all server code can access.
-declare global {
-  const Assets: {
-    absoluteFilePath(assetPath: string): string | undefined;
-  };
-}
-
```
ebroder, line 24: This is declared upstream: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4f4d4c59bbcf562c961b403ee5868f31ed982704/types/meteor/globals/tools.d.ts#L57

I'm not sure why we ever needed an ambient type declaration for it (it seems like it's been there forever), but TS seems happy to pick it up now.

imports/methods/announcements/post.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  validate({ huntId, message }: { huntId: unknown, message: unknown }) {
```
zarvox, line 13: One thing that I liked about the previous approach was that by checking each positional argument (and having to have the number match to have your callback get triggered), it made it very hard for a malicious client to smuggle data into the methods consumed by the server, by virtue of the fact that if the server didn't declare the positional argument, then additional garbage would be dropped before reaching the body of the method call.

It seems we lose that a little under this approach, by virtue of the fact that an object can contain additional garbage.  It's mitigated by the fact that when we do object destructuring we also tend to ignore unknown fields, so long as we never ever use `...rest` in any of our method signatures.  Maybe we can lint for that?  Or as noted in the top-level comment, passing the types through all the way would eliminate this entirely.

imports/methods/announcements/post.ts
```diff
@@ -0,0 +1,38 @@
...
+    check(this.userId, String);
+    check(huntId, String);
+    check(message, String);
+
+    if (!userMayAddAnnouncementToHunt(this.userId, huntId)) {
+      throw new Meteor.Error(401, `User ${this.userId} may not create announcements for hunt ${huntId}`);
+    }
+  },
+
+  run({ huntId, message }: { huntId: string, message: string }) {
```
zarvox, line 23: I don't love that we write the method signature twice here, once for validation purposes, and once for "actually do the thing" purposes, and (because of the limitations in the approach) they actually declare different types and there's no validation that the result of `validate()` is what we're passing on to `run()`.  It's creating a new possible class of "these things do not match" errors that is likely to escape detection by our tooling, which doesn't feel like much of a step forward to me.

imports/methods/announcements/post.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export default new ValidatedMethod({
+  name: 'Announcements.methods.post',
+
+  validate({ huntId, message }: { huntId: unknown, message: unknown }) {
+    check(this.userId, String);
+    check(huntId, String);
+    check(message, String);
+
+    if (!userMayAddAnnouncementToHunt(this.userId, huntId)) {
```
zarvox, line 18: So we have a number of slightly-more-complicated situations where we might need to:

* validate types of arguments
* retrieve an object or two from the database (puzzle/hunt)
* consider a field from that object and reject the request based on the current hunt configuration
* do something else based on other fields from the object

(example: [addGuessForPuzzle](https://github.com/deathandmayhem/jolly-roger/blob/main/imports/server/guesses.ts#L151-L197), where we fetch Puzzle, Hunt, and then consult `hasGuessQueue`)

Which of these checks would you imagine belong in `validate` vs in `run`?  If we expect all of the "client sent us something wrong" error classes to surface in `validate`, then we might have to fetch things (example: `Puzzle`) from the DB twice, which feels a little silly and permits TOCTOU (but we probably do that in a number of places anyway).  If the checking can be split between the methods, I wonder how much value we're actually getting from the validate/run split.

zarvox (commented):
I think I'm -1 on this overall as-is.

Some things I'm neutral or positive on:

I'm down with reworking our method naming scheme.
I'm favorable towards writing the method name string once and importing code to reference some object precisely.
I'm gently in favor of moving away from positional arguments in favor of objects with named args everywhere.

The things I'm wary or negative on:

I'm pretty hesitant about the unfortunate split, both type-wise and implementation-wise, between `validate` and `run`.

I wonder if we can do something slightly smarter, where `ValidatedMethod<T>`'s `validate()` returns precisely the type `T` that `run(args: T)` accepts as its argument?  Maybe more of a "parse" than a "validate".  Then we could potentially have the typechecks flow through.

I'm also deeply skeptical about client-side method simulation/optimistic behavior in general, which this library seems somewhat focused on.  My experience is that simulation makes it hard for people to have a clear expectation around when their changes have actually been reflected serverside (and are durable, and it's safe to close the tab or navigate away, and they should expect other users views to reflect their change), as well as complicating error handling/failure paths to the point that failures simply get ignored or silenced.  Totally fine if making methods also be installed client-side wasn't your goal here, but I think that's perhaps one of the reasons that ValidatedMethod as written is structured the way it is, and it's generally an antigoal of mine.

I'm not sure if there were other features of this library that you found compelling that didn't get demoed here, but I think if the primary reason we're using this library is to reduce the risk of errors made in writing method implementations and callsites, I'm not convinced this is really a net win as it stands.  The newly-introduced potential failure modes seem about as sharp as the old ones, and less well-tooled/linted against.

Were there particular failure modes or potential failure modes or ergonomics problems you were hoping this would address?  Happy to brainstorm on other/related approaches to solving those challenges.

ebroder:
This is helpful feedback. I'm not especially attached to the specific implementation of `mdg:validated-method`; I just in general try to use code that already exists before writing my own. I just pushed up a new implementation that uses my own `TypedMethod` implementation. It's not quite perfect (in particular, I haven't gotten the type inference working quite like I want, so you have to pass in generic parameters explicitly), but I think it does the things I care about and you might find it less objectionable.

imports/methods/announcements/post.ts
```diff
@@ -0,0 +1,40 @@
+import { check } from 'meteor/check';
+import { Meteor } from 'meteor/meteor';
+import Ansible from '../../Ansible';
+import Announcements from '../../lib/models/Announcements';
+import MeteorUsers from '../../lib/models/MeteorUsers';
+import PendingAnnouncements from '../../lib/models/PendingAnnouncements';
+import { userMayAddAnnouncementToHunt } from '../../lib/permission_stubs';
+import TypedMethod from '../TypedMethod';
+
+export default new TypedMethod<{ huntId: string, message: string }, void>({
```
ebroder, line 10: Roughly speaking, if I don't pass the generic parameters explicitly, the compiler infers the type of `Args` and `Return` as their generic values (`Record<string, TypedMethodParam>` and `TypedMethodParam`) rather than inferring something more specific.

It seems like it should be possible to declare this in a way that the inference does work out, but I haven't gotten it yet. (I suspect it involves something like having a separate `Keys extends string` parameter with `validate` taking a `Record<Keys, unknown>` but I wasn't able to get that working either)

imports/methods/announcements/post.ts
ebroder: FWIW in this revision, I'm taking the stance that only type validation of the parameters happens in `validate` and shifted the permission check into `run`.

zarvox: That seems like the easiest thing to keep consistent. 👍🏻

imports/methods/TypedMethod.ts
```diff
@@ -0,0 +1,67 @@
...
+    this.name = name;
+    this.validate = validate;
+    this.run = run;
+
+    /* eslint-disable-next-line @typescript-eslint/no-this-alias -- we need to
+       capture both the TypedMethod object (so we can find the implementation)
+       and the method invocation context (for things like userId) */
+    const self = this;
+    Meteor.methods({
+      /* eslint-disable-next-line meteor/audit-argument-checks -- because we're
```
zarvox, line 41: Just to make sure I understand correctly -- you're saying the audit-argument-checks lint will no longer fire, but the runtime checks will still complain serverside, throwing a Meteor.Error if we try to access fields without first check()-ing them?

imports/methods/announcements/post.ts
zarvox: Honestly, this feels mostly fine to me?  There's exactly one place that you are declaring what this method takes and what it returns and what its name is, which feels like keeping the API of a method all in one place.

I suppose it'd be neat if you could get everything to be inferred from the `check()` calls, but I'm not sad about having types for function signatures written down as clear API documentation, especially since we get the typechecker verifying (via our `check`s) that we're delivering on our claims.

imports/methods/TypedMethod.ts
```diff
@@ -0,0 +1,67 @@
...
+         wrapping the method call, the eslint check can't tell what's going on,
+         but the actual Meteor audit-argument-checks package will still fire if
+         we don't use check methods. */
+      [this.name](arg0: Args) {
```
zarvox, line 45: You can potentially punt on solving this until you get to the first method implementation that does a server-only import or looks at a server-only collection, but I suspect eventually we're going to get to a situation where there's an import chain:

* client/main -> 
* something in imports/client/components/* (which wants to make a `call`) -> 
* something in imports/methods/* ->
* something in imports/server/* (where `run` accesses some server-only collection or imports some server-only code)

and then the build tool will be like "you can't import a thing under server from client" and there might not be a good place to do a conditional import.

`reorganizeGoogleDrive` and `setupGetUploadToken` are both examples of methods that will require server-only imports, but which will have the client code importing the file under `methods/` -- maybe try porting one of those as a pressure test?

In turn, that may prompt us to think about if we want to always be defining these methods on both client-side and server-side, or if we'd be better off keeping the bundle smaller and not shipping method implementations to the client that are not useful to the user experience.

imports/server/announcements.ts
```diff
@@ -1,37 +1,11 @@
...
 import { Meteor } from 'meteor/meteor';
-import Ansible from '../Ansible';
 import Announcements from '../lib/models/Announcements';
 import MeteorUsers from '../lib/models/MeteorUsers';
 import PendingAnnouncements from '../lib/models/PendingAnnouncements';
-import { userMayAddAnnouncementToHunt } from '../lib/permission_stubs';
 import JoinPublisher from './JoinPublisher';
 
 Meteor.methods({
-  postAnnouncement(huntId: unknown, message: unknown) {
```
zarvox, line 11: If I'm reading this right, the server bundle never imports `imports/methods/announcements/post`, and thus no method is defined on the server, which seems wrong?  Am I missing something, or is this PR missing a file or an import somewhere, or does this just not work as written?

zarvox (commented):
Much more positive on this overall approach!

I think there are still some potential rough edges around imports that we should figure out, but I like the overall feel of defining the API between client and server, and the way that the type signatures give additional linting of the "checkpoint" type between `validate` and `run`.  More details on my outstanding questions inline.

imports/methods/TypedMethod.ts
ebroder: Yeah, that's right. `audit-argument-checks` exists as both a linter and a Meteor package. The linter statically inspects any method that's passed to `Meteor.methods` to ensure that variables get `check`'d. That's not useful here since we're effectively establishing a wrapper. However, disabling the linter doesn't do anything to the (older) runtime check, which will still fire based on the presence of the Meteor package `audit-argument-checks`.

(It would be _optimal_ if we could (say) apply the linter to the `validate` method of `TypedMethod`, but the linter is written pretty narrowly, so I don't think there's any way to plug that in.)

imports/server/announcements.ts
ebroder: Yeah that's a mistake and I'm missing the import.

imports/methods/TypedMethod.ts
ebroder: Hmm. It turns out I did not entirely understand this correctly. The `audit-argument-checks` Meteor package does object identity-based tracking of [if you've looked at the top-level argument](https://github.com/meteor/meteor/blob/9530de1ac97f7736bfa7caeffda6a756603919fd/packages/check/match.js#L470-L517), so checking a destructured object doesn't count. So we can't use both destructuring and `audit-argument-checks`, which is unfortunate.

imports/methods/TypedMethod.ts
ebroder: Pushed a new version that does this, with a strawman structure.

imports/client/components/AnnouncementsPage.tsx
```diff
@@ -53,7 +54,7 @@ const AnnouncementForm = ({ huntId }: { huntId: string }) => {
   const postAnnouncement = useCallback(() => {
     if (message) {
       setSubmitState(AnnouncementFormSubmitState.SUBMITTING);
-      Meteor.call('postAnnouncement', huntId, message, (error?: Error) => {
+      post.call({ huntId, message }, (error) => {
```
zarvox, line 57: Reflecting on this naming scheme -- I wonder if having these reduce to a single verb is going to make it harder to track semantically.  If this were Go and you always had to invoke things by `module.verb()`, then it'd be less ambiguous, but I'm thinking about what happens if we follow this to its logical conclusion and wind up with `Hunt.methods.create` and `Puzzle.methods.create` and `Tags.methods.create` and all of them want to be called `create` in the same file.

Maybe we should aim to name things as `verbNoun` instead of just `verb`?  That adds a little more context as the usage site.  If we're doing that, I'd also consider if it makes sense to name the method file `postAnnouncement` for consistency, though that makes the imports do a bit of repetition.

I guess another option is we wrap up the calls on objects onto some facade object and then maybe it's `AnnouncementMethods.post.call`.

imports/methods/TypedMethod.ts
zarvox: Ah, so to be clear -- we'd have to turn off `audit-argument-checks` for the whole program to use destructuring anywhere?  I guess that's still fine if we don't mind doing the migration all-at-once.  I guess the other option is not using destructuring in `validate()` and `check()`ing the top-level object mid-migration, but if we do that I'd want to be extra careful about rejecting unknown keys.

imports/client/components/AnnouncementsPage.tsx
ebroder: Yeah I think I agree. I'll switch to that.

zarvox (approved):
This looks great -- thanks for seeking feedback and taking a few rounds of iteration to get things working and feeling right!

PR #944, opened 2022-07-20
Keep previous `answers` value on Puzzles update

### LINKED ISSUE #943: Changing tags on a solved puzzle clears its denormed `answers`
Repro steps:

* set up a jolly-roger instance
* create the fixture hunt
* from the puzzle list page, pick any solved puzzle, click the edit button to edit its properties
* add a new tag `anything` and click save
* observe that the puzzle now has an `answers` value of `[]`, causing it to appear unsolved in grey on the PuzzleListPage (though the `Guess` objects, which drive the PuzzlePage view, are all still correct)

The server-side code performing the update is: https://github.com/deathandmayhem/jolly-roger/blob/main/imports/server/puzzle.ts#L133-L136 

I expected that a Mongo query with a `$set` would only modify the fields set, leaving any other fields untouched, but it appears that `autoValue()` seems to be called (as confirmed by some locally-added console logs), producing `answers: []`.

I suppose I can work around this by tossing in an `answers: oldPuzzle.answers,` in the modifier object, but I wanted to flag that I find the `autoValue` behavior here unfortunate.  I'm not sure if this is a recent regression or if this is longstanding behavior that we missed because we mostly don't change tags after solving puzzles.

Maybe the answer is that we should avoid `autoValue()` if we ever want the field to have a non-automatically-computed value?

zarvox:
c3ffbee79cc53ea2eb184a8f99c76e0cc4ffa4af seems to have introduced a
regression in which every subsequent update to a `Puzzle` that does not
specify an explicit value for `answers` winds up causing `answers` to
be set to `[]`, stripping the denormalized answer values, which leads
PuzzleListPage to render it as unsolved (since PuzzleListPage relies on
the denormed `answers` rather than `Guess`es.

The real purpose of this `autoValue` implementation is to uppercase all
the answers.  We do not rely on it to backfill `answers: []` in the
event a caller did not provide that key.  To that end, we should return
`undefined` rather than `[]` from our `autoValue` implementation.

We can also take this opportunity to remove the support for converting a
single string value into an array.  Migrations 20 and 21 have backfilled
arrays in all extent databases, so we can make the behavior tighter,
throwing at runtime if code uses the interface incorrectly rather than
silently changing data shape.

Doing so reveals that the `AutoValueFlatten` from `typedSchemas` appears
to no longer be needed, and produces issues around the type of the
`this` argument passed to the `autoValue` function if we want to have
`answers` be treated always as an array of strings rather than possibly
a single string value.  So this patch simply removes AutoValueFlatten,
both on the argument and the return value side of our `autoValue` type
signature.

Fixes #943.

### DIFF
--- a/imports/lib/schemas/Puzzle.ts
+++ b/imports/lib/schemas/Puzzle.ts
@@ -36,7 +36,7 @@ const PuzzleFieldsOverrides: Overrides<t.TypeOf<typeof PuzzleFields>> = {
         return this.value.map((x) => answerify(x));
       }
 
-      return [];
+      return undefined;
     },
   },
   replacedBy: {


ebroder:
> Doing so reveals that the `AutoValueFlatten` from `typedSchemas` appears to no longer be needed, and produces issues around the type of the `this` argument passed to the `autoValue` function if we want to have `answers` be treated always as an array of strings rather than possibly a single string value. So this patch simply removes AutoValueFlatten, both on the argument and the return value side of our `autoValue` type signature.

Is it possible that this comes into play with `$push` et al. operators? It doesn't feel like something I would have written as a schema upgrade consideration.

zarvox:
> > Doing so reveals that the `AutoValueFlatten` from `typedSchemas` appears to no longer be needed, and produces issues around the type of the `this` argument passed to the `autoValue` function if we want to have `answers` be treated always as an array of strings rather than possibly a single string value. So this patch simply removes AutoValueFlatten, both on the argument and the return value side of our `autoValue` type signature.
> 
> Is it possible that this comes into play with `$push` et al. operators? It doesn't feel like something I would have written as a schema upgrade consideration.

Ahhhh, I think you're probably right.  I'll back that part of the patch out and just swap out `[]` and `undefined`.

ebroder (approved):

PR #949, opened 2022-07-22
Remove gitRevision method (and associated machinery)

ebroder:
This was originally introduced for debugging purposes, but Meteor
includes a built-in primitive accomplishing the same goal
(`Meteor.gitCommitHash`). There's a small semantic difference between
the two (our revision always returned the running code on the server,
while `Meteor.gitCommitHash` returns the code on the client), but given
how Meteor's autorefresh works in production, they should always be the
same.


imports/server/git_revision.ts
```diff
@@ -1,14 +0,0 @@
-import { Meteor } from 'meteor/meteor';
```
zarvox, line 1: Don't we need to remove the import of this from `server/main.ts`?

ebroder: ...yes and I'm sure the tests will momentarily remind me of that. 🤦🏻

zarvox (approved):

PR #964, opened 2022-07-31
Lint for and address API deprecations

ebroder:
This introduces a new linting library which detects library methods that
have been marked as deprecated. With that in place, it also corrects
outstanding deprecations. Those break down into the following:

* `Mongo.Collection.prototype._ensureIndex` has been deprecated for some
  time, replaced by `Mongo.Collection.prototype.createIndex`, which is
  documented to have an identical API.
* `window.orientation` has been deprecated in favor of
  `window.screen.orientation`.
* `KeyboardEvent.keyCode` has been deprecated in favor of the
  string-typed and more cross-compatible `KeyboardEvent.key`
* `io-ts` has deprecated a number of types:
  - `t.taggedUnion` has been deprecated in favor of `t.union` (with no
    explicit declaration of which field contains the tag). I believe
    this was driven by improvements in TypeScript's inference.
  - `t.object` has been deprecated in favor of `t.UnknownRecord`, the
    equivalent of `Record<string, unknown>`. They are mostly equivalent
    in function, although the Discord API's `toJSON` method is
    sufficiently underspecified that a cast was necessary.
  - `t.Integer` has been deprecated in favor of `t.Int`. `t.Int` is a
    "branded type", which is intended to prevent accidentally assigning
    between two different types that would otherwise look identical to
    the type checker. (One example use case would be preventing "puzzle
    ID" from being assigned to a "hunt ID" field, even though both are
    strings). `io-ts` expects you to handle this by using its "encode"
    and "decode" mechanisms for runtime type validation. We instead
    convert the codecs to SimpleSchema schemas and do our runtime
    validation there. But this means we need to dance around the fact
    that we don't actually have branded types anywhere.


ebroder:
FWIW I'm somewhat on the fence about whether having a linter for deprecation is actually a good idea or not. It seems nice for code cleanliness and making sure we're keeping up with the Jones's, but it also will (e.g.) impede our ability to take updates in ways which may or may not be desirable. Very open to just chucking this PR (or chucking the linter and leaving the code fixes) if it doesn't seem worthwhile.

.eslintrc.json
```diff
@@ -126,6 +127,9 @@
     "@typescript-eslint/no-namespace": "off",
     "@typescript-eslint/prefer-namespace-keyword": ["off"],
 
+    // Show errors for deprecated code
+    "deprecation/deprecation": "error",
```
zarvox, line 131: Another option would be to leave these as warnings instead of errors, which would allow us to still take updates that deprecate things we use, while giving us the opportunity to learn about them locally with some tooling support.

zarvox (approved):

.eslintrc.json
ebroder: Ehhhh I kind of feel like warnings just lead to broken windows

ebroder: Let's try it and see if it becomes annoying. I'm open to reconsidering down the road.

PR #965, opened 2022-07-31
Add CSS linting

### LINKED ISSUE #637: Add linting for styled-components CSS
It looks like a [small error](https://github.com/deathandmayhem/jolly-roger/commit/be9386a38b60a5668cb955a1df364918b050599d#diff-a45372330478e84b301320fb902ea627dedc93f98bbea0413d9318a236215d39R12) snuck into our CSS when I converted it to styled-components (a CSS Object Model property instead of the actual CSS). @zarvox already picked up this specific error and fixed it as part of #636, but it's unfortunate that we don't have any generalized linting for errors of this form.

It seems like the `styled-components.vscode-styled-components` extension for VSCode will highlight the error through a plugin to the Typescript compiler ([typescript-styled-plugin](https://github.com/microsoft/typescript-styled-plugin)), but I can't manage to spit out errors when I invoke `tsc` directly.

styled-components [recommends using stylelint](https://styled-components.com/docs/tooling#stylelint), which is probably a better approach, but it seems like the relatively recent stylelint v14 significantly changed the interfaces for CSS syntaxes embedded in other languages. [styled-components/styled-components#3607](https://togithub.com/styled-components/styled-components/issues/3607) has some guidance for how to work around this.

As a note: I ran into issues running stylelint over `.ts` files, even though it seems like that should be legal; running it over `.tsx` files seems to work fine. We will need to do some shenanigans for `mediaBreakpointDown`. It seems like we should be able to use something like the [`sc-selector`](https://styled-components.com/docs/tooling#interpolation-tagging) magic comment, but in practice I was only able to get stylelint to accept it if I used [`sc custom "@media(min-width: 1px)"`](https://styled-components.com/docs/tooling#sccustom) or thereabouts.

I pushed a branch ([`evan/stylelint`](https://github.com/deathandmayhem/jolly-roger/tree/evan/stylelint)) which adds stylelint configuration, but I'll hold off on proposing it for merge until @zarvox is done converting the puzzle page to use styled-components. (It did find a handful of errors that mostly seem nitpicky, but I'm generally in favor of stricter style standards over laxer)

ebroder:
This configures stylelint-based linting for styled-components within tsx files. It is...not perfect - modern versions of stylelint and styled-components do not seem to play entirely nice with each other (I wrote up some of the challenges I ran into in #637, but I also completely gave up on linting the handful of raw SCSS files we still have). But I'm generally a strong believer in linting and forced style concordance, and this did catch one arguably real bug (namely that `StyledNotificationActionBar` had two different `display` declarations)

Fixes #637

### DIFF
--- a/imports/client/components/Celebration.tsx
+++ b/imports/client/components/Celebration.tsx
@@ -13,8 +13,7 @@ const CelebrationOverlay = styled.div`
   flex-direction: column;
   align-items: center;
   justify-content: center;
-
-  background-color: rgba(0, 0, 0, .20);
+  background-color: rgb(0 0 0 / 20%);
   z-index: 1050;
 `;
 
--- a/imports/client/components/DocumentDisplay.tsx
+++ b/imports/client/components/DocumentDisplay.tsx
@@ -19,10 +19,10 @@ const StyledDeepLink = styled(DeepLink)`
 `;
 
 const StyledIframe = styled.iframe`
-  /* Workaround for unusual sizing behavior of iframes in iOS Safari */
-  /* Width and height need to be specified in px then adjusted by min and max */
-  width: 0px;
-  height: 0px;
+  /* Workaround for unusual sizing behavior of iframes in iOS Safari:
+   * Width and height need to be specified in absolute values then adjusted by min and max */
+  width: 0;
+  height: 0;
   min-width: 100%;
   max-width: 100%;
   min-height: 100%;
@@ -33,7 +33,7 @@ const StyledIframe = styled.iframe`
   bottom: 0;
   left: 0;
   border: 0;
-  padding-bottom: env(safe-area-inset-bottom, 0px);
+  padding-bottom: env(safe-area-inset-bottom, 0);
   background-color: #f1f3f4;
 `;
 
@@ -41,7 +41,7 @@ export const DocumentMessage = styled.span`
   display: block;
   width: 100%;
   height: 100%;
-  background-color: #ddddff;
+  background-color: #ddf;
 `;
 
 const GoogleDocumentDisplay = ({ document, displayMode }: DocumentDisplayProps) => {
--- a/imports/client/components/NotificationCenter.tsx
+++ b/imports/client/components/NotificationCenter.tsx
@@ -46,9 +46,10 @@ const StyledDismissButton = styled.button`
   height: 32px;
   font-size: 20px;
   font-weight: bold;
-  right: 0px;
-  top: 0px;
+  right: 0;
+  top: 0;
   color: #888;
+
   &:hover {
     color: #f0f0f0;
   }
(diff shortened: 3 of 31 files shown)


imports/client/components/Celebration.tsx
```diff
@@ -13,8 +13,7 @@ const CelebrationOverlay = styled.div`
   flex-direction: column;
   align-items: center;
   justify-content: center;
-
-  background-color: rgba(0, 0, 0, .20);
+  background-color: rgb(0 0 0 / 20%);
```
zarvox, line 16: TIL yet another new syntax for expressing colors

imports/client/components/DocumentDisplay.tsx
```diff
@@ -19,10 +19,10 @@ const StyledDeepLink = styled(DeepLink)`
 `;
 
 const StyledIframe = styled.iframe`
-  /* Workaround for unusual sizing behavior of iframes in iOS Safari */
-  /* Width and height need to be specified in px then adjusted by min and max */
-  width: 0px;
-  height: 0px;
+  /* Workaround for unusual sizing behavior of iframes in iOS Safari:
+   * Width and height need to be specified in px then adjusted by min and max */
+  width: 0;
```
zarvox, line 24: Given the comment about what we're trying to work around, it's not totally clear to me that switching from `px` to unitless is actually safe here -- can we verify that this still does what we intended?

imports/client/components/NotificationCenter.tsx
```diff
@@ -81,28 +82,26 @@ const StyledNotificationMessage = styled.li`
...
+  padding: 0;
   flex-direction: row;
 `;
 
 const StyledNotificationActionItem = styled.li`
-  margin-left: 0px;
-  margin-top: 8px;
-  margin-right: 8px;
-  margin-bottom: 4px;
+  margin: 8px 8px 4px 0;
```
zarvox, line 93: In all my years of webdev I have never remembered the order of these ;_; I can probably be taught but I admit I find this less readable than how it was

zarvox (approved):
I personally find some of the choices of the linter uglier/more difficult to read (forcing `0` unitless everywhere, even next to other values like `4px`, collapsing `#aaaaaa` into `#aaa` when other colors are left as-is) but I'm willing to adapt for consistency

Before merging, I would like for us to verify that the bit where we comment about specifying width in `px` is still okay for us to specify `0` unitless, since that's already working around some mobile safari oddness, and `0` may not be parsed/handled specifically as `0px`?

imports/client/components/DocumentDisplay.tsx
ebroder: Based on [the commit that introduced this formatting](https://github.com/deathandmayhem/jolly-roger/commit/9a1a35f034f90b02aa2661a2dcac3c84ec4dfc00) and [this StackOverflow response](https://stackoverflow.com/a/23083463), I believe the relevant thing here is moving from `height: 100%; width: 100%` to using `{min,max}-{width,height}`, and the comment should probably say that "width and height need to be specified as an absolute value instead of a percentage."

However, I did borrow @ccpost's iPad to see if I could reproduce this issue, and I was unable to. I tried the configuration on this branch, the configuration on master, and I also tried reverting the change in 9a1a35f034f90b02aa2661a2dcac3c84ec4dfc00. The rendering and scrolling behavior in all three cases seemed to be the same. I'm not sure if I'm misunderstanding what the original issue was or if Apple has since changed WebKit's behavior.

I think that means we could probably revert that change - it would have the minor advantage of simplifying the styling some - but I'm not going to do that for now without a better understanding off the issue. However, I do think it's safe to leave the dimensions unitless, and I'll update the comment to be more clear.

imports/client/components/NotificationCenter.tsx
ebroder: It's clockwise, if that helps

PR #966, opened 2022-08-01
Remove celebrations

ebroder:
We've had this disabled ever since we added support for multiple answers
and haven't really missed it - especially since puzzle hunt websites
seem to mostly have their own these days.


zarvox:
This all looks good, but I think I'd also like to see a cleanup migration which:

* removes the `disable.applause` feature flag row from the DB, if present
* removes the `muteApplause` field from any user objects that already had them migrated by an earlier execution of migration 38/39, or had been set after migration 39

ebroder:
:+1: PTAL?

zarvox (approved):
Thanks!

Issue #603, opened 2022-01-16
Audio feed from Safari can be muffled by broad noise


grantaelliott:
Users on Safari are occasionally rendered unintelligible. This is resolved by leaving the call and rejoining.

This seems to happen only after the JR page loses focus, but my understanding of the necessary or sufficient conditions is inadequate.

ebroder:
I'm seeing some indication around the internet that `createMediaStreamSource` in Safari may be misbehaving, and we should use `stream.clone()` instead. I'm not entirely sure that will work for us (I think there are cases where we need an `AudioNode` and not a `MediaStream`?) but we should experiment.

grantaelliott:
No longer able to reproduce this. Whether by #967 or by a change in Safari, I believe this is fixed.

PR #967, opened 2022-08-01
Drop the gain node for mute/unmute

ebroder:
> Given mediasoup-client's native support for pausing/unpausing the producer, I think this is superfluous (and additionally, may be triggering bad behavior from Safari).
> 
> This addressed #603 in my testing.

FWIW, ^ was what I wrote in January when I pulled this commit together. I'm not sure why I didn't submit it for merging at the time, and I can no longer remember what I did to reproduce the issue "in my testing," or if I was even able to.

In any case, I think this change is worst-case harmless, as mediasoup-client _will_ stop sending data when a producer is paused (if you want to see how it threads through, [the `Producer` emits an event](https://github.com/versatica/mediasoup-client/blob/v3/src/Producer.ts#L329-L340) which [is caught by the transport](https://github.com/versatica/mediasoup-client/blob/v3/src/Transport.ts#L1134-L1141) which [replaces the track in the `RTCRtpSender`](https://github.com/versatica/mediasoup-client/blob/d510a4d3e8a76a67f4b6ef1a1287c1a308ee3c38/src/handlers/Chrome74.ts#L504-L509). That method [by spec accepts `null` and stops transmission](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/replaceTrack))

My intuition is to land this as it might help and even if it doesn't help cleans up the code a bit.

### DIFF
--- a/imports/client/components/CallSection.tsx
+++ b/imports/client/components/CallSection.tsx
@@ -102,11 +102,11 @@ const JoiningCall = ({ details }: { details?: string }) => {
 type ProducerCallback = ({ id }: { id: string }) => void;
 
 const ProducerManager = ({
-  muted,
+  paused,
   track,
   transport,
 }: {
-  muted: boolean;
+  paused: boolean;
   track: MediaStreamTrack;
   transport: types.Transport;
 }) => {
@@ -126,14 +126,18 @@ const ProducerManager = ({
   const producerServerCallback = useRef<ProducerCallback>();
 
   useEffect(() => {
-    if (producer && muted !== producer.paused) {
-      producer[muted ? 'pause' : 'resume']();
+    if (producer && paused !== producer.paused) {
+      if (paused) {
+        producer.pause();
+      } else {
+        producer.resume();
+      }
       mediasoupSetProducerPaused.call({
         mediasoupProducerId: producer.id,
-        paused: muted,
+        paused,
       });
     }
-  }, [muted, producer]);
+  }, [paused, producer]);
 
   const onProduce = useCallback((
     { kind, rtpParameters, appData }: {
@@ -270,7 +274,7 @@ const ProducerBox = ({
         {tracks.map((track) => (
           <ProducerManager
             key={track.id}
-            muted={muted}
+            paused={muted || deafened}
             track={track}
             transport={transport}
           />
--- a/imports/client/components/ChatPeople.tsx
+++ b/imports/client/components/ChatPeople.tsx
@@ -107,9 +107,7 @@ interface AudioControls {
 
 interface ChatAudioState {
   audioContext: AudioContext | undefined;
-  rawMediaSource: MediaStream | undefined;
-  gainNode: GainNode | undefined;
-  leveledStreamSource: MediaStream | undefined;
+  mediaSource: MediaStream | undefined;
 }
 
 // Helper to tear down tracks of streams
@@ -164,9 +162,7 @@ const ChatPeople = ({
 
   const [audioState, setAudioState] = useState<ChatAudioState>({
     audioContext: undefined,
-    rawMediaSource: undefined,
-    gainNode: undefined,
-    leveledStreamSource: undefined,
+    mediaSource: undefined,
   });
 
   const subscriberTopic = `puzzle:${puzzleId}`;
@@ -281,13 +277,6 @@ const ChatPeople = ({
     };
   }, [loading, subscriberTopic, huntId, puzzleId]);
 
-  const updateGain = useCallback((muted: boolean) => {
-    const newGain = muted ? 0.0 : 1.0;
-    if (audioState.gainNode && audioState.audioContext) {
-      audioState.gainNode.gain.setValueAtTime(newGain, audioState.audioContext.currentTime);
-    }
-  }, [audioState]);
-
   const updatePeerState = useCallback((muted: boolean, deafened: boolean) => {
     const effectiveState = participantState(muted, deafened);
     if (selfPeer) {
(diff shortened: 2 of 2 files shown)


zarvox:
I appreciate you looking into this, and your previous investigations into what was making audio from Safari awful. 

> mediasoup-client will stop sending data when a producer is paused

Hmmm, I'm thinking through possible partial failure modes, and while I agree with your analysis, I'm still not totally convinced this meets my privacy bar by itself.  I think pausing the track will still require a successful server roundtrip/processing of the Meteor method before the client would stop sending audio.  If somehow the Meteor thread gets tied up on some hypothetical CPU-heavy task while the mediasoup subprocesses are still active forwarding data on other cores, that means we're still transmitting audio, even though the jolly-roger client UI will suggest that we're not.

Fortunately, some more reading suggests we could toggle https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/enabled on the client side as soon as `muted` (which represents our intended state) changes to silence the track without having to wait on a roundtrip.  The MDN docs call it out specifically as the way that mute/unmute should be implemented.  I'd be happy to merge this if we replaced the gain node with setting `enabled` on the track to `!muted`.

ebroder:
> I think pausing the track will still require a successful server roundtrip/processing of the Meteor method before the client would stop sending audio

I think the alternative you propose is reasonable and I'll change to that in a bit, but for what it's worth I don't think this assessment is true? Toggling mute [changes React state](https://github.com/deathandmayhem/jolly-roger/blob/main/imports/client/components/ChatPeople.tsx#L299), which is threaded as a prop through to `ProducerManager`, which has a `useEffect` that [immediately calls `pause` or `resume`](https://github.com/deathandmayhem/jolly-roger/blob/main/imports/client/components/CallSection.tsx#L129-L130) on the mediasoup producer before calling any Meteor methods. The server-side mechanics are mostly for communication (displaying mute state on other clients) and efficiency (not sending a dead audio stream); they're not (AIUI) required for the client to stop transmitting.

imports/client/components/CallSection.tsx
```diff
@@ -126,14 +126,14 @@ const ProducerManager = ({
   const producerServerCallback = useRef<ProducerCallback>();
 
   useEffect(() => {
-    if (producer && muted !== producer.paused) {
-      producer[muted ? 'pause' : 'resume']();
+    if (producer && paused !== producer.paused) {
+      producer[paused ? 'pause' : 'resume']();
```
zarvox, line 130: I seem to have overlooked this line entirely, which means my thesis was incorrect.  My apologies.

nit: can we make this a more standard function call structure?  The method-name-in-string-in-control-flow seems like it would be harder for tooling (or a human grepping about) to notice as a callsite of pause() or resume().

```
if (paused) {
  producer.pause();
} else {
  producer.resume();
}
```

imports/client/components/ChatPeople.tsx
```diff
@@ -324,19 +313,20 @@ const ChatPeople = ({
...
+        t.enabled = !nextState.muted;
+      });
 
       // Tell the server about our new state
       updatePeerState(nextState.muted, nextState.deafened);
 
       return nextState;
     });
-  }, [updateGain, updatePeerState]);
+  }, [updatePeerState]);
```
zarvox, line 329: Please fix warning about missing dependency (which I think should be fine to be just `audioState.mediaSource`, without the safe navigation).

imports/client/components/ChatPeople.tsx
```diff
@@ -302,20 +291,20 @@ const ChatPeople = ({
...
+        t.enabled = !nextState.muted;
+      });
 
       // Tell the server about our new state
       updatePeerState(nextState.muted, nextState.deafened);
 
       return nextState;
     });
-  }, [updateGain, updatePeerState]);
+  }, [updatePeerState, audioState]);
```
zarvox, line 307: maybe this should be `audioState.mediaSource`?

imports/client/components/ChatPeople.tsx
```diff
@@ -302,20 +291,20 @@ const ChatPeople = ({
...
-      const nextEffectiveMuteState = nextState.muted;
-
-      // Update gain if needed
-      if (prevEffectiveMuteState !== nextEffectiveMuteState) {
-        updateGain(nextEffectiveMuteState);
-      }
+      // Pausing or unpausing the consumer in mediasoup should be sufficient to
+      // stop transmitting audio, but just to be sure, we disable the capture
+      // stream here as well.
+      audioState?.mediaSource?.getTracks().forEach((t) => {
```
zarvox, line 297: shouldn't `audioState` always be truthy here?  we shouldn't need the `?.` unless I'm misreading something

imports/client/components/ChatPeople.tsx
```diff
@@ -324,19 +313,20 @@ const ChatPeople = ({
...
-      const prevEffectiveMuteState = prevState.muted || prevState.deafened;
-      const nextEffectiveMuteState = prevState.muted || nextState.deafened;
-      // Update gain if needed
-      if (prevEffectiveMuteState !== nextEffectiveMuteState) {
-        updateGain(nextEffectiveMuteState);
-      }
+      // Pausing or unpausing the consumer in mediasoup should be sufficient to
+      // stop transmitting audio, but just to be sure, we disable the capture
+      // stream here as well.
+      audioState?.mediaSource?.getTracks().forEach((t) => {
```
zarvox, line 319: I think `audioState` should always be truthy here, so the first `?.` can just be `.`

zarvox (commented):
Ah, you're right, sorry for the confusion.

A couple nits but otherwise this is good to go, assuming you did some cursory functionality testing (since we don't really have any automation keeping this working).

imports/client/components/CallSection.tsx
ebroder: The tooling is actually totally fine with it :) but yes you're right, it's confusing, I'll switch

imports/client/components/ChatPeople.tsx
ebroder: Yeah it is, I thought that one of the typescript-aware linters would have complained about that. /shrug

zarvox (approved):

PR #976, opened 2022-08-02
Better relative time helpers

ebroder:
Was browsing through the code and remembered that I wanted to pull this together.

Previously `relativeTimeFormat` returned a static string, and any component that used it would need to set its own timers to re-render as it changed. Instead, this introduces a `RelativeTime` component which automatically re-renders whenever the output of `relativeTimeFormat` would change.

In order to support this, `relativeTimeFormat` now has a "complete" variant which additionally returns how long until the output would change (which is fairly easy to calculate, as we already have the "remainder" lying around). And to facilitate that change, I unified the "terse" and non-terse versions into a single method so that the logic could be shared. (Slightly confusing in the context of this PR, "complete" here is different than how we previously used that term with `relativeTimeFormat` - previously it indicated that the function would return a full breakdown across all sub-units of time. But once this is merged that shouldn't be a source of confusion going forward).

Since the code manipulations felt a little fiddly, I intentionally broke this into pretty fine-grained commits, which is likely the easiest way to review this.

### DIFF
--- a/imports/client/components/RelativeTime.tsx
+++ b/imports/client/components/RelativeTime.tsx
@@ -1,0 +1,29 @@
+import { Meteor } from 'meteor/meteor';
+import React, { useEffect, useState } from 'react';
+import { complete, RelativeTimeFormatOpts } from '../../lib/relativeTimeFormat';
+
+const RelativeTime = ({
+  date, minimumUnit, maxElements, terse, now,
+}: {
+  date: Date,
+} & RelativeTimeFormatOpts) => {
+  const [formatted, setFormatted] = useState(complete(date, {
+    minimumUnit, maxElements, terse, now,
+  }));
+
+  useEffect(() => {
+    const timeout = Meteor.setTimeout(() => {
+      setFormatted(complete(date, {
+        minimumUnit, maxElements, terse, now,
+      }));
+    }, formatted.millisUntilChange);
+
+    return () => {
+      Meteor.clearTimeout(timeout);
+    };
+  }, [date, formatted.millisUntilChange, maxElements, minimumUnit, now, terse]);
+
+  return <span>{formatted.formatted}</span>;
+};
+
+export default RelativeTime;
--- a/imports/client/components/RelativeTime.tsx
+++ b/imports/client/components/RelativeTime.tsx
@@ -21,7 +21,18 @@ const RelativeTime = ({
     return () => {
       Meteor.clearTimeout(timeout);
     };
-  }, [date, formatted.millisUntilChange, maxElements, minimumUnit, now, terse]);
+  }, [
+    date,
+    formatted.millisUntilChange,
+    maxElements,
+    minimumUnit,
+    now,
+    terse,
+    // Note that we explicitly include formatted.formatted here so that we set a
+    // new timeout if the formatted string changes but (by change) the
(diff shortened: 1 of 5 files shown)


tests/unit/imports/lib/relativeTimeFormat.ts
```diff
@@ -1,46 +1,56 @@
 import { assert } from 'chai';
-import relativeTimeFormat from '../../../../imports/lib/relativeTimeFormat';
+import { complete, RelativeTimeFormatOpts } from '../../../../imports/lib/relativeTimeFormat';
+
+const verifyFormat = (date: Date, opts: RelativeTimeFormatOpts, expected: string) => {
+  const { formatted, millisUntilChange } = complete(date, opts);
+  assert.equal(formatted, expected);
+  assert.notEqual(millisUntilChange, 0, 'millisUntilChange should never be zero');
+  const justBefore = new Date(date.getTime() - millisUntilChange + 1);
```
zarvox, line 8: This feels weirdly structured to me -- shouldn't this be comparing the same `date` against `now + millisUntilChange - 1`, instead of shuffling the reference date around?  (I understand they're ~computationally equivalent, but reading through it was pretty confusing to me, since in usage the reference date is generally the one staying fixed and `now` is generally the thing changing.)

Ditto `justAfter` below -- it'd feel more obviously like we're testing the specific case we expect to encounter if it were `now + millisUntilChange` instead.

imports/client/components/RelativeTime.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const timeout = Meteor.setTimeout(() => {
+      setFormatted(complete(date, {
+        minimumUnit, maxElements, terse, now,
+      }));
+    }, formatted.millisUntilChange);
+
+    return () => {
+      Meteor.clearTimeout(timeout);
+    };
+  }, [date, formatted.millisUntilChange, maxElements, minimumUnit, now, terse]);
```
zarvox, line 24: I think this might subtly miss updates after the first two, since `millisUntilChange` could plausibly be the same from one update-on-the-minute to the next, and then we'd lose all updates.

I think the most straightforward way to fix that is to have this deps list also include `formatted.formatted`.

tests/unit/imports/lib/relativeTimeFormat.ts
ebroder: Fair. I was trying to avoid having to dig into `opts` but you're right. I'll switch

imports/client/components/RelativeTime.tsx
ebroder: Good catch, fixed.

imports/client/components/RelativeTime.tsx
```diff
@@ -21,7 +21,18 @@ const RelativeTime = ({
...
+    date,
+    formatted.millisUntilChange,
+    maxElements,
+    minimumUnit,
+    now,
+    terse,
+    // Note that we explicitly include formatted.formatted here so that we set a
+    // new timeout if the formatted string changes but (by change) the
+    // millisUntilChange does not.
+    formatted.formatted,
```
zarvox, line 34: Ideally this would have been fixup'd into the previous commit in the series, but whatever

zarvox (approved):

PR #982, opened 2022-08-04
Use ValidateShape for TypedMethod calls

ebroder:
Since we generally require exact object matching when validating
TypedMethod arguments, we don't want to accidentally accept extra fields
in the argument type. By using ValidateShape, we can catch this at the
type level.

### DIFF
--- a/imports/lib/ValidateShape.ts
+++ b/imports/lib/ValidateShape.ts
@@ -1,0 +1,4 @@
+type ValidateShape<T, Shape> =
+  Shape & { [K in keyof T]: K extends keyof Shape ? T[K] : never };
+
+export default ValidateShape;
--- a/imports/lib/models/Base.ts
+++ b/imports/lib/models/Base.ts
@@ -1,6 +1,7 @@
 import { check, Match } from 'meteor/check';
 import { Meteor, Subscription } from 'meteor/meteor';
 import { Mongo } from 'meteor/mongo';
+import ValidateShape from '../ValidateShape';
 import { userIdIsAdmin } from '../is-admin';
 import { BaseType } from '../schemas/Base';
 
@@ -17,11 +18,6 @@ export type FindOneOptions = {
 export type FindOptions = FindOneOptions & {
   limit?: number;
 }
-
-type ValidateShape<T, Shape> =
-    T extends Partial<Shape> ?
-    Exclude<keyof T, keyof Shape> extends never ?
-    T : never : never;
 
 class Base<T extends BaseType> extends Mongo.Collection<T> {
   public name: string;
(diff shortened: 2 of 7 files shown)


imports/lib/ValidateShape.ts
```diff
@@ -0,0 +1,6 @@
+type ValidateShape<T, Shape> =
+    T extends Shape ?
```
ebroder, line 2: Quick note that this was previously `T extends Partial<Shape>`. I'm not sure why given that I was already passing in `doc: ValidateShape<U, Partial<T>>` in `Base`'s declaration of `insert`, but it's clearly not what we want for method declarations (partial types are not valid)

imports/lib/models/Base.ts
```diff
@@ -53,6 +49,14 @@ class Base<T extends BaseType> extends Mongo.Collection<T> {
...
   }
 
+  // @ts-expect-error TS is correct that Partial<T> is not equivalent to
+  //   OptionalId<T>, but that's because OptionalId<T> isn't exactly correct.
+  //   Our SimpleSchemas will fill in some of the missing fields using
+  //   autoValues, but it's not possible to have a collection type that
+  //   distinguishes between fields that are present/required on read vs. on
+  //   write. The resulting declaration is looser than it should be, but it only
+  //   pushes validation from type checking to runtime (as SimpleSchema will
+  //   verify that required fields are actually present).
```
ebroder, line 59: OK. I think I understand the history here.

AIUI, the original override of `insert` to accept `Partial<T>` (#247) _did_ require `@ts-ignore`, as TypeScript couldn't tell that this is a valid subclassing of the `Mongo.Collection` parent class - i.e. it couldn't tell if `Partial<T>` was contravariant with `OptionalId<T>`, which is what the superclass takes. Arguably it's not, but that's mostly because we know better than the upstream type declarations that a `Partial<T>` is actually quasi-acceptable here.

When we originally introduced `ValidateShape` in #366, I believe that the `@ts-ignore` was made unnecessary. I'm not sure if it was actually unnecessary, or if TS was just unable to tell that anything untoward is going on. (It was probably confused by the ternaries). In either case, we didn't notice that it was unnecessary so we left it in.

Subsequently, when I turned on new linting configuration in #893, that config banned use of `@ts-ignore`. At the time, I was surprised to find that the declaration wasn't necessary, but it was easy enough to drop.

Now, with the new, simpler definition of `ValidateShape`, TS seems to have noticed what's going on again and is displeased. However, the new linting configuration is tolerant of `@ts-expect-error`, which has approximately the same effect as `@ts-ignore` but will error if there _is_ no type checking error. This should prevent us from missing if some future change again makes it unnecessary.

imports/lib/ValidateShape.ts
```diff
@@ -0,0 +1,4 @@
+type ValidateShape<T, Shape> =
+  Shape & { [K in keyof T]: K extends keyof Shape ? T[K] : never };
```
ebroder, line 2: This is much simpler than the old declaration of `ValidateShape`, but I think should have the same effect (and I believe that the new test confirms that)

ebroder:
OK needed to play with it a bit more, but I think this is _actually_ ready for review now. (And comes with tests!)

zarvox (approved):
Brilliant.  I love that we got the immediate validation that this would have detected the failure mode we just fixed by also flagging the other possible field that could have caused trouble.  Nice that that was the only thing it seems to have complained about. 

Also appreciate the unit tests!

PR #984, opened 2022-08-07
Update to Bootstrap 5/react-bootstrap 2

### LINKED ISSUE #983: Upgrade to bootstrap 5/react-bootstrap 2
react-bootstrap's upgrade guide can be found at https://react-bootstrap.github.io/migrating/

zarvox:
Upgrades our codebase to use bootstrap 5 and the latest corresponding release of react-bootstrap.

Most of the necessary changes are called out in https://react-bootstrap.github.io/migrating/ or https://getbootstrap.com/docs/5.0/migration/.  This PR is best read commit-by-commit with those migration guides as a reference on the side.

I tried to avoid having this PR make any significant visual or behavioral changes at this time, though there were a few things where bootstrap upstream either changed default behaviors or simply removed components or behaviors we were using.  Notably:

* bootstrap upstream made links default to having underline shown at all times instead of only on hover.  This is probably helpful for mobile users, but makes our rather-link-heavy UI look a lot different.  I overrode the link decoration behavior to match our existing link behavior (no decoration by default, underline on hover).
* bootstrap removed the Jumbotron component, which we use on the splash/login/create first user/password reset pages.  I replaced it with a similarly-styled `<div>`.
* bootstrap significantly reworked how `Accordion`s are implemented, replacing the previous `<Card>` design entirely.  I adjusted the `/rtcdebug` page to make appropriate use of the new components.
* bootstrap removed the default margin from form groups.  You are instead instructed to use the margin classes like `mb-3` to provide the desired amount of margin.  I added this class to each form group I noticed, since wrapping components seemed like it would be both more component-tree overhead and would have less information scent if we added abstraction here.

Fixes #983.


imports/client/components/App.tsx
```diff
@@ -223,7 +223,7 @@ const AppNavbar = () => {
             {' '}
             <NavUsername>{displayName}</NavUsername>
           </DropdownToggle>
-          <DropdownMenu alignRight>
+          <DropdownMenu align="end">
```
ebroder, line 226: Good to know that we're ready for RTL Jolly Roger

imports/client/components/HuntListPage.tsx
```diff
@@ -65,9 +66,11 @@ const Hunt = React.memo(({ hunt }: { hunt: HuntType }) => {
...
       <ButtonGroup size="sm">
         {canUpdate ? (
-          <Button as={Link} to={`/hunts/${huntId}/edit`} variant="outline-secondary" title="Edit hunt...">
-            <FontAwesomeIcon fixedWidth icon={faEdit} />
-          </Button>
+          <LinkContainer to={`/hunts/${huntId}/edit`}>
+            <Button as="a" variant="outline-secondary" title="Edit hunt...">
+              <FontAwesomeIcon fixedWidth icon={faEdit} />
+            </Button>
+          </LinkContainer>
```
ebroder, line 73: I definitely remember the interaction here being messy when I was doing the react-router v5 upgrade, but I don't remember anything beyond that it was messy, so I guess if this works then :+1:

ebroder (approved):
Thanks for slogging through this!

imports/client/components/HuntListPage.tsx
zarvox: Yeah it seems to work and doesn't upset the typechecker, so I figured this was good enough. :)

I do think that the types are wrong because `as={Link}` works just fine at runtime, but I also didn't want to block doing this upgrade on those types being fully correct.

PR #985, opened 2022-08-07
Fix an incorrect missing dep on PuzzlePage

zarvox:
Before this change, it was possible that if you hit `F5` to reload
the page while you had a puzzle page open, you might hit a race
condition where:

* `activePuzzle` was still `undefined` despite the various subscriptions
  finishing loading, because
* `activePuzzle` did not declare a dep on `puzzleDataLoading` and the
  cached value (computed *before* the sub data had arrived) was
  initially undefined, and
* `useTracker` had not yet scheduled a rerun of that particular
  invalidated computation, because
* the subscription completion that had triggered the rerender of
  PuzzlePage was not the puzzle data subscription, but a different sub

Once hit, this would cause an exception in rendering where we'd attempt
to access `activePuzzle.title`, which was made possible in part by a
faulty non-null type assertion.

This change should prevent us from dereferencing the field incorrectly,
and also ensure that `activePuzzle` gets updated at the earliest correct
time.

### DIFF
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -1277,18 +1277,21 @@ const PuzzlePage = React.memo(() => {
       Documents.findOne({ puzzle: puzzleId }, { sort: { createdAt: 1 } })
   ), [puzzleDataLoading, puzzleId]);
 
-  const activePuzzle = useTracker(() => Puzzles.findOneAllowingDeleted(puzzleId), [puzzleId]);
-  const breadcrumbTitle =
+  const activePuzzle = useTracker(() => (
     puzzleDataLoading ?
-      'loading...' :
-      `${activePuzzle!.title}${activePuzzle!.deleted ? ' (deleted)' : ''}`;
+      undefined :
+      Puzzles.findOneAllowingDeleted(puzzleId)
+  ), [puzzleDataLoading, puzzleId]);
+
+  const puzzleTitle = activePuzzle ? `${activePuzzle.title}${activePuzzle.deleted ? ' (deleted)' : ''}` : '(no such puzzle)';
+  const title = puzzleDataLoading ? 'loading...' : puzzleTitle;
   useBreadcrumb({
-    title: breadcrumbTitle,
+    title,
     path: `/hunts/${huntId}/puzzles/${puzzleId}`,
   });
 
-  const title = `${activePuzzle ? activePuzzle.title : 'loading puzzle title...'} :: Jolly Roger`;
-  useDocumentTitle(title);
+  const documentTitle = `${title} :: Jolly Roger`;
+  useDocumentTitle(documentTitle);
 
   const onResize = useCallback(() => {
     setIsDesktop(window.innerWidth >= MinimumDesktopWidth);
@@ -1339,9 +1342,12 @@ const PuzzlePage = React.memo(() => {
   if (puzzleDataLoading) {
     return <FixedLayout className="puzzle-page" ref={puzzlePageDivRef}><span>loading...</span></FixedLayout>;
   }
+  if (!activePuzzle) {
+    return <FixedLayout className="puzzle-page" ref={puzzlePageDivRef}><span>No puzzle found :( Did you get that URL right?</span></FixedLayout>;
+  }
   const metadata = (
     <PuzzlePageMetadata
-      puzzle={activePuzzle!}
+      puzzle={activePuzzle}
       document={document}
       displayNames={displayNames}
       isDesktop={isDesktop}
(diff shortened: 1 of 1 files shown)


ebroder:
I'm fine applying this change - it seems harmless - but I'm not sure I believe the sequence of events you describe, which makes me suspicious that this will actually fix anything.

In particular, I'm not sure this agrees with my understanding of `useTracker`:

> * `useTracker` had not yet scheduled a rerun of that particular
>   invalidated computation, because
> * the subscription completion that had triggered the rerender of
>   PuzzlePage was not the puzzle data subscription, but a different sub

AIUI, `useTracker` will force a rerender by using a hook that forces a re-render (https://github.com/meteor/react-packages/blob/master/packages/react-meteor-data/useTracker.ts#L106-L107) when the `Tracker.autorun` block fires (i.e. when Meteor detects that the dependencies have changed), which I think threads causality in the opposite direction that you're implying. I also didn't think that Meteor tied reactivity of a query to specific subscriptions, but rather that when subscriptions updated the local `minimongo` collections, it would check open reactive queries to see if that changed the resultset (again, threading causality in the opposite direction).

Actually, now that I look at the code, I wonder if [these `||` operators](https://github.com/deathandmayhem/jolly-roger/blob/d5b1d8dcc6f16ef7cab3d288e6879f85c65d732a/imports/client/components/PuzzlePage.tsx#L1257-L1268) are problematic. IIRC from when I switched to using `useSubscribe`, I believe that you need to actually call the returned `isLoading` function in order to set up reactivity on the loading state changing (see also [the code](https://github.com/meteor/react-packages/blob/master/packages/react-meteor-data/useSubscribe.ts#L17) and [the docs](https://github.com/meteor/react-packages/tree/master/packages/react-meteor-data#usesubscribesubname-args-a-convenient-wrapper-for-subscriptions)), but do you run into problems if `||` short-circuits?

zarvox:
I can totally believe that I got the precise sequence of events wrong here, but just to be clear, hitting the error is not a hypothetical:

Screen Shot 2022-08-08 at 4 38 55 PM

`useTracker` will issue a rerun when the `Tracker.autorun` value has/might have changed, but (like ~all other hooks, but most notably `useMemo`) it will also *elide* rerunning the inner function if all members of the `deps` array have not changed.  Since `[puzzleId]` never changes, I believe that means that the return value of the specific `useTracker` call will only update once the `Tracker.autorun` function triggers a rerun.

Empirically, at HEAD, it is possible for `activePuzzle` to be `undefined` while `puzzleDataLoading` is `false`.  My best guess for how that happens is that maybe the rerun of the function passed to `useTracker` gets scheduled *after* all the subs are marked as complete, but that a React rerender is triggered first, somehow?

I think short-circuiting is not a problem.  If any one sub is still loading, we'll still have called that function and should get rerendered when the result of that `*Loading()` function changes.  At worst, it'd cause us to avoid rerendering immediately on the completion of some particular subs (the ones that got short-circuited away), but as soon as the final still-true one becomes ready, we'd expect to get rerun.  So the short-circuiting behavior seems like it couldn't cause incorrect results -- just slightly less-responsive ones (and in this case, we don't see any benefit from rerendering before the rest of the data has loaded anyway, because [we guard rendering anything on `puzzleDataLoading` anyway](https://github.com/deathandmayhem/jolly-roger/blob/d5b1d8dcc6f16ef7cab3d288e6879f85c65d732a/imports/client/components/PuzzlePage.tsx#L1341-L1343)).

I think the thing that went off the rails was:

* we assumed that if `puzzleDataLoading` was false, then `activePuzzle` would be truthy, and it turns out for whatever reason, it wasn't.  I think adding the additional dep to the `useTracker` call computing `activePuzzle` will make this assumption valid, at least if the puzzle ID is a real puzzle.
* we probably thought the check at https://github.com/deathandmayhem/jolly-roger/blob/d5b1d8dcc6f16ef7cab3d288e6879f85c65d732a/imports/client/components/PuzzlePage.tsx#L1341 was sufficient to avoid doing things with wrong data, but it turns out that https://github.com/deathandmayhem/jolly-roger/blob/d5b1d8dcc6f16ef7cab3d288e6879f85c65d732a/imports/client/components/PuzzlePage.tsx#L1286 is used unconditionally since it's part of an argument to a hook.

Probably only the top hunk of the diff is required here, but the bottom hunk seems reasonable anyway.

imports/client/components/PuzzlePage.tsx
```diff
@@ -1279,11 +1279,15 @@ const PuzzlePage = React.memo(() => {
...
+  const activePuzzle = useTracker(() => (
+    puzzleDataLoading ?
+      undefined :
+      Puzzles.findOneAllowingDeleted(puzzleId)
+  ), [puzzleDataLoading, puzzleId]);
   const breadcrumbTitle =
     puzzleDataLoading ?
       'loading...' :
-      `${activePuzzle!.title}${activePuzzle!.deleted ? ' (deleted)' : ''}`;
+      `${activePuzzle?.title || '(no title)'}${activePuzzle?.deleted ? ' (deleted)' : ''}`;
```
ebroder, line 1290: OK, now that I understand what's going on, I think that I would be slightly happier if this were instead

```
  const breadcrumbTitle =
    activePuzzle ?
      'loading...' :
      `${activePuzzle.title}${activePuzzle.deleted ? ' (deleted)' : ''}`;
```

and if additionally we got rid of the `!` operator on line 1344 or so (passed to `PuzzlePageMetadata`), probably by making the conditional before it into `puzzleDataLoading || !activePuzzle`.

We should ultimately have better handling for 404-like errors than we do now, but we don't need to shave that yak here.

zarvox: I decided to do at least a modicum of shaving that yak since I was unhappy with having an incorrect loading error, and I felt it would obscure the actual logic/constraints.

zarvox:
Updated so that if the user typos the `puzzleId` they get this:

Screen Shot 2022-08-10 at 3 53 51 PM

and then some conditionals get cleaner and control flow proves truthiness at various usage sites, so there's less safe-navigation/assertion sigil soup.

ebroder (approved):

PR #993, opened 2022-08-11
Install and configure zodern:types

ebroder:
This package is designed to allow for Meteor packages to ship their own types. At this point, I don't think there are any packages that use it, so it's effectively a noop change, but in theory it might see adoption in the future, which would allow us to reduce our dependence on adding our own ambient type declarations for Meteor libraries.

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -38,3 +38,4 @@ deanius:promise
 hot-module-replacement@0.5.1
 zodern:standard-minifier-js
 fetch@0.1.1
+zodern:types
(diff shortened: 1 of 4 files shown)


ebroder:
(No packages use it today, but I may go and try to chase down/patch some of the ones we use to see if I can get them to incorporate this)

zarvox (approved):
No objections to this, but on principle I feel more enthusiastic about replacing most uses of Meteor packages (especially the mostly unmaintained ones) with either NPM modules or (for the really simple ones) just library code we carry in our tree.  But if you're able to get upstreams to do releases, more power to you.

ebroder:
Yeah I mostly agree, and we don't have that many left. The packages we still have are generally (a) pretty tied into the Meteor build system/runtime in ways that seem hard to extract (b) target Meteor's mongo library specifically. Even once async mongo lands, the latter is going to be hard to untangle since Meteor will still have a wrapper around the underlying mongo library with a different API (i.e. `Async` suffixing method names and such).

PR #1023, opened 2022-08-31
Reimplement WebRTC controls as a hook

### LINKED ISSUE #594: Switching between desktop and mobile layout causes you to leave chat
This is of course because the CallSection component gets unmounted/remounted without carrying state across, but we should see if we can do better.

zarvox:
This replaces the hierarchy of components that managed our WebRTC
interactions with a new `useCallState` hook which encapsulates all of
the call state management functionality.  This provides a key benefit:
we can hoist the actual state storage higher in the component tree,
which means that we won't lose call state when switching layouts between
desktop and mobile screen widths.

As a bonus, this change also corrects a bunch of things from the audio
calls implementation that would have gone awry under `StrictMode` in
React 18 due to doing side-effectful actions outside of `useEffect`
blocks.

Fixes #594.

### DIFF
--- a/imports/client/components/AudioConfig.tsx
+++ b/imports/client/components/AudioConfig.tsx
@@ -262,7 +262,6 @@ const AudioConfig = () => {
       <audio
         ref={audioRef}
         autoPlay
-        playsInline
         muted={!loopback}
       />
     </section>
--- a/imports/client/components/ChatPeople.tsx
+++ b/imports/client/components/ChatPeople.tsx
@@ -1,5 +1,4 @@
 import { Meteor } from 'meteor/meteor';
-import { Random } from 'meteor/random';
 import { useSubscribe, useTracker } from 'meteor/react-meteor-data';
 import { faCaretDown } from '@fortawesome/free-solid-svg-icons/faCaretDown';
 import { faCaretRight } from '@fortawesome/free-solid-svg-icons/faCaretRight';
@@ -17,8 +16,7 @@ import CallHistories from '../../lib/models/mediasoup/CallHistories';
 import Peers from '../../lib/models/mediasoup/Peers';
 import relativeTimeFormat from '../../lib/relativeTimeFormat';
 import { DiscordAccountType } from '../../lib/schemas/DiscordAccount';
-import { PeerType } from '../../lib/schemas/mediasoup/Peer';
-import mediasoupSetPeerState from '../../methods/mediasoupSetPeerState';
+import { Action, CallJoinState, CallState } from '../hooks/useCallState';
 import useSubscribeAvatars from '../hooks/useSubscribeAvatars';
 import { Subscribers } from '../subscribers';
 import { trace } from '../tracing';
@@ -30,8 +28,6 @@ import {
   PeopleListDiv,
 } from './styling/PeopleComponents';
 import { PuzzlePagePadding } from './styling/constants';
-
-const tabId = Random.id();
 
 interface ViewerSubscriber {
   user: string;
@@ -83,43 +79,6 @@ const PeopleListHeader = styled(ChatterSubsectionHeader)`
   text-indent: -1rem;
 `;
 
-enum CallState {
-  CHAT_ONLY = 'chatonly',
-  REQUESTING_STREAM = 'requestingstream',
-  STREAM_ERROR = 'streamerror',
-  IN_CALL = 'call',
-}
-
-interface AudioControls {
-  // A note on mute and deafen: being deafened implies you are also not
-  // broadcasting audio to other parties, because that would allow for
-  // situations where you are being disruptive to others but don't know it.
-  // This state value for muted is "explicitly muted" rather than "implicitly
-  // muted by deafen".  You are effectively muted (and will appear muted to
-  // other) if you are muted or deafened.  The `muted` boolean field here will
-  // only track if you are explicitly muted, but in all props for all children,
-  // the muted property represents "effectively muted".  (We track them
-  // separately because if you mute before deafening, then undeafen should
-  // leave you muted, and we'd lose that bit otherwise.)
-  muted: boolean;
-  deafened: boolean;
-}
-
-interface ChatAudioState {
-  audioContext: AudioContext | undefined;
-  mediaSource: MediaStream | undefined;
-}
-
-// Helper to tear down tracks of streams
-function stopTracks(stream: MediaStream | null | undefined) {
-  if (stream) {
-    const tracks = stream.getTracks();
-    tracks.forEach((track) => {
-      track.stop();
-    });
-  }
-}
-
 const ChatterSection = styled.section`
   flex: 0;
   background-color: #ebd0e3;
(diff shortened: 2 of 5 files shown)


imports/client/components/PuzzlePage.tsx
```diff
@@ -65,6 +67,11 @@ import TagList from './TagList';
 import FixedLayout from './styling/FixedLayout';
 import { MonospaceFontFamily, SolvedPuzzleBackgroundColor } from './styling/constants';
 
+// Shows a state dump as an in-page overlay when enabled.
+const DEBUG_SHOW_CALL_STATE = true;
```
zarvox, line 71: whoops, I should switch that back to `false`

imports/client/components/ChatPeople.tsx
```diff
@@ -363,47 +250,29 @@ const ChatPeople = ({
...
 
-  useEffect(() => {
-    // When unmounting, stop any tracks that might be running
-    return () => {
-      trace('ChatPeople stop tracks on unmount');
-      // Stop any tracks that might be running.
-      stopTracks(audioState.mediaSource);
-    };
-  }, [audioState.mediaSource]);
+  // TODO: ensure we do proper cleanup of the sending track on unmount
```
zarvox, line 275: This is taken care of

zarvox: Specifically, by https://github.com/deathandmayhem/jolly-roger/pull/1023/commits/fb9d6c23d7f6a3a1c2870f7cad8703ad2f7a267a#diff-abb257e3f8aad9487024c922792f027ffc5ec9de52e60cadb72c6e439d6ef4a3R324-R333

imports/client/hooks/useCallState.ts
```diff
@@ -0,0 +1,789 @@
...
+    return () => {
+      if (state.audioState?.mediaSource) {
+        state.audioState?.mediaSource.getTracks().forEach((track) => {
+          track.stop();
+        });
+      }
+    };
+  }, [state.audioState?.mediaSource]);
+
+  const joinSubRef = useRef<Meteor.SubscriptionHandle | undefined>(undefined);
```
zarvox, line 335: The reason we handle this subscription (and the `mediasoup:transports` sub below) manually instead of using `useTracker` or `useSubscription` is because of an unfortunate interaction between:

* us using these publishes to cause server-side state changes
* us expecting `[huntId, puzzleId, tabId]` (or below, `peer._id`) to generally be unique per client, with last-writer-wins semantics on conflict
* React reserving the right to rerun `render()` multiple times, as exercised by `StrictMode`
* `useTracker` (and as a result `useSubscription`) running the inner function body during both the render phase and the effect phase, resulting in multiple subscriptions

The solution is to move our use of state-changing subscriptions to run exclusively in the effect phase, which is reasonably straightforward.

In hindsight, this probably deserves a comment.

imports/client/hooks/useCallState.ts
```diff
@@ -0,0 +1,789 @@
...
+            recvTransportConnectCallback.current = undefined;
+          }
+        },
+      });
+      return () => observer.stop();
+    }
+    return undefined;
+  }, [hasSelfPeer, selfPeer?._id, sendTransportConnectCallback, recvTransportConnectCallback]);
+
+  // const notReadyYet = !sendTransport || !recvTransport;
```
zarvox, line 448: clean up

imports/client/hooks/useCallState.ts
```diff
@@ -0,0 +1,789 @@
...
+              entry.producer = newProducer;
+            } else {
+              console.error(`no entry in producerMapRef for ${track.id}`);
+            }
+            setProducerGeneration((prevValue) => prevValue + 1);
+          })();
+        }
+
+        // Safe: we just set it above
+        const producerState = producerMapRef.current.get(track.id)!;
```
zarvox, line 571: It should be possible to phrase this without `!`

imports/client/hooks/useCallState.ts
```diff
@@ -0,0 +1,789 @@
...
+        });
+      }
+    }
+  }, [selfPeer, state.audioControls?.muted, state.audioControls?.deafened, state.audioControls]);
+
+  // Ensure we clean up producers when unmounting.
+  useEffect(() => {
+    const producerMap = producerMapRef.current;
+    return () => {
+      // TODO: dedupe this and the effect above
```
zarvox, line 653: actually do this

imports/client/components/AudioConfig.tsx
```diff
@@ -262,7 +262,6 @@ const AudioConfig = () => {
       <audio
         ref={audioRef}
         autoPlay
-        playsInline
```
zarvox, line 265: I slipped this (and the matching change in `CallSection`) into this PR since I saw it's what's causing #1029
to fail presubmit checks at https://github.com/deathandmayhem/jolly-roger/runs/8193203773?check_suite_focus=true and this change rewrites the line in `CallSection` anyway.

The backstory is kinda fun -- this is a vestige of the distant past from when I was bringing up the mesh WebRTC implementation with video because I didn't yet have anything like a spectrogram and I was sick of hearing feedback when the connection was successfully established.  Once I had the implementation working reliably, I switched the `video` element to an `audio` one, without realizing this property isn't valid for `audio`.

imports/client/hooks/useCallState.ts
```diff
@@ -0,0 +1,805 @@
...
+  }, [huntId, puzzleId, tabId]);
+
+  useEffect(() => {
+    // When mediaSource (the mic capture stream) changes, stop all the tracks.
+    return () => {
+      if (state.audioState?.mediaSource) {
+        state.audioState?.mediaSource.getTracks().forEach((track) => {
+          track.stop();
+        });
+      }
```
ebroder, line 352: This will fire on the new tracks, right? Do you want to capture a reference to the old `mediaSource` in the `useEffect` before the unwind callback?

imports/client/hooks/useCallState.ts
```diff
@@ -0,0 +1,805 @@
...
+      return undefined;
+    }
+  }, [state.audioState?.mediaSource]);
+
+  // map from track ID to state for that track, including subscription handle
+  const producerMapRef = useRef<Map<string, ProducerState>>(new Map());
+  // force side effects from onProduce
+  const [producerParamsGeneration, setProducerParamsGeneration] = useState<number>(0);
+  // force side-effects when producer is ready
+  const [producerGeneration, setProducerGeneration] = useState<number>(0);
```
ebroder, line 523: I think I understand it after reading the rest of the file, but I think I'd appreciate a comment here about what these generation numbers trigger

ebroder (approved):
Sorry for not getting to this sooner. I'm a little sad that we can't rely on React to manage more of the lifecycle for us, but given that they explicitly seem to not want to do that, this seems like it will be a better long-term structure.

I left a few comments. Feel free to address them (or not, if I misunderstood something). I'll approve now so you don't need to wait for another review from me.

imports/client/hooks/useCallState.ts
```diff
@@ -0,0 +1,805 @@
...
+  const [state, dispatch] = useReducer(reducer, INITIAL_STATE);
+
+  useEffect(() => {
+    // If huntId, puzzleId, or tabId change (but mostly puzzleId), reset
+    // call state.
+    return () => {
+      log('huntId/puzzleId/tabId changed, resetting call state');
+      dispatch({ type: 'reset' });
+    };
+  }, [huntId, puzzleId, tabId]);
```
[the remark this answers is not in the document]
zarvox: I think it's a good idea, but the obvious implementation doesn't appear to do anything because we need to hook into react-router instead, since it's not a browser-level navigation that any of our links trigger.

This was something that was included in react-router v5, but is absent in v6: https://reactrouter.com/en/main/upgrading/v5#prompt-is-not-currently-supported .  Given that this looks like it'll be a bit more work to sort out, I filed #1037 for future followup.

imports/client/hooks/useCallState.ts
zarvox: Ah, that's probably wise, thanks.  In practice, I think the way we've written the rest of our code means `mediaSource` never transitions from truthy to truthy, so we never hit this edge case, but it's good to be resilient.

imports/client/hooks/useCallState.ts
zarvox: Wrote up a bit more about how the producer flow works, including what fields these generation counters are a proxy for changing, and which effects read those fields (and thus, need to rerun when the generation counter changes).

PR #1040, opened 2022-09-10
Update to React 18 and enable StrictMode

zarvox:
React 18 introduces a new root API which replaces `ReactDOM.render` as discussed in https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis .  Switching to the new root API enables concurrent mode.

Concurrent mode introduces some potentially surprising observable behaviors, which `<StrictMode>` is designed to help developers catch.  Most notably, in concurrent mode, React may call a render function, `setState()`, or reducer dispatch function multiple times during a single update cycle.  Code which relies on a 1-to-1 mapping between `render()` and effect calls will likely break, as will code which performs side-effects in `render()` or in a reducer (including within a `setState` function body).  More about what StrictMode does can be found at https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-strict-mode which I recommend reading.

Fortunately, the hairiest of our unsafe reliance on `render()` producing side effects got cleaned up in the call state rewrite from #1023.  The only remaining instance of odd behavior I was able to observe with StrictMode was in breadcrumbs, which...has always been a bit of a problem child, regardless of the implementation. 
 Since the original design that tries to keep track of the topological order of the breadcrumbs in the render tree seems just utterly hopeless, I gave up on that approach and switched to a strategy of sorting breadcrumbs by url path length, since our URLs get longer as they get more detailed and our breadcrumbs follow the same hierarchy as our URLs.

There were a couple other things that enabling `<StrictMode>` warned about, which were pretty straightforward to resolve:

* invalid DOM hierarchies (particularly: `<button>` as a child of another `<button>` all over the accordions on RTCDebugPage)
* use of findDOMNode in react-bootstrap, when `<OverlayTrigger>`'s immediate child is not a DOM node

Recommended review order:

* read the two React docs links from above to understand the implications of `<StrictMode>` and concurrent rendering going forward
* read this PR commit-by-commit

### DIFF
--- a/imports/client/components/NotificationCenter.tsx
+++ b/imports/client/components/NotificationCenter.tsx
@@ -238,9 +238,11 @@ const GuessMessage = React.memo(({
         <StyledNotificationActionBar>
           <StyledNotificationActionItem>
             <OverlayTrigger placement="top" overlay={copyTooltip}>
-              <CopyToClipboard text={guess.guess}>
-                <button type="button" aria-label="Copy"><FontAwesomeIcon icon={faCopy} /></button>
-              </CopyToClipboard>
+              {({ ref, ...triggerHandler }) => (
+                <CopyToClipboard text={guess.guess} {...triggerHandler}>
+                  <button ref={ref} type="button" aria-label="Copy"><FontAwesomeIcon icon={faCopy} /></button>
+                </CopyToClipboard>
+              )}
             </OverlayTrigger>
           </StyledNotificationActionItem>
           <StyledNotificationActionItem>
--- a/imports/client/components/RTCDebugPage.tsx
+++ b/imports/client/components/RTCDebugPage.tsx
@@ -23,10 +23,9 @@ import { faVolumeUp } from '@fortawesome/free-solid-svg-icons/faVolumeUp';
 import { faWifi } from '@fortawesome/free-solid-svg-icons/faWifi';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import React, {
-  useContext, useEffect, useMemo, useState,
+  useCallback, useEffect, useMemo, useState,
 } from 'react';
 import Accordion from 'react-bootstrap/Accordion';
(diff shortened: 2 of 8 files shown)


zarvox:
It occurs to me now that I've opened this PR that our current `useDocumentTitle` implementation is probably also wrong under concurrent mode, given how `useImmediateEffect` seems to be hard to make work correctly with concurrent mode.

Maybe we should reflect on how we manage the document title anyway?  It's less-detailed than breadcrumbs for nearly everything but `PuzzlePage`.  I'm not sure if that's by design or an accident of history.  I also notice that we don't construct the whole breadcrumb path in reverse in the document title, just the top-level name and `:: Jolly Roger`.

One option could be to use a similar strategy to `useBreadcrumb` where each `useDocumentTitle` call also specifies a path, and we just flush the longest-pathed (and thus most-specific) title to `document.title` in the effect phase whenever the list is changed.

ebroder:
I'm not sure I had an especially coherent plan with document title. I do think it's reasonable that it's less detailed than breadcrumbs (tabs in modern browsers are generally pretty small so the title mostly gets truncated anyway). I think the two important use cases I can imagine are:

- Make it easy to distinguish between multiple open JR tabs (which may be an argument for a little more detail - e.g. including "firehose" / "announcements" / etc if you're on those pages)
- Maaaaybe having a useful default title if you bookmark a JR page (which, frankly, I doubt anyone does)

Do you want me to go ahead and review this now or wait until you look at the test failures? (I did a quick skim and things seem basically fine)

zarvox:
Thanks for the input!

> Do you want me to go ahead and review this now or wait until you look at the test failures? (I did a quick skim and things seem basically fine)

I'm fine either way.  I haven't had a chance to look deeply at the test failures yet.  I observe *different* failures locally, which seem to center on `resetDatabase` failing, whereas what I believe CI is hitting is more likely related to the `LocationCapture` route-capturing mechanism.

ebroder (approved):

PR #1058, opened 2022-09-23
Define our own Id, Email, and Url validation

zarvox:
For a couple of reasons, `simpl-schema` upstream has removed their copy of these regexes in the latest version, which we'd like to be able to upgrade to.  To unblock taking that upgrade, instead of relying on `SimpleSchema.RegEx`, define our own equivalent regexes (for Id and Email) and a custom validator for Url that just attempts to construct a URL object (since that regex was vulnerable to regex DoS attacks).

Unblocks #1050.

### DIFF
--- a/imports/lib/schemas/Announcement.ts
+++ b/imports/lib/schemas/Announcement.ts
@@ -1,6 +1,6 @@
 import * as t from 'io-ts';
-import SimpleSchema from 'simpl-schema';
 import { BaseCodec, BaseOverrides } from './Base';
+import { Id } from './regexes';
 import { Overrides, buildSchema, inheritSchema } from './typedSchemas';
 
 const AnnouncementFields = t.type({
@@ -10,7 +10,7 @@ const AnnouncementFields = t.type({
 
 const AnnouncementFieldsOverrides: Overrides<t.TypeOf<typeof AnnouncementFields>> = {
   hunt: {
-    regEx: SimpleSchema.RegEx.Id,
+    regEx: Id,
   },
 };
 
--- a/imports/lib/schemas/typedSchemas.ts
+++ b/imports/lib/schemas/typedSchemas.ts
@@ -1,7 +1,7 @@
 import { Meteor } from 'meteor/meteor';
 import * as t from 'io-ts';
 import { date } from 'io-ts-types';
-import SimpleSchema, { SimpleSchemaDefinition } from 'simpl-schema';
+import SimpleSchema, { SchemaDefinition, SimpleSchemaDefinition } from 'simpl-schema';
 import { uint8Array } from './types';
 
 type NumberOverrides<T> = T extends number ? {
@@ -70,6 +70,7 @@ type AutoValueReturn<T> = undefined | T | AutoValueFlatten<T> | (T extends any[]
 type SharedOverrides<T> = {
   defaultValue?: NonNullable<T>;
   autoValue?: (this: AutoValueThis<T>) => AutoValueReturn<T>;
+  custom?: SchemaDefinition['custom'];
 
   // These fields are from collection2
   index?: boolean | 1 | -1;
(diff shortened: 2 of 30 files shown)


imports/lib/schemas/Announcement.ts
```diff
@@ -1,6 +1,6 @@
 import * as t from 'io-ts';
-import SimpleSchema from 'simpl-schema';
 import { BaseCodec, BaseOverrides } from './Base';
+import { Id } from './regexes';
```
ebroder, line 3: OOC, is there any built-in value `Id` that we need to worry about shadowing by importing at top-level? I assume not if eslint was fine with it but just checking...

imports/lib/schemas/typedSchemas.ts
```diff
@@ -30,6 +30,7 @@ type StringOverrides<T> = T extends string ? {
   exclusiveMin?: boolean;
   exclusiveMax?: boolean;
   regEx?: RegExp | RegExp[];
+  custom?: SchemaDefinition['custom'];
```
ebroder, line 33: Does this only apply to string or should it be part of `SharedOverrides`?

imports/lib/schemas/validators.ts
```diff
@@ -0,0 +1,16 @@
+import { SchemaDefinition } from 'simpl-schema';
+
+const ValidUrl: SchemaDefinition['custom'] = function () {
+  // Tests if a URL is valid by attempting to construct a URL object from it.
+  if (!this.isSet) return undefined;
+  try {
+    // eslint-disable-next-line no-new
```
ebroder, line 7: Aside: I'm surprised there's not a `(void)new URL` override or something similar for this /shrug

ebroder (approved):
Questions are all non-blocking. Thanks for putting this together

imports/lib/schemas/Announcement.ts
zarvox: I checked the node and browser and Meteor shell environments and there weren't any globals or types named `Id` in any of them, so I think we're safe in that regard.

Happy to rename this or restructure this if you have another way to spell this in mind. :)

ebroder:
Nope no objections just curious

imports/lib/schemas/typedSchemas.ts
zarvox: Huh.  I think I don't fully understand how `FieldOverrides` works -- to me it reads like the sum type of a variety of other types, so if I had a particular concrete type T that extended string, then FieldOverrides<T> would be:

```
FieldOverrides<T> =
  Record<string, never> | // NumberOverrides<T> |
  Record<string, never> | // DateOverrides<T> |
  Record<string, never> | // ArrayOverrides<T> |
  StringOverrides<T> |
  Record<string, never> | // ObjectOverrides<T> |
  SharedOverrides<T>;
```

which means that the overrides for that field can include fields from `StringOverrides`, or things from `SharedOverrides`, but not both -- the type must fit into one of the two outlines.

In practice this works out; the only times we specify `custom` we specify no other overrides, so as written in this PR, the type is inferred as `StringOverrides<T>`.  If I moved `custom?` to `SharedOverrides<T>`, I imagine that the type of a field override would be inferred as matching `SharedOverrides<T>` instead. 

Maybe we actually meant for this to be more like:

```
export type FieldOverrides<T> = (
  NumberOverrides<T> |
  DateOverrides<T> |
  ArrayOverrides<T> |
  StringOverrides<T> |
  ObjectOverrides<T>) &
  SharedOverrides<T>;
```

instead, where we'd accept the additional SharedOverrides specifiers with any type, and (when the type matched) also allow keys from the more-specific `*Overrides` type?  That'd align more with my guess of what these types were trying to model, but also some other things fail to typecheck if we do that.

I can move `custom` to `SharedOverrides` though; it doesn't need to be bound to only be usable on strings.

imports/lib/schemas/validators.ts
zarvox: Looks like I can spell this

```
void (new URL(...))
```

PR #1127, opened 2022-11-04
Cleanup

ebroder:
Cleanup a Meteor package that we were only barely using and some type declarations that got merged upstream.


ebroder:
And snuck in one more quick change - a postshrinkwrap script that corrects SHA1 integrity hashes in `package-lock.json` back to SHA512.

tests/acceptance/lib.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable import/prefer-default-export */
 /* eslint-disable no-console */
+import { promisify } from 'util';
```
zarvox, line 3: I guess this comes from the meteor node stubs?  IIRC `util` is a strictly-nodejs thing.

tests/acceptance/profiles.tsx
```diff
@@ -15,6 +16,21 @@ const addUserToRole = new TypedMethod<{ userId: string, scope: string, role: str
 const createHunt = new TypedMethod<{ name: string }, string>('test.methods.profiles.createHunt');
 const joinHunt = new TypedMethod<{ huntId: string, userId: string }, void>('test.methods.profiles.joinHunt');
 
+const subscribeAsync = (name: string, ...args: any[]) => new Promise<Meteor.SubscriptionHandle>(
+  (resolve, reject) => {
+    const handle = Meteor.subscribe(name, ...args, {
+      onError: (reason?: Meteor.Error) => {
```
zarvox, line 22: Per https://docs.meteor.com/api/pubsub.html#Meteor-subscribe there is no `onError`, only `onStop` (which fires with no args on client-driven teardown and a Meteor.Error if the publish function throws or the server cancels the sub).

zarvox (commented):
Loving this -- super excited to be getting rid of the monkeypatches (and dropping the coffeescript dependency to boot).

It looks like the removal of the WebApp type decls might have made `imports/server/assets.ts` no longer typecheck?  Might need an explicit `any` or some other type hint on https://github.com/deathandmayhem/jolly-roger/blob/395d9baea3026f792b64352fb949d38d60882b2b/imports/server/assets.ts#L85

tests/acceptance/lib.ts
ebroder: Yeah, correct on both counts.

ebroder: (Specifically, `meteor-node-stubs` pulls in https://npmjs.com/package/util)

ebroder:
> It looks like the removal of the WebApp type decls might have made `imports/server/assets.ts` no longer typecheck? Might need an explicit `any` or some other type hint on
> 
> https://github.com/deathandmayhem/jolly-roger/blob/395d9baea3026f792b64352fb949d38d60882b2b/imports/server/assets.ts#L85

Ah yeah - I forgot that upstream requested that I declare that as `unknown` instead of `any`, so there's a bit more type nudging required. Pushing a fix momentarily.

zarvox (approved):
Lovely, thanks for doing this!

PR #1129, opened 2022-11-05
Almost completely eliminate Promise.await

ebroder:
As Meteor moves towards deprecating support for fiber-based code execution, the `Promise.await` method represents a code smell. Now that we have access to async (or at least "async") database operations, we can mostly move away from needing to forcibly fiber-ize otherwise perfectly good promise-based code.

That's mostly what this PR does - replace uses of `Promise.await` with native `await`, propagating up the call stack as needed. To support that, it also modifies a few of our internal frameworks (HookRegistry, MigrationRegistry) to support both sync and async implementations. In general, I tried to stay away from behavioral changes - if we previously ran operations serially, I didn't parallelize them here (even if the parallelized promise could would be straightforward to write).

This PR leaves behind two uses of `Promise.await`, both of which are necessary due to lack of comprehensive async support in Meteor:
- In `imports/server/discord.ts`, we integrate with Meteor's native OAuth support for linking users' Discord profiles. Currently, the `handleOauthRequest` method must be synchronous. It looks like this should be fixed in 2.9 once meteor/meteor#12156 is available, at which point the conversion should be relatively straightforward.
- In `imports/server/mediasoup-api.ts`, we call an async function from within a publish method. This is supported by Meteor, but not by the type declarations. I opened DefinitelyTyped/DefinitelyTyped#63095 with a fix, at which point we can do a fairly straightforward conversion.

This should be reviewable commit-by-commit, although I think the changes are mostly sufficiently rote that you could probably just review the PR as a whole.


imports/server/garbage-collection.ts
```diff
@@ -37,15 +37,18 @@ function cleanup() {
   }
 
   // Run all hooks.
-  globalGCHooks.forEach((f) => f(deadServers));
+  await globalGCHooks.reduce(async (p, f) => {
+    await p;
+    await f(deadServers);
+  }, Promise.resolve());
```
ebroder, line 43: Hat tip to [this StackOverflow answer](https://stackoverflow.com/a/24586168) (and to VSCode's "convert this function to async/await" helper) for this pattern, which I found to be quite pleasant for cases where you basically want to run through a loop and have each step run sequentially.

imports/lib/models/Base.ts
```diff
@@ -64,6 +64,14 @@ class Base<T extends BaseType> extends Mongo.Collection<T> {
     return super.insert(<any>doc, callback);
   }
 
+  // @ts-expect-error See above
+  insertAsync<U>(
+    doc: ValidateShape<U, Partial<T>>,
+    callback?: (err?: Error, id?: string) => void,
+  ): Promise<string> {
+    return super.insertAsync(<any>doc, callback);
```
zarvox, line 72: I find it odd that `insertAsync` still takes a callback parameter?  Wouldn't we just `reject` the promise instead?

imports/server/assets.ts
```diff
@@ -55,7 +54,7 @@ const hashes = [
   });
 });
 
-MeteorPromise.awaitAll(hashes);
+await Promise.all(hashes);
```
zarvox, line 57: Are...are we allowed to just have awaits at global scope?  https://github.com/meteor/meteor/issues/12287 suggests that we can't yet.

imports/server/gdrive.ts
```diff
@@ -149,63 +149,68 @@ export function makeReadOnly(fileId: string): void {
...
-  MeteorPromise.await(client.permissions.create({
+  await client.permissions.create({
     fileId,
     requestBody: { role: 'reader', type: 'anyone' },
-  }));
+  });
 
   // Delete any editor permissions
-  permissions.forEach((permission) => {
+  await permissions.reduce(async (promise, permission) => {
```
zarvox, line 158: Oh that's a cute way to phrase that with promise chains

zarvox (approved):
Looks mostly good, modulo a couple things that I wasn't sure about.  Approving since I expect they're trivial enough to fix

imports/lib/models/Base.ts
ebroder: I admittedly didn't think very hard about this, but was just copying what the upstream type declarations did. But it does seem kind of odd. I'll switch

imports/server/assets.ts
ebroder: I also found that PR, but when I tested locally this seemed to work. I suspect it's possible that one will block the application from starting up and one won't, in which case potentially if you load the app very quickly after the server starts you might see problematic behavior (although maybe we should use `Meteor.startup` to get that behavior explicitly).

ebroder: Did some very basic testing with a `while true; do curl ...` loop, and according to that test there was no race at startup. I'm not willing to say definitively that there's no race, but if there is it's narrow enough that I don't feel obliged to track it down.

PR #1130, opened 2022-11-05
Upgrade simpl-schema to the latest version

ebroder:
This requires a little rooting around to get at the appropriate types, but it is doable.

### DIFF
--- a/imports/lib/schemas/typedSchemas.ts
+++ b/imports/lib/schemas/typedSchemas.ts
@@ -1,8 +1,20 @@
 import { Meteor } from 'meteor/meteor';
 import * as t from 'io-ts';
 import { date } from 'io-ts-types';
-import SimpleSchema, { SchemaDefinition, SimpleSchemaDefinition } from 'simpl-schema';
+import SimpleSchema from 'simpl-schema';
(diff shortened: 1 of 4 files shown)


imports/lib/schemas/typedSchemas.ts
```diff
@@ -1,9 +1,21 @@
 import { Meteor } from 'meteor/meteor';
 import * as t from 'io-ts';
 import { date } from 'io-ts-types';
-import SimpleSchema, { SchemaDefinition, SimpleSchemaDefinition } from 'simpl-schema';
+import SimpleSchema from 'simpl-schema';
+import type { FieldInfo, SchemaDefinitionWithShorthand, TypeDefinitionProps } from 'simpl-schema/dist/esm/types';
```
zarvox, line 5: Ahhhh, I struggled to figure out how to phrase this to get access to the types, since that file isn't actually packaged by the `package.json` -- good to know

zarvox (approved):

imports/lib/schemas/typedSchemas.ts
ebroder: I just dug around with auto-complete until I found it :joy:

PR #1149, opened 2022-11-17
Await on critsection of lock before cleaning up

ebroder:


ebroder:
This is a (at least partial) fix from #1129

zarvox (approved):

PR #1161, opened 2022-11-25
Remove all direct (and some indirect) uses of underscore

zarvox:
It seems that the Meteor ecosystem is moving away from underscore, at least on the client-side, which means that if we finish removing some of our remaining uses of it, we might be able to save a decent chunk of code size on the client.

This PR removes our remaining direct usages by reimplementing the relevant subset of functionality, and also removes one indirect usage via `xolvio:cleaner` by reimplementing the single serverside function we call directly.  (This also affords us the opportunity to move some of those DB interactions to use non-deprecated methods and async/await while we're there).

The one big remaining dependency that appears to be pulling in `underscore` on the client side is `meteorhacks:cluster`, which is probably a large enough rewrite that we should take it on as a separate PR.

### DIFF
--- a/tests/acceptance/lib.ts
+++ b/tests/acceptance/lib.ts
@@ -4,8 +4,8 @@ import { promisify } from 'util';
 import { check } from 'meteor/check';
 import { DDP } from 'meteor/ddp';
 import { Meteor } from 'meteor/meteor';
+import { MongoInternals } from 'meteor/mongo';
 import { Tracker } from 'meteor/tracker';
-import { resetDatabase as xolvioResetDatabase } from 'meteor/xolvio:cleaner';
 import TypedMethod from '../../imports/methods/TypedMethod';
 
 export const USER_EMAIL = 'jolly-roger@deathandmayhem.com';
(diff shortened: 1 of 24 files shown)


tests/acceptance/lib.ts
```diff
@@ -53,7 +53,20 @@ if (Meteor.isServer) {
...
       currentConn = this.connection;
 
-      xolvioResetDatabase();
+      // Remove all the contents of all not-mongo-internal collections
+      const excludedCollections = ['system.indexes'];
+      const db = MongoInternals.defaultRemoteCollectionDriver().mongo.db;
+      const collections = await db.collections();
+      const appCollections = collections.filter((col) => {
+        // Exclude velocity collections and system indexes.
+        return col.collectionName.indexOf('velocity') !== 0 && excludedCollections.indexOf(col.collectionName) === -1;
```
ebroder, line 62: I'd be slightly more comfortable if this was `col.collectionName.startsWith('system.')`

ebroder (approved):

tests/acceptance/lib.ts
zarvox: Can do.  This was just the existing `xolvio:cleaner` behavior.

PR #1163, opened 2022-11-25
Upgrade to modern version of marked

### LINKED ISSUE #1162: Reevaluate customization of markdown formatting
We're using a somewhat old version of `marked` these days. It's something of a broken window, as there's an outstanding regex DoS vulnerability, but I haven't dealt with it, mostly because I haven't wanted to fight with our customization of its formatting. Before fighting with our customization, I think it's worth having a discussion about whether we actually care.

We've historically made modifications to how `marked` formats Markdown to make it consistent with how Slack formatted text - Slack uses `*` for bold and `_` for Italics, instead of Markdown standard (`*` or `_` for italics and `**` or `__` for bold). I think a decent number of chat apps (e.g. Skype) uses the same rough formatting rules as Slack, but Discord and Github both use standard markdown. This means that when we pipe chat messages into Discord, they get formatted differently.

My intuition is that we should just switch to default behavior and let people adjust. It'll mis-format old messages but not dramatically, and people may be slightly confused, but realistically the difference between bold and italic is mostly pretty slight.

ebroder:
Fixes #1162 (and as a side-effect, changes Markdown to standard formatting).


imports/client/markdown.ts
```diff
@@ -1,17 +1,6 @@
 import DOMPurify from 'dompurify';
-import marked from 'marked';
-
-(marked.InlineLexer.rules.gfm as marked.Rules).em = /^\b_((?:__|[^_])+?)_\b/;
-(marked.InlineLexer.rules.gfm as marked.Rules).strong = /^\*\b((?:\*\*|[^*])+?)\b\*/;
-
-const renderer = new class extends marked.Renderer {
-  link(href: string, title: string, link: string) {
-    const realLinkFunc = marked.Renderer.prototype.link.bind(this);
-    return realLinkFunc(href, title, link)
```
zarvox, line 10: I think we probably do still want to make links open in a new tab (`target="_blank"`), with `noopener`/`noreferrer`?

ebroder: Oh yeah, good call. Fixed

zarvox (approved):

PR #1166, opened 2022-11-25
Replace `meteorhacks:cluster` process load balancer

zarvox:
This follows the same overall architecture (same dependencies, same general flow), but drops large chunks of functionality we did not use, and is basically completely rewritten with ES6 classes and TypeScript.

Incidentally, as `meteorhacks:cluster` was the last remaining client-side importer of `underscore`, this change has the side effect of reducing the bundle size from 1.77 MB to 1.75 MB.

To test this, note that you'll need to set `CLUSTER_WORKERS_COUNT` to either `auto` or the number of additional workers you'd like to spawn.  We do this in production but only for instances with RAM exceeding a certain size, so we may not have a ton of coverage for this behavior.


imports/server/WorkerPool.ts
```diff
@@ -0,0 +1,169 @@
...
+    this.closed = false;
+    for (let i = 0; i < size; i++) {
+      this.createWorker();
+    }
+    this.recentReconnects = 0;
+    this.lastReconnectAt = 0;
+
+    (['SIGINT', 'SIGHUP', 'SIGTERM'] as const).forEach((signal: HandledSignal) => {
+      process.once(signal, this.cleanup.bind(this));
+    });
```
ebroder, line 65: Hmm. Is this problematic given what we ran into with #483?

imports/server/LoadBalancer.ts
```diff
@@ -0,0 +1,173 @@
...
+//
+// Otherwise, it spawns no children, and the listeners simply propagate the
+// requests as middleware usually do, which causes the requests to be serviced
+// from the main process.
+export default class LoadBalancer {
+  proxy: HttpProxy;
+
+  workers: WorkerPool;
+
+  workerMapping: Record<string, { worker: Worker; lastUpdate: number; }>;
```
ebroder, line 33: There's no garbage collection of this, right?

imports/server/WorkerPool.ts
zarvox: I don't think so?  This is logic that previously existed in the dependency, and the `cleanup` method below also re-signals itself after the hook.

ebroder (approved):
OK. I agree that this reimplements what `meteorhacks:cluster` was doing, and is an improvement in that it doesn't carry a dependency on a package that hasn't been updated in 7 years.

There's some stuff in the original implementation that's carried forward here that I'm not wild about, but nothing that seems like an actual showstopper. This also feels like a case where our needs shouldn't be very differentiated from other users, which makes me skeptical of having this code live in our codebase long term; I'd prefer to outsource undifferentiated logic like that to external libraries where possible.

I think that means that we should see if we can sub out the `cluster` module at some point.

imports/server/LoadBalancer.ts
zarvox: None at the moment (which matches the old upstream's behavior).  On errors, we [delete the entry](https://github.com/deathandmayhem/jolly-roger/pull/1166/files/dc423e5bc55568fa7959deeea46d38215ba3a2a4#diff-a74780b1bdb4e723525910156555ece545b6f8b0e739193495e470a6e8fa7ef4R78) so we'll pick a new backend at random, but we don't do any behavior like "if we haven't seen a request from this sockjs session in N minutes, drop the backend mapping".

zarvox:
> OK. I agree that this reimplements what `meteorhacks:cluster` was doing, and is an improvement in that it doesn't carry a dependency on a package that hasn't been updated in 7 years.
> 
> There's some stuff in the original implementation that's carried forward here that I'm not wild about, but nothing that seems like an actual showstopper. This also feels like a case where our needs shouldn't be very differentiated from other users, which makes me skeptical of having this code live in our codebase long term; I'd prefer to outsource undifferentiated logic like that to external libraries where possible.
> 
> I think that means that we should see if we can sub out the `cluster` module at some point.

I agree.  FWIW, I suspect we'd also probably do fine removing this approach to multi-process behavior entirely.  To date, we've generally not been CPU-bound; what CPU-heavy work we do (mediasoup worker) is already farmed out to separate processes, and we tend to mostly be blocked on I/O or database throughput, which actually favors having /fewer/ workers trying to coordinate through DB writes.  I guess the tradeoffs are that we lose some potential parallelism which might improve latency when utilization is low, and we also have potentially larger blast radius if we write a bug that takes the process down entirely.

That may change if we add more compute-intensive work (e.g. complicated string-matching for dingword notification rules), but in the meantime I think it's fair.  How much do we really want to keep any of this?

ebroder:
> To date, we've generally not been CPU-bound; what CPU-heavy work we do (mediasoup worker) is already farmed out to separate processes

We are somewhat architecturally limited today, because mediasoup workers and Meteor processes are one-to-one. I guess if we broke that I'd be reasonably comfortable dropping down to a single Meteor process per instance.

PR #1174, opened 2022-11-30
Migrate from launch configurations to launch templates

ebroder:
Launch templates are AWS's newer way of specifying an instance and are replacing launch configurations. AWS recently sent out a notice that new instance types won't be supported in launch configurations going forward, so it's time for us to switch. The migration is fairly straightforward, although it does require an extra layer of indentation (thanks, YAML) which makes the diff look worse than it is. It's pretty minimal with a word diff.

### DIFF
--- a/cloudformation/jolly-roger.yaml
+++ b/cloudformation/jolly-roger.yaml
@@ -405,299 +405,301 @@ Resources:
         - Type: forward
           TargetGroupArn: !Ref AppAlbTargetHTTPS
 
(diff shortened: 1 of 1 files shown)


ebroder:
This is already running in production without issue. Best reviewed with a whitespace-agnostic diff: https://github.com/deathandmayhem/jolly-roger/pull/1174/files?diff=split&w=1.

zarvox (approved):

PR #1183, opened 2022-12-04
Mute when joining a large call

### LINKED ISSUE #501: When joining large calls, default to starting out muted
In larger group real-time conversations, it's often preferable that newcomers not generate noise until they intend to.  Last year, we got feedback that some people were surprised that voice-chat was mic-on by default.

We might want to pick a threshold of call participants N for which, if you are just now joining the call and there are already N participants, you will start out muted, to help keep large calls tidy, especially as people come and go.

zarvox:
This improves the experience of using audiocalls that have a whole bunch of people listening in at once.

We achieve this by adding a field on `Peer` that indicates the server's desired stream states, and then having the client respect that, showing a popover message if the server has requested they be muted.

To avoid producing strange effects on server disconnection, we also track more explicitly if we're creating a Peer where the same hunt/call/tab had already joined that call (indicating reconnection on server failure/disconnection, rather than initial join), and if so, propagate the last-known state from the previous Peer record (since we'd like for reconnections not to perturb client state).

Screenshot 2022-12-03 at 5 47 56 PM

Fixes #501.

### DIFF
--- a/imports/server/mediasoup-api.ts
+++ b/imports/server/mediasoup-api.ts
@@ -137,18 +137,50 @@ Meteor.publish('mediasoup:join', function (hunt, call, tab) {
     // sure that the client no longer trusts that the previous `call.join` is
     // active, and we can just take this as a signal to hasten that cleanup.
     // If the room is not yet created, create it.
-    await Peers.removeAsync({ hunt, call, tab });
+    const maybeOldPeer = await Peers.findOneAsync({ hunt, call, tab });
+    if (maybeOldPeer) {
+      Ansible.log('Removing peer with same hunt/call/tab', { peer: maybeOldPeer._id });
+      await Peers.removeAsync({
+        _id: maybeOldPeer._id, hunt, call, tab,
+      });
+    }
+
+    const peerCount = await Peers.find({ hunt, call }).countAsync();
+    // If we are a new joiner and would be the 8th (or more) peer, join the
+    // call starting out muted, because large calls can get noisy.
+    // In local development, make this just 3 because it's annoying to open that many browser tabs.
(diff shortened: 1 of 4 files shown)


imports/server/mediasoup-api.ts
```diff
@@ -137,18 +137,43 @@ Meteor.publish('mediasoup:join', function (hunt, call, tab) {
...
+    // mute state was.
+    if (maybeOldPeer) {
+      let oldPeerState;
+      if (maybeOldPeer.deafened) {
+        oldPeerState = 'deafened' as const;
+      } else if (maybeOldPeer.muted) {
+        oldPeerState = 'muted' as const;
+      } else {
+        oldPeerState = 'active' as const;
+      }
```
ebroder, line 162: OOC, does TypeScript back-propagate the types from these assignments to `oldPeerState`, or does it effectively end up as `any`?

ebroder: Yeah looks like it ends up as `any`: https://tsplay.dev/mL5qVW

ebroder: Ah no, my mistake - it is `any` when it's first declared, but the type is narrowed by having an exhaustive set of if statements, so when the actual subsequent assignment happens, it has a real type: https://tsplay.dev/w1APXw

ebroder (approved):
LGTM modulo lint failures

imports/server/mediasoup-api.ts
zarvox: It appears to know the narrower type after the assignments in the flow.  (The type on line 155 is `any`, but by the time we reach line 163 it's `'active' | 'muted' | 'deafened'`.)

PR #1187, opened 2022-12-04
Remote mute implementation

### LINKED ISSUE #598: Remote mute
I'm quite anxious about the potential for abuse here, but there were plenty of situations where someone on voice didn't realize they were off mute and wasn't listening to audio for someone to tell them, so it seems like we should try and come up some way to do this.

A few things I'll note:
- From a technical perspective, this may be a little tricky. We somehow need to push state to the client so they can unmute themselves
- It seems problematic to limit this to operators - they're not in all audio calls, and we shouldn't be dependent on them jumping in to fix this sort of issue
- At a bare minimum, to prevent abuse, we probably want to block remote mute if the person has recently unmuted themselves (so that once they unmute, someone can't remote mute them again), or something along those lines. We really need to think carefully about this

Fixing #501 would almost certainly reduce the number of situations where this is necessary, but unfortunately I don't think it's likely to eliminate them.

ebroder:
Fixes #598


imports/server/methods/mediasoupRemoteMutePeer.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    const peer = Peers.findOne({ _id: peerId });
+    if (!peer) {
+      throw new Meteor.Error(404, 'Peer ID for remote peer not found');
+    }
+
+    const selfPeer = Peers.findOne({ call: peer.call, createdBy: this.userId });
+    if (!selfPeer) {
+      throw new Meteor.Error(404, 'Peer ID for self not found');
```
zarvox, line 33: maybe make this error message explain that you must be a member of the call for which you're trying to mute the specified peer?

imports/server/methods/mediasoupRemoteMutePeer.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    if (selfPeer.createdBy !== this.userId) {
```
zarvox, line 36: This block seems redundant

zarvox (approved):

imports/server/methods/mediasoupRemoteMutePeer.ts
ebroder: Fixed

PR #1189, opened 2022-12-05
Build moira from source

ebroder:
Debathena seems to be sluggish about updating to newer Ubuntu releases than 18.04, and with it about to go EOL, we need an alternative strategy. While it would be great if we didn't have a dependency on moira and blanche in the long term, this moves us forward in the short term.


Dockerfile
```diff
@@ -13,15 +13,48 @@ ENV DEBIAN_FRONTEND noninteractive
...
+		python3 python3-pip python3-dev python3-setuptools python3-wheel \
+		comerr-dev libkrb5-dev libreadline-dev libhesiod-dev libncurses5-dev autotools-dev
 
 	# Install chromium-browser's dependencies, which should match puppeteer's
 	# dependencies. Note: this will need to be updated when we upgrade to 20.04,
 	# as chromium-browser on 20.04 is a wrapper around a snap package (although
-	# `apt-get satisfy` will make it easier)
-	apt-get install --no-install-recommends -y $(apt-cache depends chromium-browser | sed -ne 's/^ *Depends://p')
+	# apt-get satisfy will make it easier)
+	apt-get install --no-install-recommends -y \$(apt-cache depends chromium-browser | sed -ne 's/^ *Depends://p')
```
zarvox, line 34: At a glance, I'm guessing the build failure about an unexpected `(` is from this `$` -> `\$` conversion here.

```suggestion
	apt-get install --no-install-recommends -y $(apt-cache depends chromium-browser | sed -ne 's/^ *Depends://p')
```

zarvox (approved):

PR #1197, opened 2022-12-07
Fix up Dockerfile

ebroder:
It turns out that the behavior of Dockerfile heredocs with an explicitly specified interpreter (i.e. `RUN <<EOF bash`) is a bit wonky and hard to understand. Much clearer is putting a shebang line in the heredoc - docker will handle that by spitting out a temporary file and executing it; much clearer.

However, that doesn't seem to work with whitespace stripping, so just unindent everything instead.

And finally, for consistency, switch all of the scripts to use bash (instead of the few cases where we were using dash) and make sure to set pipefail everywhere - even if it's not necessary, it's a good belt-and-suspenders protection.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,28 +10,30 @@ FROM ubuntu:18.04 AS buildenv
 ENV DEBIAN_FRONTEND noninteractive
 
 # Install build deps
-RUN <<-'EOF' bash
-	set -eux
-	apt-get update
-	# We need:
-	# * git: for fetching moira and capturing git rev in Meteor artifact
-	# * curl: for the Meteor installer and fetching new apt keys
-	# * gnupg: for installing new apt keys
-	# * python3 et al: for building mediasoup
-	# * comerr-dev et al: for building moira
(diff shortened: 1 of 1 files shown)


zarvox (approved):
Thanks for digging into this!

ebroder:
```
#29 60.19 + pip3 install credstash
#29 62.09 Collecting credstash
#29 62.23   Downloading https://files.pythonhosted.org/packages/39/8a/6f1c1637d59f16a59fd79adf2ba94a238796adc6e425c997ef2455c52b9d/credstash-1.17.1-py3-none-any.whl
#29 62.27 Requirement already satisfied: boto3>=1.1.1 in /usr/lib/python3/dist-packages (from credstash)
#29 62.27 Requirement already satisfied: cryptography>=2.1 in /usr/lib/python3/dist-packages (from credstash)
#29 62.28 Installing collected packages: credstash
#29 62.34 Successfully installed credstash-1.17.1
```

Looks like we're good.

PR #1198, opened 2022-12-08
Upgrade Docker build to Ubuntu 22.04

ebroder:
Now that we are no longer blocked on Debathena giving us a build of moira for modern releases, we can upgrade to jammy without a ton of work. A few specific changes:

* The `apt-key` command has been deprecated, in favor of just dropping files into `/etc/apt/trusted.gpg.d/`. However, this lets us drop our dependency on the `gnupg` package, as we were never using it directly.
* Ubuntu's `chromium-browser` package is now just a wrapper around a snap package, so we can't use that to ensure we have the dependencies for puppeteer. Instead, add the Google Chrome apt repository so we can get metadata on what Chrome-like dependencies look like (but don't actually install Chrome).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 # We use separate stages for running lint and test vs. building the production
 # bundle so that they can run in parallel
 
-FROM ubuntu:18.04 AS buildenv
+FROM ubuntu:22.04 AS buildenv
 
 ENV DEBIAN_FRONTEND noninteractive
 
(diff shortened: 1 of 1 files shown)


ebroder:
Note that this does not upgrade our production hosts to prod. Those are already running focal - 20.04 - so are slightly less pressing, and also somewhat harder to test (though hopefully even easier, since we depend on so little native platform capability), so I figured it was easier to split up those two changes.

zarvox (approved):

PR #1204, opened 2022-12-08
Upgrade hosts in CloudFormation to jammy

ebroder:
(As per usual for CloudFormation changes, this is already running in production without incident)


zarvox (approved):
Remarkably terse for a distro upgrade :)

PR #1213, opened 2022-12-15
Track activity on Google Drive documents

### LINKED ISSUE #107: Try to track user activity in google spreadsheets
Today, Jolly Roger only knows who's working on a puzzle by virtue of who has the puzzle page open, but that's pretty coarse - people tend to open lots of puzzle pages when they're skimming through, and leave them open even when they're not actively working on the puzzle.

It'd be nice if we could get an event changestream from Google Sheets to tell when someone was actively editing the sheet, since that's a much stronger indicator of activity.

ebroder:
This uses the Google Drive API's `files.watch` API method to receive webhooks when changes are made to a document. We setup a watch on every document in the database and handle renewing it as necessary.

When we're notified by Google that a document has been changed, we store a record in the database - effectively a tuple of `(timestamp, document)`. Absence of a record indicates lack of activity.

Google's notifications aren't especially timely - in practice they seem to send notifications at about 3 minute granularity - so we use 5 minute granularity for our own tracking. It's a nice round number, and pretty well approximates the granularity of the notifications.

Activity records are currently never garbage collected. This should be safe because generally puzzles don't see constant activity over the course of a hunt - e.g. in 2022, puzzles were open for an average of 5 hours. Assuming perfect density, that's a total of roughly 12,000 records - not nothing, but not completely unmanageable. We can always revisit in the future and add a expiration index if needed.

Since we're now asking an external service to reach out to us, we need some way of addressing the server from the internet. In production, this is easy. In development, if Meteor otherwise only knows itself to be accessible over localhost, we attempt to integrate with ngrok to setup a tunnel.

This commit only wires up collection of activity data; serving that to clients and displaying it in a useful way is left as an exercise for the reader (or more likely, me).

As it's written, I think this fixes #107, although arguably that's a bit of a cop-out given lack of any display of the data.

### DIFF
--- a/imports/server/gdriveDocumentWatcher.ts
+++ b/imports/server/gdriveDocumentWatcher.ts
@@ -1,0 +1,284 @@
+import { Meteor } from 'meteor/meteor';
+import { Random } from 'meteor/random';
+import { WebApp } from 'meteor/webapp';
+import express from 'express';
+import Ansible from '../Ansible';
+import Flags from '../Flags';
+import DocumentActivities from '../lib/models/DocumentActivities';
+import Documents from '../lib/models/Documents';
+import FeatureFlags from '../lib/models/FeatureFlags';
+import DriveClient from './gdriveClientRefresher';
+import ignoringDuplicateKeyErrors from './ignoringDuplicateKeyErrors';
+import DocumentWatches from './models/DocumentWatches';
+import Locks from './models/Locks';
+import { DocumentWatchType } from './schemas/DocumentWatch';
+
+/* expire more quickly in development to minimize requests hitting after server shutdown */
+const EXPIRE_WINDOW = Meteor.isDevelopment ?
+  5 * 60 * 1000 :
+  24 * 60 * 60 * 1000;
+const RENEW_WINDOW = 60 * 1000; // milliseconds
+const ACTIVITY_GRANULARITY = 5 * 60 * 1000; // milliseconds
+const WEBHOOK_PREFIX = '/_gdrive/watch';
+
+// Handle incoming webhooks from Google Drive
+const webhook = express();
+// Express's API is entirely callback-driven, so it doesn't matter if our
+// function is async (it just means that exceptions won't be caught).
+// eslint-disable-next-line @typescript-eslint/no-misused-promises
+webhook.post('/', async (req, res) => {
+  if (req.get('X-Goog-Resource-State') === 'sync') {
+    res.sendStatus(200);
+    return;
+  }
+
+  const documentId = req.get('X-Goog-Channel-Token');
+  if (!documentId) {
+    res.sendStatus(400);
+    return;
+  }
+
+  const document = Documents.findOne(documentId);
+  if (!document) {
+    res.sendStatus(404);
+    return;
+  }
+
+  // Round the current time to the nearest 5 seconds
+  const roundedTime = new Date(
+    Math.round(new Date().getTime() / ACTIVITY_GRANULARITY) * ACTIVITY_GRANULARITY
+  );
+  await ignoringDuplicateKeyErrors(async () => {
+    await DocumentActivities.insertAsync({
+      ts: roundedTime,
+      hunt: document.hunt,
+      puzzle: document.puzzle,
+      document: document._id,
+    });
+  });
+  res.sendStatus(200);
+});
+WebApp.connectHandlers.use(WEBHOOK_PREFIX, Meteor.bindEnvironment(webhook));
+
+// Setup and maintain watches on all Google Drive documents
+class GDriveDocumentWatcher {
+  rootUrl: string;
+
+  watchUrl: string;
+
+  documentWatchTimeouts: Map<string, number>;
+
+  documentObserver: Meteor.LiveQueryHandle;
+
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
     "monocle-ts": "^2.3.13",
     "mustache": "^4.2.0",
     "newtype-ts": "^0.3.5",
+    "ngrok": "^4.3.3",
```
ebroder, line 48: You mentioned in Discord discussion that you didn't want ngrok to be present in our production builds - how strongly do you feel about that?

It is, as it turns out, is tricky to arrange. The library is the easiest way to talk to a running ngrok daemon, but it also includes the ngrok daemon (or really, fetches it as a post-install). Making it a dev dependency doesn't work, as Meteor's build system won't let you import a dev dependency in application code:

```
W20221214-18:33:52.911(-8)? (STDERR) (node:41101) UnhandledPromiseRejectionWarning: Error: Cannot find module "/node_modules/ngrok/index.js". Try installing the npm package or make sure it is not a devDependency.
```

imports/server/schemas/DocumentWatch.ts
```diff
@@ -0,0 +1,32 @@
+import * as t from 'io-ts';
+import { date } from 'io-ts-types';
+import { Id } from '../../lib/schemas/regexes';
+import { Overrides, buildSchema } from '../../lib/schemas/typedSchemas';
+
+export const DocumentWatchCodec = t.type({
+  _id: t.string,
+  document: t.string,
+  watchId: t.string,
```
zarvox, line 9: I'm a little confused why we need both `watchId` and `_id` as separate fields here -- they seem to have the same lifetime and lifecycle, and last I checked I thought you could pass an explicit `_id` to `insert()` if you wanted to pick the ID in advance with `Random.id()`?

ebroder: They don't actually have the same lifecycle today - we update the `watchId` every time we refresh the watch, but the database document remains the same (i.e. the `_id` persists)

imports/FixtureHunt.ts
```diff
@@ -6,7 +6,7 @@ import { TagType } from './lib/schemas/Tag';
 type FixtureHuntType = Pick<HuntType, '_id' | 'name'> & {
   tags: Pick<TagType, '_id' | 'name'>[];
   puzzles: (Pick< PuzzleType, '_id' | 'title' | 'url' | 'expectedAnswerCount' | 'tags'> & {
-    guesses: Pick<GuessType, '_id' | 'guess' | 'state'>[];
+    guesses: (Pick<GuessType, '_id' | 'guess' | 'state'> & { additionalNotes?: string })[];
```
zarvox, line 9: I suspect the changes in this file belong in the other PR's branch

imports/lib/schemas/Guess.ts
```diff
@@ -19,11 +19,23 @@ const GuessFields = t.type({
   confidence: t.union([t.Int, t.undefined]),
   // The state of this guess, as handled by the operators:
   // * 'pending' means "shows up in the operator queue"
-  // * 'correct', "incorrect", and "rejected" all mean "no longer in the operator queue"
-  // * 'incorrect' answers should be listed or at least discoverable on the puzzle page
+  // * "intermediate", "correct", "incorrect", and "rejected" all mean "no longer in the operator
```
zarvox, line 22: This also belongs on the other branch

imports/server/gdriveDocumentWatcher.ts
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    if (shouldRestart && newRootUrl && enabledByFeatureFlag) {
+      Ansible.log('Starting Google Drive webhooks with public URL', { url: newRootUrl });
+      watcher = new GDriveDocumentWatcher(newRootUrl);
+    }
+  };
+  Meteor.setInterval(updateWatcher, 5 * 1000);
+  FeatureFlags.find({ name: 'disable.gdrive_watchers' }).observe({
+    added: () => { void updateWatcher; },
```
zarvox, line 280: Does this actually call `updateWatcher()`?  It looks to me like it's not invoking `updateWatcher` but maybe I'm confused by how promises are awaited

ebroder: Oh no, you're right. I originally wrote it as `added: updateWatcher`, which would have worked, but needed the extra arrow function for the void cast. Pushed a fix.

types/ngrok.d.ts
```diff
@@ -0,0 +1,10 @@
+declare module 'ngrok' {
+  namespace Ngrok {
+    interface Tunnel {
+      config: {
```
zarvox, line 4: Should this also include the `proto` field we reference on https://github.com/deathandmayhem/jolly-roger/pull/1213/files#diff-5f3a785e0ccb0416158647d90d43c17f2789d516b0d6bdb2ef7f8b0ee0f79e0eR242 ?

ebroder: No, we get that from upstream. The upstream type declarations were just missing `config` (They took a PR but I don't think they've released a new version yet)

zarvox (approved):

PR #1217, opened 2022-12-18
Fix document watches to try and avoid spamming everyone

ebroder:
First, when we re-create the document watcher, make sure we clean up the old timeouts we set. Second, use the p-throttle library to limit the number of requests we have open at a time.


zarvox (approved):

zarvox:
Failed attempt https://github.com/deathandmayhem/jolly-roger/actions/runs/3723265242/attempts/1 looks like an internal GitHub failure, so I hit the retry button.

PR #1224, opened 2022-12-20
Fix multiple guess queue bugs and add countdown spinner

zarvox:
Fixes:

* a sign bug in the client-side refreshing that caused us to busy-loop timeouts updating values until completed guesses were marked (1c119d890f65ff97b081cf4d5867e69405ecbf19)
* incorrect `observeChanges` usages in `JoinedObjectObserver` (2cdb64ee1cdc713a6eb8edb767f0d9786d55b358) which, through spurious `decref`s, was causing us to lose client-side record of the user who submitted a guess on pages where we did not otherwise have that record locally due to another subscription
* incorrect `observeChanges` usages in `PendingGuessWatcher` (32b967d2c9baa148c2e03fd42335c4987930a769) which we probably would never hit since we don't remove the `roles` field but it's good to be correct

Adds:

* a spinner which indicates when a serviced guess will disappear from being shown

What it looks like now:

https://user-images.githubusercontent.com/307325/208751321-dc2101f7-782f-455f-93a4-9e72ac0954ab.mov

### DIFF
--- a/imports/client/components/SpinnerTimer.tsx
+++ b/imports/client/components/SpinnerTimer.tsx
@@ -1,0 +1,94 @@
+import React, { useCallback, useEffect, useRef } from 'react';
+
+// A component which animates a grey spinner between startTime and endTime like
+//     _____         __            __            __            __            __
+//    /  _  \       /  |  _       /  |          /  |          /  |          /  |           |
+//   /  / \  \     /  /  \ \     /  /          /  /          /  /           \_/
+//  |  |   | | -> |  |   | | -> |  |   --| -> |  |       -> |  |       ->            ->
+//   \  \_/  /     \  \_/  /     \  \_/  /     \  \_         \_/
+//    \_____/       \_____/       \_____/       \__|
+//
+//   startTime                              halfway between                             endTime
+const SpinnerTimer = ({
+  width,
+  height,
+  startTime,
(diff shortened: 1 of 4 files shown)


imports/client/components/SpinnerTimer.tsx
```diff
@@ -0,0 +1,94 @@
...
+    <canvas
+      className={className}
+      ref={canvasRef}
+      width={width}
+      height={height}
+    />
+  );
+};
+
+export default SpinnerTimer;
```
ebroder, line 94: Can we just use FontAwesome's spinners with their timing customizations? (https://fontawesome.com/v6/docs/web/style/animate#customization)

zarvox: I don't see how I'd achieve the same visual effect with their any of their spinners.  Those spinners generally suggest no start or end point; here the interpolation (and in particular the closeness to the end) is kinda the goal of the animation, so that you are made aware that the thing is going to vanish (and possibly reflow other notifications) before you take an action on a target other than your intent.

If you have a particular glyph that you think fits the bill, I'm open to trying this, but I didn't see something that had the feel I was going for, because none of the animations change only part of the content from what I could tell.

ebroder (approved):

PR #1226, opened 2022-12-22
Switch from watching individual files to fetching all changes

ebroder:
Google Drive imposes secret rate limits on setting individual file watches, to the point that it's difficult to set up watches on all files in our production instance (roughly 1500 files) without running afoul of them.

So instead, we switch to using the `changes.list` method, which enumerates changes to all files. I initially overlooked this due to a poor understanding of its behavior, but in practice it has all of the data we need.

We could use `changes.watch` to get notified when new changes are available, but those notifications aren't sent out very frequently, and Google Drive's rate limits are high enough that we can just poll roughly every 15 seconds without introducing significant latency on updates. We use a single lock to ensure that only one server is polling for changes at a time, to ensure that multiple servers don't inadvertently stack up against the rate limits.

Therefore, for now, we no longer need any of the incoming webhook infrastructure from the previous implementation. (If we pursue a publicly available Jolly Roger instance in the future such that users are required to link their own Google Drives for document storage, we may need to fetch changes from enough distinct drives that we need to revisit this.)


imports/server/gdriveChangeFetcher.ts
```diff
@@ -0,0 +1,151 @@
...
+import Locks, { PREEMPT_TIMEOUT } from './models/Locks';
+
+const ACTIVITY_GRANULARITY = 5 * 60 * 1000; // milliseconds
+
+async function recordDriveChange(file: Pick<drive_v3.Schema$File, 'id' | 'modifiedTime'>) {
+  if (!file.id || !file.modifiedTime) {
+    return;
+  }
+
+  const document = Documents.findOne(file.id);
```
zarvox, line 20: I don't think Google's notion of the `file.id` is going to match our `Document`'s  `_id`.  I would expect it to match [`value.id`](https://github.com/deathandmayhem/jolly-roger/blob/18609e4a627f15766e6e24f4b7b365c9462fd887/imports/lib/schemas/Document.ts#L22).

imports/server/migrations/41-document-activities.ts
```diff
@@ -8,7 +7,5 @@ Migrations.add({
   async up() {
     await DocumentActivities.createIndexAsync({ hunt: 1 });
     await DocumentActivities.createIndexAsync({ document: 1, ts: 1 }, { unique: true });
-    await DocumentWatches.createIndexAsync({ document: 1 }, { unique: true });
```
zarvox, line 11: In general, how do we want to deal with collections that we explicitly abandon?  Would it make sense to add a migration with a `collection.drop()` here?

ebroder: We have historically just abandoned them and not explicitly cleaned them up. (You can see this in production with `jr_document_perms`, which was superseded by `jr_folder_perms` after #568 and with `jr_profiles` which was rolled into `users)

My general rule of thumb is that dropping a collection is irreversible (which makes the commits irrevertable) and scary, and none of our data is big enough that this is a major concern. But I could be convinced otherwise

imports/server/gdriveChangeFetcher.ts
ebroder: D'oh. Fixed.

ebroder: (And confirmed in local testing that the end-to-end does actually capture `DocumentActivity` records)

imports/server/migrations/41-document-activities.ts
zarvox: I can see where you're coming from, though I think that in general, our migrations are not safely revertable (and accordingly, we should be careful with what we put in them).  We do sometimes stage our rollouts to make them at least still smoothly operable at both version N-1 and version N.

I don't think we need to eagerly drop collections specifically?  I'm just thinking about the possibility that future us decide to make a collection which happens to be named the same as one that we have long since forgotten about, but where the production DBs still have data (possibly in a different schema) and that creates trouble (e.g. a migration creates a unique index on a key that the prod DB contents lack, or the leftover indexes prevent inserting new records with different contents).  I guess we could also do something like keep tombstone collections around so we'd be forced to think about what to do with the dangling records if we ever try to return to them later?

The upside of dropping the collection is that we don't have to remember the constraints imposed by the previously-abandoned data.  Maybe it'd be reasonable for us to clean these abandoned collections up after some purgatory period (a few months?  one Hunt?)

We can totally solve that in a separate PR, though.

imports/server/gdriveChangeFetcher.ts
```diff
@@ -0,0 +1,151 @@
...
+import Locks, { PREEMPT_TIMEOUT } from './models/Locks';
+
+const ACTIVITY_GRANULARITY = 5 * 60 * 1000; // milliseconds
+
+async function recordDriveChange(file: Pick<drive_v3.Schema$File, 'id' | 'modifiedTime'>) {
+  if (!file.id || !file.modifiedTime) {
+    return;
+  }
+
+  const document = Documents.findOne({ 'value.id': file.id });
```
zarvox, line 20: We might want to add an index on this field if we're querying by it?

zarvox (approved):
LGTM but also feel free to add a migration to index `value.id` on `Document`.

imports/server/gdriveChangeFetcher.ts
ebroder: Huh yeah I guess I assumed we had this already. Fixing.

zarvox (approved):
Thanks!

PR #1227, opened 2022-12-22
Feature flag improvements

ebroder:
This eliminates the "random_by" feature flag mode, which as far as I'm aware we have never used, and uses the resulting simplification to introduce a `Flags.observeChanges` helper. I wanted the latter after feeling like I had written that logic too many times (including in #1226; if this lands first I'll update it)

### DIFF
--- a/imports/Flags.ts
+++ b/imports/Flags.ts
@@ -1,11 +1,9 @@
-import { Match, check } from 'meteor/check';
-import { SHA256 } from 'meteor/sha';
+import { check } from 'meteor/check';
 import FeatureFlags from './lib/models/FeatureFlags';
 
 const Flags = {
-  active(name: unknown, shard?: unknown) {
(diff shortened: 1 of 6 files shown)


imports/Flags.ts
```diff
@@ -1,5 +1,4 @@
 import { Match, check } from 'meteor/check';
-import { SHA256 } from 'meteor/sha';
 import FeatureFlags from './lib/models/FeatureFlags';
 
 const Flags = {
```
zarvox, line 4: Maybe also drop the `shard?` argument on the next line too, as it's now unused?

zarvox (approved):
We can remove a couple more lines but otherwise LGTM

imports/Flags.ts
ebroder: Oh good catch. I'll clean that up

Issue #600, opened 2022-01-16
Some UI for re-spacing guesses


ebroder:
In #541, @grantaelliott specifically called out

> Historically, hunters and operators have been inconsistent about spaces in answers. Hopefully, display flexibility will encourage us to standardize on preserving natural spaces.

However, there's no way today to edit the spacing in a guess once it's been submitted (short of using the JS console). It would be nice if there was some way for the operator to canonicalize a guess (ideally with guard-rails to prevent accident edits of content) based on, e.g., the way the answer is formatted back at them from the website when the right answer goes in.

grantaelliott:
+1

PR #1229, opened 2022-12-22
Guess Improvements: Round 2

### LINKED ISSUE #231: Track instruction submission
There's no way today to create a guess for an instruction rather than an answer (or to signal that a guess is for an instruction)

Not sure exactly how we should handle this, but one option would be a checkbox in the guess submission page. Notably marking an instruction as "correct" shouldn't get copied over to the puzzle answer, because it's not an answer.

ebroder:
This introduces UI for collecting additional notes about a guess as part of transitioning a guess into the "incomplete" or "rejected" state.

Here's what the flow looks like on desktop:

https://user-images.githubusercontent.com/28167/209242450-a26412ee-3c44-46a1-b7e4-c78a6636e79b.mov

and on mobile:

https://user-images.githubusercontent.com/28167/209242465-ccda2a85-ef0d-4b28-a007-6e227a955617.mov

Fixes #231.

### DIFF
--- a/client/stylesheets/_theme.scss
+++ b/client/stylesheets/_theme.scss
@@ -5,6 +5,8 @@ $font-family-base: $font-family-sans-serif;
 $link-decoration: none;
 $link-hover-decoration: underline;
 
+$toast-max-width: 380px;
+
 @import "../../node_modules/bootstrap/scss/functions";
 @import "../../node_modules/bootstrap/scss/variables";
 @import "../../node_modules/bootstrap/scss/mixins";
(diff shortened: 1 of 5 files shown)


client/stylesheets/_theme.scss
```diff
@@ -5,6 +5,8 @@ $font-family-base: $font-family-sans-serif;
 $link-decoration: none;
 $link-hover-decoration: underline;
 
+$toast-max-width: 380px;
```
ebroder, line 8: We need a little more width if we want to actually fit all 4 buttons on a single row. There's a little bit of buffer here because I expect to followup this PR with a fix for #600, which means that "Correct" is also going to get an ellipsis.

imports/client/components/NotificationCenter.tsx
```diff
@@ -52,7 +54,7 @@ const StyledNotificationActionBar = styled.ul`
   list-style-type: none;
   margin: 0;
   padding: 0;
-  flex-direction: row;
+  flex-flow: wrap row;
```
ebroder, line 57: More width (see above) helps on the desktop, but some phone screens are still too small to fit them all on one row, so we need to be willing to wrap.

zarvox (approved):
Looks great!

PR #1234, opened 2022-12-24
Avoid use of the same <Tooltip> `id` in multiple places

zarvox:
Each of the following ids are used by both the `GuessQueuePage` data table as well as the `NotificationCenter` toast:

- `guess-${guess._id}-direction-tooltip`
- `guess-${guess._id}-confidence-tooltip`
- `guess-${guess._id}-copy-tooltip`

We should avoid reusing ids on the page.  One way to help achieve that is to encode something about the file in which the tooltip appears in the id string itself, so I switched the ones in the NotificationCenter to include `notification-`.

These were the only reused `id`s I found.

### DIFF
--- a/imports/client/components/NotificationCenter.tsx
+++ b/imports/client/components/NotificationCenter.tsx
@@ -115,22 +115,22 @@ const GuessMessage = React.memo(({
   }, [onDismiss, guess._id]);
 
   const directionTooltip = (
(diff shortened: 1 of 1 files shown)


ebroder (approved):
I wonder if we can, like, throw an error in the smoke tests if we ever notice duplicate IDs being rendered to the DOM...

PR #1235, opened 2022-12-24
Avoid browser-driven form submission when submitting Rejected reason

zarvox:
The default event handler for Clicking a button with `type="submit"` causes a form submission which causes a navigation.

We could also have kept the `type="submit"` and overridden an event handler to call `preventDefault()` if we wanted to be more semantically-correct, but this seemed simpler.


ebroder (approved):
I think I originally did this for the benefit of mobile, but couldn't actually get it to change the "enter" button into a more submit-looking "done" or "go" or whatever button.

PR #1236, opened 2022-12-24
Collect consolidated activity tracking data

ebroder:
As part of addressing #172, we need better time-series data on how many users are working on any given puzzle. This collects that data for calls and chats. Together with the data we already get from documents, this should be enough for us to aggregate and display that data.

The first two commits are mostly cleanup and refactor to support the third commit, which is effectively the crux of this PR. The commit message is (IMO) important enough that I'll just paste it here for reference:

> In order to display chat and call activity data to clients without processing or shipping an overwhelming amount of data, we need to compress it in two ways. First, we need to aggregate multiple operations from a single user within the same time window. Second, we need to aggregate a count of unique users per time window.
> 
> This introduces a two-stage process for performing that aggregation for both chat and call data. First, as activity occurs, we create "recent" activity records. These dedup multiple events from a single user within our activity window but are otherwise unaggregated and represent a convenient queue of records for further consolidation. After our activity window has passed, we aggregate these recent records into summary records, which only contain unique user counts.
>
> Since Google Drive data does not require aggregation for either reason (we don't currently get per-user data and we only see records roughly every 3-5 minutes anyway), this preserves Google Drive document activity as a separate collection.
>
> As an implementation note, the consolidation processes takes advantage of MongoDB's transaction support to ensure that we delete recent records and increment consolidated records atomically. Transactions are supported by MongoDB at the version we run, and the resulting oplog entries are understood by Meteor. However, it's not possible to create a transaction using Meteor's MongoDB APIs, so we have to use the underlying driver directly.
>
> Note that this doesn't yet publish the resulting activity data to the client.

### DIFF
--- a/imports/server/activityTracking.ts
+++ b/imports/server/activityTracking.ts
@@ -1,0 +1,123 @@
+import { Meteor } from 'meteor/meteor';
+import { MongoInternals } from 'meteor/mongo';
+import { Random } from 'meteor/random';
+import Ansible from '../Ansible';
+import { ACTIVITY_GRANULARITY } from '../lib/config/activityTracking';
+import ConsolidatedActivities from './models/ConsolidatedActivities';
+import RecentActivities from './models/RecentActivities';
+import { RecentActivityType } from './schemas/RecentActivity';
+
+interface ActivityKey {
+  ts: Date;
+  hunt: string;
+  puzzle: string;
+}
+
+function serializeActivityKey({ ts, hunt, puzzle }: ActivityKey) {
+  return `${ts.getTime()}-${hunt}-${puzzle}`;
+}
+
+function deserializeActivityKey(key: string) {
+  const [ts, hunt, puzzle] = key.split('-');
+  if (!ts || !hunt || !puzzle) {
+    throw new Error(`Invalid activity key: ${key}`);
+  }
+  return { ts: new Date(parseInt(ts, 10)), hunt, puzzle };
+}
+
+async function fetchAndDeleteRecentActivity(
+  cutoff: Date,
+  session: InstanceType<typeof MongoInternals.NpmModules.mongodb.module.ClientSession>,
+) {
+  if (!session) {
+    throw new Error('Must be called with a session');
+  }
+
+  const usersByKeyAndType = new Map<string, Map<RecentActivityType['type'], Set<string>>>();
+  while (true) {
+    /* eslint-disable no-await-in-loop */
+    const recent = await RecentActivities.rawCollection().findOneAndDelete(
+      { ts: { $lt: cutoff } } as any,
+      { session }
+    );
+    if (!recent.value) {
+      break;
+    }
+
+    const key = serializeActivityKey(recent.value);
+    const usersByType = usersByKeyAndType.get(key) ?? new Map();
+    const users = usersByType.get(recent.value.type) ?? new Set();
+    users.add(recent.value.user);
+    usersByType.set(recent.value.type, users);
+    usersByKeyAndType.set(key, usersByType);
+    /* eslint-enable no-await-in-loop */
+  }
+
+  return usersByKeyAndType;
+}
+
+async function consolidateRecentActivity() {
+  const ageForConsolidation = ACTIVITY_GRANULARITY * 2;
+  const cutoff = new Date(Date.now() - ageForConsolidation);
+
+  // By using a transaction, we avoid races with other servers, since we only
+  // count records that we delete (and if we fail to update the consolidated
+  // counters, the deletions will be rolled back).
+  //
+  // However: note that using the raw collection bypasses our schema
+  // enforcement, so we need to make sure of compliance
+  const { client } = MongoInternals.defaultRemoteCollectionDriver().mongo;
+  const session = client.startSession();
+  await session.withTransaction(async () => {
+    const usersByKeyAndType = await fetchAndDeleteRecentActivity(
+      cutoff,
+      session,
+    );
+
+    await [...usersByKeyAndType.entries()].reduce(async (p, [key, usersByType]) => {
+      await p;
+
+      const matcher = deserializeActivityKey(key);
+      const increments = new Map<'total' | `components.${RecentActivityType['type']}`, number>();
+      const allUsers = new Set<string>();
+      [...usersByType.entries()].forEach(([type, users]) => {
+        const incrementKey = `components.${type}` as const;
+        increments.set(incrementKey, (increments.get(incrementKey) ?? 0) + users.size);
+        users.forEach((user) => allUsers.add(user));
+      });
+      increments.set('total', allUsers.size);
+
+      await ConsolidatedActivities.rawCollection().updateOne(
+        { ...matcher },
+        {
+          $setOnInsert: {
+            _id: Random.id(),
+          },
+          $inc: Object.fromEntries(increments),
+        },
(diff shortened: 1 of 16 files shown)


imports/server/mediasoup.ts
```diff
@@ -335,12 +339,32 @@ class SFU {
...
-        hunt: observerAppData.hunt,
-        call: observerAppData.call,
-      }, { $set: { lastActivity: new Date() } });
-    }), RECENT_ACTIVITY_TIME_WINDOW_MS);
+    const updateLastActivity = throttle(Meteor.bindEnvironment(
+      (volumes: types.AudioLevelObserverVolume[]) => {
+        CallHistories.upsert({
+          hunt: observerAppData.hunt,
+          call: observerAppData.call,
+        }, { $set: { lastActivity: new Date() } });
```
ebroder, line 347: I left our old call activity record around for the time being, but I suspect that we may be able to clean it up once all of the new activity stuff is implemented and displayed.

imports/server/mediasoup.ts
```diff
@@ -335,12 +339,32 @@ class SFU {
 
     const observerAppData = observer.appData as AudioLevelObserverAppData;
 
-    const updateLastActivity = throttle(Meteor.bindEnvironment(() => {
-      CallHistories.upsert({
-        hunt: observerAppData.hunt,
-        call: observerAppData.call,
-      }, { $set: { lastActivity: new Date() } });
-    }), RECENT_ACTIVITY_TIME_WINDOW_MS);
+    const updateLastActivity = throttle(Meteor.bindEnvironment(
```
zarvox, line 342: Hmmm, if I'm reading this right, this broad throttling mechanism means we'll likely only ever see one user at a time with appreciable `volume`, since most of the time only one person is talking at a time.  That kinda misses the point of counting active users by number of unique speakers in a time interval.

We probably want, instead, to throttle writes to `RecentActivities` on a per-user basis.

ebroder: Yeah, you're right. Actually, I think we don't want to throttle the `RecentActivities` write at all, since we're already creating the audio level observer with `threshold: 100`.

zarvox: I assume you meant `interval: 100`.  If so:

Do we really want to be doing 10 writes/sec/speaker for each (speaking) audio call participant?  That seems like an order of magnitude more than anything else we've done to date.  I'd favor throttling activity tracking writes to, say, one write per second per user.  Having a short interval means we notice shorter bursts of conversation; having a long throttle period means we avoid excessive quantities of writes that contribute little to our understanding of activity.

I think it should be correct to do throttling in memory on the backend that provides the call's router -- for any given call, we're guaranteed to only observe call activity on that one backend.

ebroder: Yeah, OK, you're right - pushed a new version that throttles per-second-per-user.

imports/server/migrations/43-puzzle-activity.ts
```diff
@@ -0,0 +1,21 @@
...
+import Migrations from './Migrations';
+
+Migrations.add({
+  version: 43,
+  name: 'Indexes for puzzle activity tracking',
+  async up() {
+    await RecentActivities.createIndexAsync({
+      ts: 1,
+      type: 1,
+      puzzle: 1,
```
zarvox, line 12: Any reason you left `hunt` out of the unique index here? Practically, I imagine it doesn't /really/ matter since puzzle IDs are expected to be unique, but I couldn't tell if this was intentional and if there's a performance or size or correctness reason to include it or not.

ebroder: Hmm. I think in general when I have unique indexes I prefer for them to be minimally defined (i.e. not include fields that don't contribute to the uniqueness), so my instinct would be to drop it from the `ConsolidatedActivities` index.

We'll need another index when we get around to querying anyway, since we need an index that starts with `hunt`

imports/server/activityTracking.ts
```diff
@@ -0,0 +1,118 @@
+import { Meteor } from 'meteor/meteor';
+import { MongoInternals } from 'meteor/mongo';
+import { Random } from 'meteor/random';
+import { ACTIVITY_GRANULARITY } from '../lib/config/activityTracking';
+import PuzzleActivities from './models/ConsolidatedActivities';
```
zarvox, line 5: nit: Maybe rename this import to `ConsolidatedActivities`?

imports/server/activityTracking.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  // By using a transaction, we avoid races with other servers, since we only
+  // count records that we delete (and if we fail to update the consolidated
+  // counters, the deletions will be rolled back).
+  //
+  // However: note that using the raw collection bypasses our schema
+  // enforcement, so we need to make sure of compliance
+  const { client } = MongoInternals.defaultRemoteCollectionDriver().mongo;
+  const session = client.startSession();
+  await session.withTransaction(async () => {
```
zarvox, line 70: What exactly happens here if the transaction aborts?  Does this backend throw a top-level exception that gets ignored, but then fails to reschedule?  Does it retry the transaction?  Something else?

zarvox (approved):

imports/server/migrations/43-puzzle-activity.ts
zarvox: Thanks for the explanation and guidelines!

imports/server/activityTracking.ts
ebroder: Bah yes, that's an oversight - I renamed the collection a few times while figuring out the flow. Will fix

imports/server/activityTracking.ts
ebroder: `withTransaction` is smart enough to retry the transaction if it knows that the transaction is retryable (I think this mostly happens in the event of a primary failover), but otherwise will rollback and throw. I think that would block rescheduling, so yes this should probably catch the exception so that we continue to re-schedule the consolidation.

imports/server/activityTracking.ts
```diff
@@ -0,0 +1,123 @@
...
+      });
+      increments.set('total', allUsers.size);
+
+      await ConsolidatedActivities.rawCollection().updateOne(
+        { ...matcher },
+        {
+          $setOnInsert: {
+            _id: Random.id(),
+          },
+          $inc: Object.fromEntries(increments),
```
zarvox, line 96: In normal operation, we don't expect to see these records getting updated, only inserted, right?  Because if we did, then we'd potentially double-count user ids.

ebroder: Oh hmm...there's some risk of concurrency - multiple servers could potentially be running consolidation simultaneously (and I assumed that would be fine because of the transactions). I may have to think a little more about how to do this correctly.

ebroder: This is actually safe - I sort of assumed that MongoDB would do the dumbest thing possible and allow dirty reads within a transaction (i.e. the equivalent of `READ UNCOMMITTED`). However, MongoDB does seem to do something more intelligent here. From [https://www.mongodb.com/docs/manual/core/transactions/#transactions-and-atomicity](Transactions and Atomicity):

> Until a transaction commits, the data changes made in the transaction are not visible outside the transaction.

So even if two servers are running this process concurrently, they won't see the other process's deletes, the transactions will conflict, and everything will be fine.

PR #1241, opened 2022-12-26
Activity sparklines

ebroder:
As a note, this doesn't actually deprecate any of our previous activity tracking systems. We show both puzzle "subscribers" (viewers) and the most recent call activity time in the call section of the puzzle page, and in both cases I think that information is better than what we have through general activity tracking.

Here's what the new activity section looks like:

Screenshot 2022-12-26 at 2 37 05 PM


zarvox (approved):
In a word: beautiful.  Thanks so much for speaking this into existence!

I ran `bundle-visualizer`; the `react-sparklines` dep adds 28.8kB in bundle size, which seems pretty reasonable for something as complicated as charting.

I also looked at things in a profiler and the whole thing seems quite fast and empirically seems to avoid spurious rerenders.

:shipit:

imports/client/components/PuzzleListPage.tsx
```diff
@@ -489,9 +489,8 @@ const PuzzleListPage = () => {
   const tagsLoading = useSubscribe('mongo.tags', { hunt: huntId });
   const loading = puzzlesLoading() || tagsLoading();
 
-  // Don't bother including these in loading - it's ok if they trickle in
-  useSubscribe('mongo.mediasoup_call_histories', { hunt: huntId });
-  useSubscribe('subscribers.counts', { hunt: huntId });
+  // Don't bother including this in loading - it's ok if they trickle in
+  useSubscribe('huntActivity', huntId);
```
zarvox, line 493: As a possible future consideration: since `huntActivity` is likely to be one of the larger datasets getting pushed down, to reduce time-to-first-useful-render, we might consider waiting to request `huntActivity` until the other higher-priority subs have finished loading?  It'd be counterproductive to starve the `mongo.puzzles` or `mongo.tags` subs of network bandwidth or CPU time because we're too busy loading `huntActivity`.  (It might also be that this never happens in practice due to the server responding to subs in the order they came in anyway?  🤷🏻 )

I could also believe that it might make sense on the client to defer rendering sparklines until the full initial dataset has made it down to the client (which would mean threading down a `huntActivityLoading()` somewhere), to avoid rerenders mid-load on incomplete history while streaming in initial sparkline data (which is somewhat more likely to cause recomputations/rerenders on the page), but maybe that's fine too.

PR #1246, opened 2022-12-28
Migrate away from synchronous mongo access in server code

ebroder:
This PR migrates us away from depending on synchronous mongo calls in server-side-only code. It does so by introducing a new custom ESLint rule banning usage of those methods, applying that rule's autofixer, and then cleaning up the carnage that it left behind.

Before adding and running the new rule, I did a little bit of cleanup work to make the conversion more straightforward:

- I overrode the Airbnb style guide (where we get most of our defaults) to explicitly allow for-of loops. As far as I can tell, the async version of for-of loops (`for await (i of ...)`) is currently the only way to stream results from a mongo query under the async APIs, which to me makes them an acceptable tradeoff. (A few other notes: The Airbnb style guide also suggests that `regenerator-runtime` is sufficiently heavyweight as to be undesirable, but the Meteor build config pulls it in anyway. This does not enable "for-in" loops, a different and older construct, which interact poorly with prototypical inheritance. We should still stay away from those.)
- I update `onExit` to accept async handlers.
- I enabled the `no-extra-parens` lint rule and applied its autofixes to our existing code (in advance of turning on the new custom rule). In order to ensure correctness, when the `no-sync-mongo-methods` rule autofixes an expression, it needs to add parentheses around the new `await` expression. However, this results in a lot of spurious parentheses. By having `no-extra-parens` enabled as well, it can clean up those parens without requiring manual cleanup. We could theoretically only turn the rule on in server code, but I'm a fan of consistency and generally didn't find its edits too offensive.

The overall diff on this PR is fairly sizeable, but hopefully it's not too unreasonable outside of the one autofix commit:

```
94ea35fc Other async fixups
 24 files changed, 140 insertions(+), 126 deletions(-)
bded96be Use sync mongo methods in a few places
 2 files changed, 4 insertions(+), 2 deletions(-)
fb74a4c1 Propagate async/await
 74 files changed, 130 insertions(+), 130 deletions(-)
27290a79 Enable and autofix no-sync-mongo-methods
 109 files changed, 334 insertions(+), 328 deletions(-)
90a5f4da Add an in-house lint rule to block sync mongo methods
 10 files changed, 260 insertions(+), 73 deletions(-)
eeb96136 Configure no-extra-parens lint rule
 26 files changed, 134 insertions(+), 115 deletions(-)
1d639de8 Allow onExit handlers to be async
 1 file changed, 9 insertions(+), 4 deletions(-)
12f96907 Allow for-of loop syntax
 2 files changed, 30 insertions(+)
```


ebroder:
Bah. If that doesn't fix the build I'll take another look at it tomorrow. It's clean locally but the local package is messing with somethings in Docker.

imports/server/accounts.ts
```diff
@@ -44,10 +44,10 @@ const summaryFromLoginInfo = function (info: LoginInfo) {
   }
 };
 
-Accounts.onLogin((info: LoginInfo) => {
+Accounts.onLogin(async (info: LoginInfo) => {
```
zarvox, line 47: `Accounts.onLogin` doesn't explicitly handle async functions, but it also doesn't explicitly block, so this seems fine.

imports/server/accounts.ts
```diff
@@ -116,6 +116,7 @@ const DEFAULT_ENROLL_ACCOUNT_TEMPLATE = 'Hiya!\n' +
     'This message was sent to {{email}}';
 
 function makeView(user: Meteor.User | null, url: string) {
+  // eslint-disable-next-line jolly-roger/no-sync-mongo-methods
```
zarvox, line 119: Ah, and this is needed because Meteor's `Email` package doesn't yet support `sendAsync()`, nor do the various `Accounts.emailTemplates` generator functions get awaited at their invocations ( https://github.com/meteor/meteor/blob/b43018cdd6a20744869a3b1f24c7a4c115fbef64/packages/accounts-base/accounts_server.js#L1399-L1421 ).  While the former [got added in Meteor 2.9](https://github.com/meteor/meteor/pull/12101), we can't take 2.9 until 2.9.1 due to the fetch() type bug, and I don't see the latter fixed upstream yet, so yeah, better to punt on this for now.

imports/server/discord.ts
```diff
@@ -184,6 +184,7 @@ class DiscordBot {
 }
 
 const handleOauthRequest = (query: any) => {
+  // eslint-disable-next-line jolly-roger/no-sync-mongo-methods
```
zarvox, line 187: Also implemented upstream but blocked on the Meteor 2.9 update: https://github.com/meteor/meteor/pull/12156/files#diff-6cc58ccee99c3992649e7b63e5df7590bdaae86d9dca037b2e573227b2249552R10

imports/server/getTeamName.ts
```diff
@@ -1,7 +1,7 @@
 import Settings from '../lib/models/Settings';
 
-function getTeamName(): string {
-  const teamNameObj = Settings.findOne({ name: 'teamname' });
+async function getTeamName(): Promise<string> {
```
zarvox, line 3: You appear to have not updated any of the call sites of `getTeamName()` in `imports/server/gdrive.ts` nor `imports/server/methods/updatePuzzle.ts` so I think those will wind up with `[object Promise]` instead of the team name.

I'm guessing maybe template literals are a blind spot for the unawaited-promises lint?  Can we disallow promises in template literals, since I can't imagine us ever wanting to print `[object Promise]`?

zarvox (changes requested):
Mostly good, modulo the build woes.  I spotted one bug converting a function to async without converting its callers.

imports/server/getTeamName.ts
ebroder: Good catch. I found a linter ([@typescript-eslint/no-base-to-string](https://typescript-eslint.io/rules/no-base-to-string/)) that catches this. Fortunately, it doesn't seem like we had any preexisting failures of that linter, so I feel relatively good about its precision. Should be fixed now.

zarvox: Excellent, thanks!

zarvox (approved):
Once you've sorted out the npm-in-Docker build issues, :shipit:

PR #1248, opened 2022-12-28
Make some improvements to lint config

ebroder:
- Enable a handful of new rules from typescript-eslint's strict set. This includes "no-non-null-asserted-nullish-coalescing", "non-nullable-type-assertion-style", "prefer-for-of", "prefer-return-this-type", and "prefer-string-starts-ends-with". These are mostly stylistic improvements and require very few changes to existing code, although "non-nullable-type-assertion-style" did catch an unintentional and unnecessary casting away of nullability.
- Pull all directives into the top-level .eslintrc.json. Previously, we had some environment declarations in the server and client directories, but we can just put those in top-level overrides. This is both more correct (as it now handles any directory with "client" or "server", not just the ones directly under imports/) and more discoverable.
- Be consistent about style - if we are only turning a rule on or off, we don't need to pass the new level as an array; we can just pass it directly as a string.


zarvox (approved):

ebroder:
Sorry - pushed one more thing if you don't mind taking another quick look. (Globally disabling some rules that have been tripping in ambient type declarations)

zarvox (approved):
Still LGTM :)

.eslintrc.json
```diff
@@ -231,7 +245,10 @@
       "rules": {
         // type declarations for other libraries don't get to choose their
         // export structure
-        "import/prefer-default-export": ["off"]
+        "import/prefer-default-export": "off",
+        // we often need to declare unused variables when extending generic
+        // types
+        "@typescript-eslint/no-unused-vars": "off"
```
zarvox, line 251: Actually, is this the rule that keeps us from accidentally forgetting to await promises?  I think this is probably useful to keep around if we don't mind having a few localized eslint-disables running around.

zarvox: Answering my own question: I think that's `no-misused-promises`; this is about creating variable bindings that are not used for anything else.

ebroder: Yeah that's right. Also, if you expand the context a bit, this rule configuration is specifically scoped to anything under `types/`, so won't apply to general code (in general, I think this is a good rule!)

PR #1249, opened 2022-12-28
Revert #404

### LINKED ISSUE #406: Reenable bootstrap-react Modal animations if upstream fixes things
#404 disabled animations to work around an upstream bug (details in #359).  If upstream fixes the issue, we can remove the `animation={false}` again.

ebroder:
We disabled animations for ModalForm due to poor interactions with autofocused inputs. That has since been fixed by upstream, so we can re-enable them.

I'll note that there's a comment saying that disabling the animations makes the UI feel snappier. I think the lack of animations feels a bit abrupt, but I don't have extraordinary strong opinions on the matter. I will accept either merging this PR or closing #406 as wontfix, but we should definitely do one or the other 🙂 

Fixes #406.

Before:

https://user-images.githubusercontent.com/28167/209875983-d1b40734-e5ad-4931-a9cb-6a70a1961dbd.mov

After:

https://user-images.githubusercontent.com/28167/209875993-40560e34-93fb-48c1-ae16-a835963fb86e.mov

### DIFF
--- a/imports/client/components/ModalForm.tsx
+++ b/imports/client/components/ModalForm.tsx
@@ -60,20 +60,8 @@ const ModalForm = React.forwardRef((
   const submitLabel = props.submitLabel ?? 'Save';
   const submitStyle = props.submitStyle ?? 'primary';
 
-  // Note: the animation=false is working around a regression in
-  // react-bootstrap that otherwise breaks autofocus:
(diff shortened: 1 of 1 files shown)


zarvox (approved):
Glad to see this got fixed!

PR #1255, opened 2022-12-31
Make RemoteMuteConfirmModal visible on first render

ebroder:
In order to avoid the overhead of the remote mute modal until we actually need it, we don't render it to the DOM until we're ready to display it. If it defaults to being hidden when it is first rendered, then you have to click on someone twice in order to mute them.


ebroder:
(Noticed this when I was copy-pasta-ing the logic for my new image upload modal)

zarvox (approved):

PR #1257, opened 2022-12-31
Use queries to narrow down the return type

ebroder:
If we make a query with a literal value, we can use that to infer something about the types of records that will be returned. This is especially useful for the Settings collection, where we often query on the tag of a tagged union.

I think the biggest argument against this change is that the runtime checks ensure that we aren't, e.g., getting back bogus data from the database. But I'd argue that we're in for a bad time if that's happening regardless of whether or not we have runtime checks in place to prevent it.

### DIFF
--- a/imports/lib/models/Base.ts
+++ b/imports/lib/models/Base.ts
@@ -20,6 +20,11 @@ export type FindOptions = FindOneOptions & {
   limit?: number;
 }
 
+type SelectorToResultType<T, Selector extends FindSelector<T>> =
+  Selector extends string ? T & { _id: Selector } :
+  Selector extends Mongo.ObjectID ? T & { _id: Selector } :
+  T & { [K in keyof Selector & keyof T]: Selector[K] extends T[K] ? Selector[K] : never };
+
 class Base<T extends BaseType> extends Mongo.Collection<T> {
   public name: string;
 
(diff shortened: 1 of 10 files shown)


ebroder:
(Also, we aren't doing runtime checks to ensure that the data we get back from the database matches the declared type of the collection in the first place, so this isn't really that much worse than just assuming that the data has _any_ structure at all)

imports/lib/models/Base.ts
```diff
@@ -20,6 +20,11 @@ export type FindOptions = FindOneOptions & {
   limit?: number;
 }
 
+type SelectorToResultType<T, Selector extends FindSelector<T>> =
+  Selector extends string ? T & { _id: Selector } :
+  Selector extends Mongo.ObjectID ? T & { _id: Selector } :
+  T & { [K in keyof Selector & keyof T]: Selector[K] extends T[K] ? Selector[K] : never };
```
zarvox, line 26: Ah, I think I get it -- the string/ObjectID cases seem straightforward, and for this case: if we see a key that matches a key known to be in the model type `T`, then we propagate that as a constraint, and if it's not there (like `$in` or `$and` or potentially `arrayName.0` or other things that are permitted in `Mongo.Selector<T>` but aren't keys of `T`) then that adds no additional constraint.

Clever!

zarvox (approved):

PR #1260, opened 2022-12-31
Ignore whitespace when looking for duplicate answers

ebroder:
We give a client-side warning when a user attempts to submit an answer that has already been submitted. However, this warning only looks for exact equality, meaning that we won't warn if the only difference is whitespace.

Since this is only a warning and not a block, it seems acceptable if this results in some false-positives, but it should have prevented duplicate answers that were ultimately rejected in hunts past.

### DIFF
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -878,7 +878,10 @@ const PuzzleGuessModal = React.forwardRef(({
   }, []);
 
   const onSubmitGuess = useCallback(() => {
-    const repeatGuess = guesses.find((g) => { return g.guess === guessInput; });
+    const strippedGuess = guessInput.replaceAll(/\s/, '');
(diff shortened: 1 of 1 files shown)


imports/client/components/PuzzlePage.tsx
```diff
@@ -878,7 +878,9 @@ const PuzzleGuessModal = React.forwardRef(({
   }, []);
 
   const onSubmitGuess = useCallback(() => {
-    const repeatGuess = guesses.find((g) => { return g.guess === guessInput; });
+    const repeatGuess = guesses.find((g) => {
+      return g.guess.replaceAll(/\s/, '') === guessInput.replaceAll(/\s/, '');
```
zarvox, line 882: nit: can you hoist `guessInput.replaceAll(/\s/, '')` outside the loop so we compute it just once instead of on every loop iteration?

zarvox (approved):

PR #1264, opened 2023-01-01
New helper for optional TypedMethod args

ebroder:
Match.Optional and Match.Maybe have slightly weird behavior when used in a pattern for an object. In an object, they match if the key is absent, but _not_ if the key is explicitly set to undefined.

This is mostly not a problem for our TypedMethods, because null and undefined values are stripped out by EJSON encoding. But that doesn't happen when we invoke a method from server-side code (such as when we use Ansible.log), which can cause spurious match failures.

This introduces a custom matcher which allows for explicit undefine and switches all TypedMethod's check methods to use it. This isn't strictly necessary (since it only matters for methods we want to invoke from the server with explicit undefined values), but is good for consistency and to encourage good copy-pasting.

I specifically ran into this with the "Reorganize Google Drive" button, since the implementation of that method logs something without passing an object as a second argument.

### DIFF
--- a/imports/methods/TypedMethod.ts
+++ b/imports/methods/TypedMethod.ts
@@ -1,4 +1,4 @@
-import { check } from 'meteor/check';
+import { check, Match } from 'meteor/check';
 import { EJSONable, EJSONableProperty } from 'meteor/ejson';
 import { Meteor } from 'meteor/meteor';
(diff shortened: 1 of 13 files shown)


ebroder:
Bah, OK, this fails in the opposite direction - absent keys aren't allowed anymore. Will figure out a different approach.

PR #1268, opened 2023-01-02
Fix update hunt with empty values

ebroder:


zarvox (approved):
This feels like the sort of thing that we could catch fairly easily if we added a bit of coverage to our testsuite

PR #1269, opened 2023-01-03
Guess modal improvements

ebroder:
A quick breakdown of what this changes:

- Increases the weight for Source Code Pro from 300 to 400 (i.e. "normal") across the board to give answer displays just a little more weight
- Switches away from displaying direction and confidence on the guess queue page as sliders:

  Screenshot 2023-01-02 at 10 29 04 PM
  
  This also lets us stick with the full-sized layout down to the "medium" screen size cutoff, instead of "large", so there's less no-man's-land in the middle sizes where the mobile layout looks huuge.
- Dramatically improves how the sliders in the guess modal are rendered. As it turns out, `FormControl type="range"` isn't really intended to work. Here's the before:

  Screenshot 2023-01-02 at 10 30 26 PM

  and after:

  Screenshot 2023-01-02 at 10 31 21 PM
- More or less completely redoes the table of historical guess submissions in the modal on the puzzle page, taking advantage of some of the design language we developed on the guess queue page:

  Screenshot 2023-01-02 at 10 32 53 PM

  and here's what it looks like at mobile sizes:

  Screenshot 2023-01-02 at 10 34 02 PM

### DIFF
--- a/client/stylesheets/app.scss
+++ b/client/stylesheets/app.scss
@@ -1,7 +1,7 @@
 // Standard font
 @import url("https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,400;0,700;1,400;1,700&display=swap");
 // Answer monospace font
-@import url('https://fonts.googleapis.com/css2?family=Source+Code+Pro:wght@300&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Source+Code+Pro:wght@400&display=swap');
 
 @import "theme";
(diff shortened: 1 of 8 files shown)


zarvox (approved):
Lovely cleanups, both in design, layout, and implementation structure with some nice reusable elements and style constants.

If I'm picking nits, I would ask that the horizontal size used by the two range inputs match, with their endpoints at the same horizontal positions, forming a vertical line, but honestly this is enough of an improvement that I'm motivated to get the rest of this landed first.

PR #1275, opened 2023-01-03
Bump json5 for security update

ebroder:


ebroder:
Oh, nevermind - dependabot got it

PR #1278, opened 2023-01-03
Store Google account IDs in addition to emails

ebroder:
In order to cross-reference Google Drive activity with Jolly Roger users, we need their account ID, not just their email address, so start capturing it going forward.

To avoid requiring everyone to re-link their Google accounts, introduce a new "configureCollectGoogleAccountIds" Meteor method. This takes advantage of the fact that anyone who has linked their Google account shows up in the "other contacts" API, and we can mine their account ID from there. This should be purely transitional - anyone setting up Jolly Roger in the future shouldn't need to do this - so it's not especially optimized for efficiency (many of the queries are poorly indexed) and there's no UI exposure.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -58,7 +58,7 @@
     ],
     "camelcase": ["error", {
       "properties": "never",
-      "allow": ["__meteor_runtime_config__", "script_v1", "drive_v3"]
+      "allow": ["__meteor_runtime_config__", "script_v1", "drive_v3", "people_v1"]
     }],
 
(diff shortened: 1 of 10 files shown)


zarvox (approved):
Okay, that's clever!  I presume you'll just call this manually from a browser console?

ebroder:
Yeah, that's the plan. Once that's done, I'll probably go poke around in the MongoDB cloud console to see if there are any that haven't been filled in, and investigate from there (but I have reasonably high hopes for coverage)

PR #1280, opened 2023-01-03
Fix guess queue flicker

grantaelliott:
Apply background color only to cells of the guess queue, instead of all descendants, to fix a flicker caused by animated button backgrounds. This also restores the expected background color of the button.

@aldeka raised this issue in the Discord.


ebroder (approved):
(We also need this change on the guess modal in `PuzzlePage.tsx`)

zarvox (approved):

PR #1282, opened 2023-01-03
Use direction/confidence components in notifications

ebroder:
Before:

Screenshot 2023-01-03 at 3 49 19 PM

and after:

Screenshot 2023-01-03 at 3 47 39 PM

I'm not totally in love with it, but I think it's at least as good as the current layout, and I like that it is slightly more compact and reuses what we've now made into a common design language.


zarvox (approved):
Approved per chat conversation

PR #1283, opened 2023-01-04
Use Drive Activity API for document activity

ebroder:
This replaces our old method of using the Drive list changes API to find changes to instead using the Drive Activity API, which exposes higher granularity data and lets us see who made an edit.

There are a few challenges to this API compared to the old implementation. Records are returned newest-to-oldest, so there's no way to ask for "records newer than what we've seen." Work around that by tracking the most recent event timestamp we've seen, and specifying a filter that the timestamp must be greater than that. (But include a 5 minute buffer just in case the API inserts records out of order.)

Since the API gives us user data, we add a new user field to DocumentActivity (and rework the unique index to include it). The API gives us back a series of Google account user IDs. However, there are various reasons that we might not be able to map that to a Jolly Roger user, so we allow the user field to be blank, indicating a user we weren't able to identify.

Finally, we rework the published activity bucket to include a count of document editors, instead of just a boolean.


imports/server/gdriveActivityFetcher.ts
```diff
@@ -0,0 +1,171 @@
...
+          return previousLatestTimestamp;
+        }
+
+        if (!activity.timestamp || !activity.targets || !activity.actors) {
+          return previousLatestTimestamp;
+        }
+
+        const ts = new Date(activity.timestamp);
+
+        const documentIds = [...activity.targets.reduce<Set<string>>((acc, target) => {
```
zarvox, line 87: It took me a couple reads through this before I realized that in all likelihood any given `activity` will have just 1 document id or just 1 actor id, and we're just handling both possible aggregations (and if the power set really happens, then we'll handle that correctly too).  I might add a comment here explaining that we mostly expect one or the other to be a single element, but this handles all the possibilities neatly.

imports/server/gdriveActivityFetcher.ts
```diff
@@ -0,0 +1,171 @@
...
+const FEATURE_FLAG_NAME = 'disable.gdrive_document_activity';
+
+async function featureFlagChanged() {
+  return new Promise<void>((r) => {
+    let handle: Meteor.LiveQueryHandle | undefined;
+    const cleanup = () => {
+      handle?.stop();
+      r();
+    };
+    handle = Flags.observeChanges(FEATURE_FLAG_NAME, cleanup);
```
zarvox, line 127: Doesn't `Flags.observeChanges` always call the callback with the current value during startup, either because of an `added` or synthesizing an event?  Seems like this will return ~immediately regardless of feature flag state.

imports/server/schemas/DriveActivityLatest.ts
```diff
@@ -0,0 +1,23 @@
+import * as t from 'io-ts';
+import { date } from 'io-ts-types';
+import { Overrides, buildSchema } from '../../lib/schemas/typedSchemas';
+
+const DriveActivityLatestCodec = t.type({
+  _id: t.string,
```
zarvox, line 6: Please add a comment that this is expected to be a singleton collection, and that the single known `_id` should be `default`.

imports/lib/schemas/DocumentActivity.ts
```diff
@@ -7,10 +7,11 @@ import { Overrides, buildSchema } from './typedSchemas';
    server, not by users */
 export const DocumentActivityCodec = t.type({
   _id: t.string,
-  ts: date, /* 5 minute granularity */
+  ts: date, /* rounded to ACTIVITY_GRANULARITY */
   hunt: t.string,
   puzzle: t.string,
   document: t.string,
+  user: t.union([t.undefined, t.string]),
```
zarvox, line 14: Maybe comment on this field is expected to be `undefined` if we don't recognize the google user ID, which can happen in normal operation.

imports/server/activityTracking.ts
```diff
@@ -56,8 +56,8 @@ class HuntActivityAggregator {
         const [bucket, newBucket] = this.lookupBucket(puzzle, ts);
         if (!bucket) return;
 
-        bucket.documentActivity = true;
-        bucket.allUsers.add('__document__');
+        bucket.documentUsers.add(fields.user ?? '__document__');
```
zarvox, line 59: nit: dedupe the `fields.user ?? '__document__'` on this line and the next

imports/server/gdriveActivityFetcher.ts
```diff
@@ -0,0 +1,171 @@
...
+    const document = await Documents.findOneAsync({ 'value.id': fileId });
+    if (!document) {
+      // eslint-disable-next-line no-continue
+      continue;
+    }
+
+    for await (const googleAccountId of googleAccountIds) {
+      // if we can't look up the account ID, we'll record as user=undefined
+      // which we'll count as a seprate user
+      const user = await MeteorUsers.findOneAsync({ googleAccountId, hunts: document.hunt });
```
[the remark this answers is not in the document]
ebroder: We can sort by `createdAt` so that we'll pick a consistent user, so long as they don't link/unlink their account (in which case I think anything goes, but that seems acceptable)

imports/server/gdriveActivityFetcher.ts
ebroder: Yeah, you're right - this structure doesn't work in the way I thought it did. I'll rework it to behave better.

ebroder:
Thanks for the feedback. I think I've gotten everything.

imports/server/gdriveActivityFetcher.ts
zarvox: I think this is still incorrect?  `featureFlagChanged()`, as written, will never block for more than about a database roundtrip so I expect that if you disabled the feature flag the `while(true)` below would turn into a busy loop and spin up your fans.  I believe that if you want to achieve the "promise blocks until database record changes" semantics, you need to prepare for having the `cleanup` callback not stop the watcher during the initial callback.

I'd suggest having an `let value: boolean | undefined;` outside of the loop, and then in the callback, when value is undefined, you assign it the value of the callback argument (currently unbound), and when value is not undefined, if it changed, you stop the handle and resolve the promise.

ebroder: Yeah, you're right, I had forgotten that I made `observeChanges` emit on initial query to match the behavior of `Cursor.observeChanges`. I did something slightly different, but I think it should work as desired now? (I flipped the flag back and forth a few times to verify that it did not busy-loop)

zarvox: Hmmm, this works now, but I'm a little worried about what happens once we make a future change to make `Flags.active()` -> `await Flags.activeAsync()` since then `initializing` would potentially be set to false again before the async function in the initial `added()` completes.

What you have here won't block the way you want it to once the `Flags.active` call in the `observeChanges` is made non-Fiber-blocking, unless upstream also makes `observeChanges` await each of the callbacks it issues.  I guess we'll notice since we'd have to void the promise, but I worry we'd have new bugs from the ordering-barrier going missing.

I'm not sure how upstream will attempt to handle this, but it seems subtle and like something we can avoid entirely by setting initializing to false in the first triggered callback instead of after the observeChanges.

Possible we'll need to carefully review our usage of observeChanges more broadly as this situation evolves.

zarvox (approved):
Approving, since I want to get this landed, but with a cautionary note on how we use `observeChanges`

imports/server/gdriveActivityFetcher.ts
ebroder: Hmm. Yeah, I think our strategy here mostly depends on what happens to the behavior of `observeChanges` when they actually switch to an async implementation; there are a few different ways it could go.

PR #1287, opened 2023-01-04
Fix backwards logic on updateProfile

ebroder:


ebroder:
I'm going to force merge this as it's occurring in production and blocked at least one person from updating their profile.

PR #1291, opened 2023-01-05
Handle websockets in cluster children processes

ebroder:
In cluster mode, the coordinator process sends incoming websocket connections to a worker process via child_process.send. However, when we ported this code away from meteorhacks:cluster, we didn't add a listener in the child to handle that message. This meant that if we had clustering enabled, we would just silently drop websocket connections. This didn't _break_ things, because we would fall back on XHR, but it's definitely not great.

Bring back a handler for the IPC message in the child.


zarvox (approved):
Thanks for tracking this down!

PR #1293, opened 2023-01-05
Introduce `TypedPublication`

ebroder:
I started sketching this out initially because I was angry about the lack of error reporting for publish methods, and this seemed like an easier way to address that than trying to hook into the Meteor guts somewhere. Opening this as a draft for now, but @zarvox curious to get your thoughts.

`TypedPublication` is a lot like `TypedMethod`, but, as you might expect, for publications.

Unlike methods, publications don't need to have a name. If you publish with a name of `null`, the publication is automatically pushed to clients. This is supported through a `DefaultTypedPublication`, which requires that the publication take no arguments.

We could theoretically continue to use `Meteor.publish` directly for those, but `TypedPublication` also introduces error handling, which seemed useful for all publications, not just ones that could be explicitly subscribed. As it turns out, Meteor by default more or less completely swallows any exceptions thrown by a publish function. (Specifically, it cancels the subscription and sends the exception to the client in a "nosub" DDP message, which the client completely ignores)


imports/lib/publications/pendingAnnouncementsForSelf.ts
```diff
@@ -0,0 +1,5 @@
+import TypedPublication from './TypedPublication';
+
+export default new TypedPublication<void>(
+  'PendingAnnouncements.publications.forSelf'
```
zarvox, line 4: convention bikeshedding: I'm not sure we need to put `.publications` in here?  These are already `TypedPublication` and the context these strings will pop up in will be "a publication name".  Was there a reason you added the additional layer here, or a pattern of use you were imagining?

ebroder: Possibly not. I did the same thing in `TypedMethod`, and was basically riffing off of the style that [mdg:validated-method seems to recommend](https://github.com/meteor/validated-method#using-simpleschema)

Honestly, naming in general is a little more annoying here than it was with `TypedMethod`, because in most cases the thing you're subscribing to is a noun, and those nouns already exist as names. With `TypedMethod` at least it's usually a noun+verb. (You can see in these examples how I have to bend over a bit to avoid naming things in ways that conflict with other files - e.g. this isn't just `pendingAnnouncements`, even though that's what we called the old subscription)

ebroder:
As a general rule of thumb, I really like smaller files - I find it generally much easier to navigate between files than within a single large file. I do think we can pretty plausibly move away from the flat bags style of organization in favor of grouping by...noun, probably? (E.g. we could have `server/guesses/{schemas/,models/,methods/,publications/}`)

imports/lib/publications/TypedPublication.ts
```diff
@@ -0,0 +1,142 @@
...
+        },
+        onReady: () => {
+          resolve(handle);
+        },
+      });
+    });
+  }
+
+  useSubscribe<T>(...args: TypedMethodSubscribeArgs<T, Args>) {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
zarvox, line 134: I'm curious what makes this rule fire and if it makes sense to try to abide by its constraints or not.  Is this just the rule failing to infer this as something that should only be called where a hook can be called?

ebroder: If I'm reading this right, it trips any time you call a function name that looks like a hook inside of any class method: https://github.com/facebook/react/blob/main/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js#L509-L519

zarvox: Ahhh.  What do you think about shaping this as a standalone hook function that you pass the publication into instead?  That might be more in line with how React wants you to use hooks for testability, even though we don't really exercise that much?

Something like (probably needs type parameters worked out):

```typescript
function useTypedSubscribe<T>(publication: TypedPublication<T>, ...args: TypedMethodSubscribeArgs<T, Args>) {
  useSubscribe(pub.name, ...args);
}
```

I'm also fine with this as it is, just exploring the design space a little.

ebroder: Yeah, that seems like a good pattern. I'll switch to that.

ebroder: I think that also gives us a slightly nicer path forward on conditional subscribes - we can just make the publication an optional parameter.

zarvox: Yeah that turned out well -- winds up feeling very similar to the existing `useSubscribe` at the callsite.

imports/server/publications/pendingGuessesForSelf.ts
```diff
@@ -83,23 +84,14 @@ class PendingGuessWatcher {
   }
 }
 
-// Publish pending guesses enriched with puzzle and hunt. This is a dedicated
```
zarvox, line 86: This comment seems like helpful context for understanding the publication; is there a reason you're removing it?

zarvox (approved):
Not sure if you were still iterating on this (the PR is still a draft), but personally I think it looks great.

imports/server/publications/pendingGuessesForSelf.ts
ebroder: Huh, no it must have gotten dropped by accident. I'll add it back, although my inclination is to put high-level commentary about how to interact with a publication on the declaration, rather than the definition.

ebroder:
Hmm. I had sort of wanted to test `typedSubscribe.async`, but I'm also not sure I want to untangle `imports/servers/users.ts` enough or start to convert the default `mongo.*` subscription, which are the two things that I'm currently using async subscriptions on. So sure, yeah, let's go ahead and get this ready for landing.

zarvox (approved):
Looks like this is needing a rebase due to my recent changes to NotificationCenter, but otherwise :shipit:

PR #1294, opened 2023-01-05
Fix Discord avatar vertical alignment

grantaelliott:
Fix a minor error in #691 that causes a 1px vertical misalignment of avatars.


ebroder (commented):
Can you double-check the `/rtcdebug` page before merging this? That looks to be the only place we use `Avatar` with `inline=true`.

grantaelliott:
I wasn't expecting anything to change since this only affects the inner content, but the vertical centering in rtcdebug gets thrown off by the fix. Good catch.

Added classNames to Avatar for future styling flexibility. Solved the rtcdebug case just by adding a vertical-align.

ebroder (approved):

zarvox (approved):

PR #1298, opened 2023-01-06
Fix validation for server-side method invocations

### LINKED ISSUE #1297: Difference in behavior between client and server with optional arguments
We use `Match.Optional` for optional arguments to a `TypedMethod`. However, the behavior of `Match.Optional` is actually slightly subtle - when used in an object, `Match.Optional(T)` (and `Match.Maybe(T)`) means that a key can be absent, but if it is present, it must match `T`. Or in other words (from Meteor's documentation):

```
const pattern = { name: Match.Maybe(String) };
 
check({ name: 'something' }, pattern); // ✅
check({}, pattern); // ✅
check({ name: undefined }, pattern); // ❌
check({ name: null }, pattern); // ❌
```

That's a somewhat unfortunate position for Meteor to take, as TypeScript historically does not recognize a distinction between an absent key and a key with value `undefined`:

Screenshot 2023-01-05 at 6 22 19 PM

(As an aside, there is a newer configuration flag - `exactOptionalPropertyTypes` - which draws this distinction. And. I did come up with [some fun shenanigans](https://github.com/DefinitelyTyped/DefinitelyTyped/compare/master...ebroder:DefinitelyTyped:meteor-optional-matches) to take advantage of that in the type inferences of `check`. However, enabling it would be a significant migration, as we pretty frequently declare properties to be optional and then pass undefined - especially in React components).

In any case, Meteor is documented to work in this way because of the behavior of `EJSON`, which is used to serialize arguments to DDP messages and which drops keys with a value of `undefined` during serialization. This makes the distinction more or less irrelevant in practice...except when we invoke Meteor methods from server-side code. When we do that, the arguments aren't round-tripped through `EJSON`, which means that `undefined` values are still present. This can causes us to fail `check` calls.

I've observed this most notably in `configureOrganizeGoogleDrive`, which calls `Ansible.log` with no `obj` argument. That method call in turn complains because `obj` is passed through as `undefined`.

I think an easy short-term fix would be to modify `TypedMethod.call` and `TypedMethod.callPromise` to run the provided arguments through `EJSON.clone` if they are called in server-side code, to ensure that all invocations of methods get the same round-tripping.

ebroder:
When invoked from the client, method arguments are encoded with EJSON before being sent over DDP. This has the specific side-effect of stripping any key-value pairs where the value is undefined.

This doesn't happen when methods are invoked on the server. That interacts poorly with `Match.Optional`, which permits keys to be absent but does _not_ permit them to be `undefined`.

To make the behavior consistent, round-trip arguments through EJSON with server-side invocations.

Fixes #1297.

### DIFF
--- a/imports/methods/TypedMethod.ts
+++ b/imports/methods/TypedMethod.ts
@@ -134,10 +134,29 @@ export default class TypedMethod<
   }
 
   call<T>(...args: TypedMethodCallArgs<T, Args, Return>): void {
-    return Meteor.call(this.name, ...args);
+    // If we invoke a function on the client, the arguments get encoded to EJSON
+    // before being sent over DDP. This has the effect (among other things) of
+    // stripping out any key-value pairs where the value is undefined. This
+    // doesn't happen on the server, which can cause bad interactions with
+    // Match.Optional, which permits keys to be absent but does _not_ permit the
+    // value to be undefined
+    //
+    // Round-tripping through EJSON on the server ensures that we get the same
+    // behavior
+    const cleaned = args.map((a) => {
+      return Meteor.isServer &&
+        typeof a === 'object' ? EJSON.parse(EJSON.stringify(a)) :
+        a;
+    });
+    return Meteor.call(this.name, ...cleaned);
   }
 
   callPromise<T>(...args: TypedMethodCallPromiseArgs<T, Args>): Promise<Return> {
-    return Meteor.callAsync(this.name, ...args);
+    const cleaned = args.map((a) => {
+      return Meteor.isServer &&
+        typeof a === 'object' ? EJSON.parse(EJSON.stringify(a)) :
+        a;
+    });
+    return Meteor.callAsync(this.name, ...cleaned);
   }
 }


zarvox:
Hmmm.  This is more correct at the margins, but I think I'd be more interested in seeing us refactor things to avoid calling Meteor methods from the server in the first place, in favor of just using normal function calls.  I think that the isomorphic code ship has largely sailed, and to the extent that we want to make code usable in both contexts, we should extract the bits that do I/O, or simply stick to interfaces that actually behave equivalently.

Indeed, a quick search for `.call(` reveals that `Ansible.log` is the only place that we make use of server-originated Meteor method calls, and (correct me if I'm wrong) `Ansible.log` seems pretty trivial to replace with a regular function call?  If we need to pull the current `userId` (if any exists) from implicit method or publication context, we can always pull it out of `DDP._CurrentMethodInvocation.get()` or `DDP._CurrentPublicationInvocation.get()`, or if you want to use only public API, use

```typescript
let userId = undefined;
try {
  userId = Meteor.userId();
} catch (e) {
  // not in a method call or publication
}
```

Then we can just avoid defining `.call()` and `.callAsync()` on the server, precluding the problems described above.

Another reason to avoid server-to-server Meteor method calls is that last I checked, by default they blocked in fibers on the serverside at the callsite if you don't pass an async callback.  If we're going to do different things on client and server, I'd rather sidestep the whole Meteor serialization/method blocking/queuing/separate error reporting complexity and just call an implementation function directly on the serverside if we don't have a real use case for threading all that through.

ebroder:
Yeah, OK, I'm mostly convinced here, although it's sent me down a rabbit hole of reworking how we do logging everywhere. I'll push up some new work once I have it ready (hopefully soon-ish), but I think this branch is dead.

PR #1300, opened 2023-01-07
Fix non-operator mobile view of PuzzleListPage

zarvox:
PR #1289 introduced a regression in the view here, primarily by making the puzzle title column into an `inline` span rather than an `inline-block` div.  We can achieve the intended effect by placing the Breakable attributes on the puzzle title column element.

Before:

Screenshot 2023-01-06 at 9 41 58 PM

After:

Screenshot 2023-01-06 at 9 41 41 PM


imports/client/components/Puzzle.tsx
```diff
@@ -68,8 +67,9 @@ const StyledButton = styled(Button)`
   }
 `;
 
-const PuzzleTitleColumn = styled(Breakable)`
+const PuzzleTitleColumn = styled(PuzzleColumn)`
   flex: 4;
+  overflow-wrap: word;
```
grantaelliott, line 72:
```suggestion
  overflow-wrap: break-word;
```
I think this is just a typo. With that change, I believe it works. Thanks for catching this.

grantaelliott (changes requested):

imports/client/components/Puzzle.tsx
zarvox: Ah, yeah, I misread `Breakable.tsx`.  Thanks for catching my wrong-fix!

grantaelliott (approved):

PR #1304, opened 2023-01-07
Fix exception in Bugsnag collection

ebroder:
On server-side code, it's not sufficient to be in a Fiber for Meteor.user to be legal to call.

### DIFF
--- a/imports/server/bugsnag.ts
+++ b/imports/server/bugsnag.ts
@@ -1,7 +1,7 @@
 import { Meteor } from 'meteor/meteor';
 import Bugsnag from '@bugsnag/js';
-import Fiber from 'fibers';
 import isAdmin from '../lib/isAdmin';
+import MeteorUsers from '../lib/models/MeteorUsers';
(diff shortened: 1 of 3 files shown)


imports/server/bugsnag.ts
```diff
@@ -22,7 +22,12 @@ if (apiKey) {
     ],
     onError: (event) => {
       if (Fiber.current) {
```
zarvox, line 24: If we try/catch the `Meteor.user()` call, is there any point in the `Fiber.current` check?

imports/server/bugsnag.ts
```diff
@@ -22,7 +22,12 @@ if (apiKey) {
...
-        const user = Meteor.user();
+        let user;
+        try {
+          user = Meteor.user();
```
zarvox, line 27: Should this be something like `await MeteorUsers.findOne(Meteor.userId()).fetchAsync()` since we're in `server` code?

ebroder:
Yeah, I think correct on both counts. Please take another look.

zarvox (approved):

PR #1305, opened 2023-01-07
Fix server-side source code capture for Bugsnag

ebroder:
If you supply a projectRoot, Bugsnag will find source code in that directory and include it as part of the error report. This doesn't work out of the box in production because we don't have the non-compiled source anymore. However, we do have it in the form of sourcemaps (which contain the full uncompiled source code). So use the source maps to materialize a temporary directory to use as a projectRoot for Bugsnag.

(We can't get Bugsnag to use the sourcemaps because Meteor installs source-map-support, which rewrites the stack traces before Bugsnag gets a chance to see them, so Bugsnag can't map the post-rewrite stack frame onto the source map, which, since it refers to the post-compilation filename, doesn't match)

Note: The next thing I'm going to look into is source map uploads for client-side code. In theory, Bugsnag should fetch this without any intervention, since the URL of the JS source file in stack traces is publicly accessible, and Meteor serves up an `X-SourceMap` header for that file, but in practice it doesn't seem to be working: https://app.bugsnag.com/death-and-mayhem/jolly-roger/errors/63b942f468b4b5000861d06f. So the easiest thing is to just upload them ourselves


ebroder:
(You can see this working in the latest event for https://app.bugsnag.com/death-and-mayhem/jolly-roger-dev/errors/63b992526e7df30008088195)

zarvox (approved):

PR #1308, opened 2023-01-08
Fix server-side source code capture for Bugsnag (Take 2)

ebroder:
If you supply a projectRoot, Bugsnag will find source code in that directory and include it as part of the error report. This doesn't work out of the box in production because we don't have the non-compiled source anymore. However, we do have it in the form of sourcemaps (which contain the full uncompiled source code). So use the source maps to materialize a temporary directory to use as a projectRoot for Bugsnag.

(We can't get Bugsnag to use the sourcemaps because Meteor installs source-map-support, which rewrites the stack traces before Bugsnag gets a chance to see them, so Bugsnag can't map the post-rewrite stack frame onto the source map, which, since it refers to the post-compilation filename, doesn't match)

Unlike the first attempt in 7107fb74, this drops the source-map repository, which uses wasm for faster parsing of mappings data. That isn't compatible with Meteor's use of fibers, but we don't need it, so it doesn't matter.


ebroder:
I'm not sure how to get Github to show it, but here's a comparison between this and #1305:

```
evan@mathias jolly-roger (evan/server-source-maps) % interdiff <(git show 7107fb74) <(git show 43db0a89)
diff -u b/imports/server/bugsnag.ts b/imports/server/bugsnag.ts
--- b/imports/server/bugsnag.ts
+++ b/imports/server/bugsnag.ts
@@ -7,13 +7,17 @@
 import Bugsnag from '@bugsnag/js';
 import Fiber from 'fibers';
 import glob from 'glob';
-import { SourceMapConsumer } from 'source-map';
 import Ansible from '../Ansible';
 import isAdmin from '../lib/isAdmin';
 import { userIsOperatorForAnyHunt } from '../lib/permission_stubs';
 import addRuntimeConfig from './addRuntimeConfig';
 import onExit from './onExit';

+interface SourceMap {
+  sources: string[];
+  sourcesContent: string[];
+}
+
 const apiKey = process.env.BUGSNAG_API_KEY;

 if (apiKey) {
@@ -73,12 +77,12 @@
     // server-side code bundle
     const sourceMapPaths = await promisify(glob)('**/*.map', { cwd: process.cwd() });
     await Promise.all(sourceMapPaths.map(async (sourceMapPath) => {
-      const sourceMapConsumer = await new SourceMapConsumer(await fs.readFile(sourceMapPath, 'utf8'));
-      await Promise.all(sourceMapConsumer.sources.map(async (sourcePath) => {
+      const sourceMap = JSON.parse(await fs.readFile(sourceMapPath, 'utf8')) as SourceMap;
+      await Promise.all(sourceMap.sources.map(async (sourcePath, i) => {
         if (!sourcePath.startsWith('meteor://💻app/')) {
           return;
         }
-        const source = sourceMapConsumer.sourceContentFor(sourcePath);
+        const source = sourceMap.sourcesContent[i];
         if (!source) {
           return;
         }
diff -u b/package-lock.json b/package-lock.json
--- b/package-lock.json
+++ b/package-lock.json
@@ -10860,11 +10860,6 @@
         "smart-buffer": "^4.2.0"
       }
     },
-    "source-map": {
-      "version": "0.7.4",
-      "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.7.4.tgz",
-      "integrity": "sha512-l3BikUxvPOcn5E74dZiq5BGsTb5yEwhaTSzccU6t4sDOH8NWJCstKO5QT2CvtFoK6F0saL7p9xHAqHOlCPJygA=="
-    },
     "source-map-js": {
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/source-map-js/-/source-map-js-1.0.2.tgz",
diff -u b/package.json b/package.json
--- b/package.json
+++ b/package.json
@@ -67,7 +67,6 @@
     "react-split-pane": "^0.1.92",
     "react-textarea-autosize": "^8.4.0",
     "simpl-schema": "^3.2.0",
-    "source-map": "^0.7.4",
     "stacktrace-js": "^2.0.2",
     "styled-components": "^5.3.6",
     "use-persisted-state": "^0.3.3"
```

zarvox (approved):
Looks good but needs a rebase

PR #1309, opened 2023-01-08
Upload client source maps to Bugsnag

ebroder:
In theory, Bugsnag should be able to fetch these on its own, but in practice it doesn't seem to be happening.

I think there may still be followup work for dynamically imported files - they ship source maps in a different format. Specifically, for the main JS bundle, Meteor uses the `sourceMap` and `sourceMapUrl` properties in the manifest to generate `X-SourceMap` headers. For dynamically imported files, the contents of the files include a `sourceMappingURL` comment, which refers to a file that's served independently. I tried to write some code to upload those as well, but couldn't get it working in the amount of time I had allotted myself to work on this today. I think this is still a useful incremental improvement, though, and worth landing in the interim.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -46,6 +46,8 @@
 
     // sometimes else after a return reads more clearly
     "no-else-return": "off",
+    // same for continue
+    "no-continue": "off",
 
      // We sometimes use these in various forms
     "no-underscore-dangle": ["error",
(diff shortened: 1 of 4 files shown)


zarvox (approved):

ebroder:
(I'll land this tonight when I'm back at my computer, just out of an abundance of caution, although this should be safer than the server-side handling since it doesn't block startup)

PR #1310, opened 2023-01-09
Chat v2: @-mentions, formatting preview

### LINKED ISSUE #337: Add chat notifications based on username mentions

zarvox:
This PR implements a more complicated text editor using [Slate](https://docs.slatejs.org/).  Its two primary feature improvements are:

* support for @-mentions
* Discord-inspired inline preview of markdown formatting, powered by the `marked` lexer

To achieve this, we need to evolve our chat message input format from a single flat string in a textarea to a more structured representation.  Since both client and server need to understand the meaning of the structured representation (the client must author and display messages; the server must process @-mentions, dingwords, and propagate messages to Discord), it is convenient to store the structured form in the database, and to use it in our types.  I was unable to convince the typedSchemas-to-simpl-schema glue to work with the full structure, but was still able to define types while declaring the object opaque.

This PR is easiest to understand and review in commit order, but you will probably need to read some of the Slate docs to really grasp what is happening and why some things are structured the way they are.

I recommend reading:

* [Concepts: Nodes](https://docs.slatejs.org/concepts/02-nodes)
* [Concepts: Locations](https://docs.slatejs.org/concepts/03-locations)
* [Rendering: Decorations](https://docs.slatejs.org/concepts/09-rendering#decorations) since they are how we do all the markdown inline previews (and avoid having to keep a bunch of stateful block-level logic around)
* [Normalizing](https://docs.slatejs.org/concepts/11-normalizing), especially the Built-in Constraints

before reading through 6244d965c38e2585ddb2d9df12cfc8201719d7dc.  After that, I recommend reading through the changeset in commit order -- starting with the editor and the schemas.  The rest is pretty much what you'd expect thereafter.

It was easiest to incrementally implement the new message support alongside the old one.  There's actually a spot on PuzzlePage where I now have a boolean that determines which input system to use, and all the renderers and backend implementations and hooks and whatnot support both the old and the new format.  If we wanted, we could add a circuit breaker/feature flag there to control how new messages are authored or transmitted. 

Lastly, a few comments on editor behavior:

* I explicitly intend to support bold, italics, underline, strikethrough, code spans, code blocks, and auto-linking URLs
* I aim to **not** support ordered lists (because the numbering is deeply surprising and confusing to people), unordered lists (for consistency with Discord, and because they're not *that* much of an improvement if you already have soft-breaks), the `[link text](destination)` link format, or anything that involves injecting HTML, table extensions, image extensions, etc.
* I aim to preserve soft-breaks (basically: have all input respect `white-space: pre-wrap`), because most humans expect a newline to cause their sent chat message to have a newline
  * Due to inconsistencies in how `marked` handles newlines, I occasionally add or remove newlines when rendering to attempt to avoid certain surprising behavior.
  * To simplify how we connect the editor to the `marked` lexer (and to avoid statefully reimplementing the entire markdown spec in our editor & types), the message format is mostly flat, and we normalize to having a single MessageElement.

Fixes #337.


imports/client/components/FancyEditor.tsx
```diff
@@ -0,0 +1,639 @@
...
+    onSelected(user);
+  }, [onSelected, user]);
+  return (
+    <div
+      key={user._id}
+      style={{
+        padding: '1px 3px',
+        borderRadius: '3px',
+        background: selected ? '#e0ecfc' : 'transparent',
+      }}
```
ebroder, line 179: Should this be a styled component?

imports/client/components/FancyEditor.tsx
```diff
@@ -0,0 +1,639 @@
...
+    const displayName = u.displayName?.toLowerCase() ?? '';
+    const foundDisplayName = displayName.includes(needle);
+    const startsDisplayName = displayName.startsWith(needle);
+    const emails = u.emails?.map((addrObj) => addrObj.address.toLowerCase()) ?? [];
+    const foundEmail = emails.some((addr) => addr.includes(needle));
+    const startsEmail = emails.some((addr) => addr.startsWith(needle));
+    const googEmail = u.googleAccount?.toLowerCase() ?? '';
+    const foundGoogle = googEmail.includes(needle);
+    const startsGoogle = googEmail.startsWith(needle);
+    const discordName = u.discordAccount ? `${u.discordAccount.username.toLowerCase()}#${u.discordAccount.discriminator}` : '';
```
ebroder, line 219: Just as a note I opened #1312 as vaguely related to this, though not worth blocking over

imports/client/components/FancyEditor.tsx
```diff
@@ -0,0 +1,639 @@
...
+  });
+
+  // The floating autocomplete box for @-mentions
+  const ref = useRef<HTMLDivElement>(null);
+  // selection target?
+  const [target, setTarget] = useState<Range | undefined>(undefined);
+  // Offset into the list of potentially-matching usernames
+  const [index, setIndex] = useState<number>(0);
+  // current needle for search in display names
+  const [search, setSearch] = useState('');
```
ebroder, line 429: should these have names that incorporate "autocompletion" or something?

imports/lib/schemas/ChatMessage.ts
```diff
@@ -4,12 +4,47 @@ import { BaseCodec, BaseOverrides } from './Base';
...
   hunt: t.string,
   // The puzzle to which this chat was sent.
   puzzle: t.string,
   // The message body. Plain text.
-  text: t.string,
+  text: t.union([t.string, t.undefined]),
+
+  // The message contents, as an opaque object which has the
+  // schema of Content above
+  content: t.union([t.UnknownRecord, t.undefined]),
```
ebroder, line 46: If we're going the opaque object route, it should be possible to export a `ChatMessageType` that uses `ChatMesageContentType` so that we don't actually have to cast.

imports/lib/schemas/ChatNotification.ts
```diff
@@ -17,8 +17,10 @@ const ChatNotificationFields = t.type({
   puzzle: t.string,
   // The hunt in which the puzzle resides.
   hunt: t.string,
-  // The message body. Plain text.
-  text: t.string,
+  // The message body, if chat message v1. Plain text.
+  text: t.union([t.string, t.undefined]),
+  // The message content, if chat message v2.  Actually contains ChatMessageContentType
+  content: t.union([t.UnknownRecord, t.undefined]),
```
ebroder, line 23: (Similarly here, I'd prefer if we can make `ChatNotificationType` use the actual `ChatMessageContentType`)

imports/client/components/FancyEditor.tsx
```diff
@@ -0,0 +1,639 @@
...
+  tokens.forEach((token) => {
+    // mutual recursion requires this
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    walkToken(token, callback, end);
+    end += token.raw.length;
+  });
+};
+
+const walkToken = (token: marked.Token, callback: TraverseCallback, offset: number) => {
+  if ((token as any).tokens) {
```
ebroder, line 291: FTR, I think you can phrase this as `'tokens' in token && token.tokens` to avoid the `any` cast

imports/client/components/ChatMessageV2.tsx
```diff
@@ -0,0 +1,127 @@
...
+  border-radius: 4px;
+  padding: 4px;
+  width: 100%;
+  background-color: #eee;
+  color: black;
+  margin-bottom: 0;
+`;
+
+function isMention(thing: ChatMessageContentNodeType): thing is ChatMessageMentionNodeType {
+  return (thing as any).type === 'mention';
```
ebroder, line 38: (similarly this could phrase as `'type' in thing && thing.type === 'mention'`

imports/server/methods/sendChatMessageV2.ts
```diff
@@ -0,0 +1,45 @@
...
+    return arg;
+  },
+
+  async run({ puzzleId, content }: { puzzleId: string, content: string }) {
+    check(this.userId, String);
+    const contentObj = JSON.parse(content);
+    check(contentObj, {
+      type: String,
+      children: [Match.OneOf({
+        type: String,
```
ebroder, line 23: Can we just check for a literal `'mention'` here instead of scanning below?

imports/server/methods/sendChatMessageV2.ts
```diff
@@ -0,0 +1,45 @@
...
+      if (type !== undefined && type !== 'mention') {
+        throw new Meteor.Error(400, 'Non-text content child must have type "mention"');
+      }
+    });
+
+    await sendChatMessageInternalV2({
+      puzzleId,
+      content: contentObj as ChatMessageContentType,
+      sender: this.userId,
+    });
```
ebroder, line 43: Specifically, I think we could write this like so:

```suggestion
    const contentObj: unknown = JSON.parse(content);
    check(contentObj, {
      type: 'message' as const,
      children: [Match.OneOf({
        type: 'mention' as const,
        userId: String,
      }, {
        text: String,
      })],
    });

    await sendChatMessageInternalV2({
      puzzleId,
      content: contentObj,
      sender: this.userId,
    });
```

imports/client/components/PuzzlePage.tsx
```diff
@@ -99,7 +103,7 @@ const DEBUG_SHOW_CALL_STATE = false;
 
 const tabId = Random.id();
 
-const FilteredChatFields: ('_id' | 'puzzle' | 'text' | 'sender' | 'timestamp')[] = ['_id', 'puzzle', 'text', 'sender', 'timestamp'];
+const FilteredChatFields: ('_id' | 'puzzle' | 'text' | 'content' | 'sender' | 'timestamp')[] = ['_id', 'puzzle', 'text', 'content', 'sender', 'timestamp'];
 type FilteredChatMessageType = Pick<ChatMessageType, typeof FilteredChatFields[0]>
```
ebroder, line 107: I know you didn't write it like this, but I think this would be better phrased as

```suggestion
const FilteredChatFields = ['_id', 'puzzle', 'text', 'content', 'sender', 'timestamp'] as const;
type FilteredChatMessageType = Pick<ChatMessageType, typeof FilteredChatFields[number]>
```

ebroder (commented):
OK, overall this looks pretty good. I think I'm most interested in seeing some of the typing improvements I left comments about, especially the type we use for `ChatMessageType` and `ChatNotificationType`, but in general anything that lets us do less casting (since that's a good way to let unintended bugs sneak in)

imports/server/methods/sendChatMessageV2.ts
zarvox: The docs for `check` do not suggest they support value literals.  I guess I could use a `Match.Where(condition)` instead?

imports/client/components/FancyEditor.tsx
zarvox: Yes.  Addressed in 398e9b23095f36dab44b10bd1a73983d5eed48ff when I made autocomplete show avatars

imports/server/methods/sendChatMessageV2.ts
ebroder: Huh. I...am pretty sure I didn't make that up, but I agree it's not documented behavior. We've certainly relied on it in other places we've used `Match.OneOf`, though, and the code seems to explicitly intend to support it: https://github.com/meteor/meteor/blob/9cc480a1ab3bfc85ebc86a86a85bfa92a19103b9/packages/check/match.js#L240-L250

```
> const pat = { type: 'mention' };
> check({ type: 'mention' }, pat)
> check({ type: 'not-a-mention' }, pat);
Uncaught:
errorClass [Error]: Match error: Expected mention, got "not-a-mention" in field type
```

imports/client/components/FancyEditor.tsx
zarvox: Sure, these can be described a bit better.  Addressed in 62f0a8803e341e77b4579a8155455b2fb566b657

imports/server/methods/sendChatMessageV2.ts
zarvox: Oh, the docs don't say it, but it works, and if I throw in an `as const` it even gets correctly inferred as being compatible with `ChatMessageContentType`.  Fixed in afc4e4e78da6f8c07dd9e0716cbeecbe68d1f72e

imports/lib/schemas/ChatNotification.ts
zarvox: Fixed in afc4e4e78da6f8c07dd9e0716cbeecbe68d1f72e

imports/server/methods/sendChatMessageV2.ts
zarvox: Ah, I should have read all the comments before editing, but I appear to have wound up with the same implementation as you have here in afc4e4e78da6f8c07dd9e0716cbeecbe68d1f72e

imports/client/components/ChatMessageV2.tsx
zarvox: This can just use the shared function from the ChatMessage schema file.  Fixed in 7cbd867e0b78b131bd0536cd99367ba791475764

imports/client/components/FancyEditor.tsx
zarvox: Ah, neat. Fixed in 7cbd867e0b78b131bd0536cd99367ba791475764

imports/client/components/PuzzlePage.tsx
zarvox: Fixed in c22660e87c59e080e44c30f43aaa30059a093c01.  `as const` is magical.

imports/client/components/FancyEditor.tsx
zarvox: Yeah, that seems both reasonable and largely orthogonal to this work.

zarvox:
> One thing I'll note is that rendering in the editor of mid-composition multi-line code blocks is a little odd. Here's what I saw while I was composing:
> 
> Screenshot 2023-01-09 at 2 02 07 PM
> 
> Screenshot 2023-01-09 at 2 02 11 PM
> 
> Screenshot 2023-01-09 at 2 02 19 PM
> 
> Screenshot 2023-01-09 at 2 02 22 PM
> 
> I assume this is the behavior of marked's lexer and not something that's easily under our control?

Yeah.  Check out the token series from:

* [first](https://marked.js.org/demo/?outputType=lexer&text=here%27s%20multiline%20code%3A%0A%60%60%60&options=%7B%0A%20%22async%22%3A%20false%2C%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22extensions%22%3A%20null%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tokenizer%22%3A%20null%2C%0A%20%22walkTokens%22%3A%20null%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=master)
* [second](https://marked.js.org/demo/?outputType=lexer&text=here%27s%20multiline%20code%3A%0A%60%60%60%0A&options=%7B%0A%20%22async%22%3A%20false%2C%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22extensions%22%3A%20null%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tokenizer%22%3A%20null%2C%0A%20%22walkTokens%22%3A%20null%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=master)
* [third](https://marked.js.org/demo/?outputType=lexer&text=here%27s%20multiline%20code%3A%0A%60%60%60%0Aabc&options=%7B%0A%20%22async%22%3A%20false%2C%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22extensions%22%3A%20null%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tokenizer%22%3A%20null%2C%0A%20%22walkTokens%22%3A%20null%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=master)
* [fourth](https://marked.js.org/demo/?outputType=lexer&text=here%27s%20multiline%20code%3A%0A%60%60%60%0Aabc%0A&options=%7B%0A%20%22async%22%3A%20false%2C%0A%20%22baseUrl%22%3A%20null%2C%0A%20%22breaks%22%3A%20false%2C%0A%20%22extensions%22%3A%20null%2C%0A%20%22gfm%22%3A%20true%2C%0A%20%22headerIds%22%3A%20true%2C%0A%20%22headerPrefix%22%3A%20%22%22%2C%0A%20%22highlight%22%3A%20null%2C%0A%20%22langPrefix%22%3A%20%22language-%22%2C%0A%20%22mangle%22%3A%20true%2C%0A%20%22pedantic%22%3A%20false%2C%0A%20%22sanitize%22%3A%20false%2C%0A%20%22sanitizer%22%3A%20null%2C%0A%20%22silent%22%3A%20false%2C%0A%20%22smartypants%22%3A%20false%2C%0A%20%22tokenizer%22%3A%20null%2C%0A%20%22walkTokens%22%3A%20null%2C%0A%20%22xhtml%22%3A%20false%0A%7D&version=master)

It's plausible to me that if we wanted we could produce a `marked` extension that provided a slightly different [tokenizer](https://marked.js.org/using_pro#tokenizer)?  It's possible if we go that route that we'll be able to tune the matching rules to reduce the amount of slightly finicky trailing/leading newline rejection/insertion I'm currently doing at the renderer layer.

imports/client/components/FancyEditor.tsx
```diff
@@ -0,0 +1,676 @@
...
+function matchUsers(users: Meteor.User[], searchString: string): AugmentedUser[] {
+  // No point doing all this matching work if there's no search string to match against.
+  if (!searchString) return [];
+
+  // We approximate case-insensitive search by converting both needles and
+  // haystacks to lowercase.
+  const needle = searchString.toLowerCase();
+
+  // Augment users with whether assorted fields match the search string or not, and where.
+  const augmentedUsers: AugmentedUser[] = users.map((u) => {
```
grantaelliott, line 223: If I'm not mistaken, we currently only subscribe to other users' display names, so these other filters can only ever return yourself.

ebroder: IIRC, https://github.com/deathandmayhem/jolly-roger/blob/main/imports/server/users.ts#L12-L28 ensures that you always have your own profile information available (and I don't think any of our subscriptions explicitly _exclude_ self anyway)

grantaelliott: Sorry, my commend was ambiguously worded. It definitely searches your own profile. The problem (I think...) is that it can't search email addresses (or other fields) of other users because it doesn't subscribe to that information, so the results are display name matches for anyone and any kind of match for yourself.

grantaelliott (commented):
I'm still working through this, but I have to step away for a while (maybe for the evening) and wanted to get some thoughts down, sorted more or less by importance:

- (see above) I believe there's an issue with searching other users by anything other than display name.
- Deep blockquotes can widen the chat and force horizontal scrolling. Can we set a maximum quote depth? (There's probably not really much use case for nesting quotes at all.) Setting `overflow-x: hidden` on Chat History might be worthwhile in any case.
- User names can overflow in at least two places:
  - In a match candidate, names can wrap and overflow the element. Unfortunately, this is very easy to trigger with even short names on mobile if the @ happens to be near the right of the input box because the match candidates are horizontally shifted to the cursor.
  - In a mention, a long name will wrap, and similarly escape the styled mention.
- I have mixed feelings about pre-wrap. While preserving newlines feels natural, preserving extra spaces (and tabs, unless they start a line) outside of a code block feels strange to me. But that's probably a discussion for another day.

Functionally, this is all great. I'm particularly happy to see mentions working and the implementation is slick.

imports/client/components/FancyEditor.tsx
ebroder: Hmm. Yeah, I think that's right. It effectively means we need a subscription to `huntProfiles`.

zarvox: Ah, good catch.  I suppose this would be the sort of thing that we could have the editor itself subscribe to, since nothing else on the page cares about the non-displayName fields for other users.

Alternately, we could say that completions have to target `displayName`?

grantaelliott: I think right now this also lets you mention someone not subscribed to this hunt, which is at best weird and at worst an easy way to accidentally send spoilers.

imports/client/components/PuzzlePage.tsx
```diff
@@ -1785,6 +1894,11 @@ const PuzzlePage = React.memo(() => {
       Puzzles.findOneAllowingDeleted(puzzleId)
   ), [puzzleDataLoading, puzzleId]);
 
+  const selfUserId = useTracker(() => Meteor.userId()!, []);
+  const users = useTracker(() => {
+    return MeteorUsers.find({}).fetch();
```
ebroder, line 1899:
```suggestion
    return MeteorUsers.find({ hunts: huntId }).fetch();
```

imports/client/components/FancyEditor.tsx
ebroder: Yeah, although that's an easy fix - I'll post a suggestion.

Oh - for what it's worth, we don't just subscribe to `displayNames`. `CallSection` subscribes to `avatars`, which means we have discord account info, so that's already available, though sort of by accident. I think it'd be pretty reasonable to subscribe to `huntProfiles` and have all the info we want, although I would be happiest if we could make it so that the puzzle page _only_ subscribed to `huntProfiles` and didn't _also_ subscribe to `displayNames` and `avatars`

ebroder:
Also, for what it's worth, it does seem worthwhile to have a circuit breaker to switch the editor back to the old behavior. I don't think I'd want that circuit breaker to affect _rendering_ of messages (once the messages are there, they're there, and we have to deal with them), but being able to turn off the editor feels like a good non-disruptive safety net in case we miss anything in our testing, since the surface area here is somewhat wide.

imports/client/components/PuzzlePage.tsx
zarvox: Good catch; we definitely only want to complete users associated with this hunt.

Maybe we should enforce that in `ChatNotificationHooks` as well?  Verify that a user is in the same hunt as the chat message before allowing them to be the target of an @-mention.

grantaelliott: I think so. Also I think in matchUsers, just in case a future subscription pulls in more users.

imports/client/components/FancyEditor.tsx
zarvox: Hmmm.  `huntProfiles` is a fair bit more data than `displayNames`, and it's used only by the chat input, so I feel a little bad having the other components on the page block on "all the data of all the people" when rendering the chat log and the rest of the page could proceed with `huntProfiles` getting loaded lazily in the background without triggering rerender thrashing for the whole PuzzlePage.

I think my current approach will be to sub to `huntProfiles` in `ChatInput`, far down in the component hierarchy and to leave the `displayNames` and `avatars` subs where they are for now.

imports/client/components/PuzzlePage.tsx
zarvox: Addressed in 01e2af8ff1ca5bd4f746ef44ecfe224112b60c0a

zarvox:
Rebased atop main.

Added a circuit breaker in 10265e31d00a8569cfe46bab9700bd9a8fee7684.

> (see above) I believe there's an issue with searching other users by anything other than display name.

Should be addressed by 1ff559f6a118766448c90e9ea79e3a31895becce

> Deep blockquotes can widen the chat and force horizontal scrolling. Can we set a maximum quote depth? (There's probably not really much use case for nesting quotes at all.) Setting overflow-x: hidden on Chat History might be worthwhile in any case.

Hmm.  To be honest, blockquotes are a bit of a pain and lead to confusing behavior because they more-or-less require you to do two line breaks between them and the next thing you want to write, which is awkward for something that does live display but avoids tracking blocks statefully, which is why I don't do anything to preview them.

I can try the `overflow-x: hidden` suggestion. because the lexer was throwing out my whitespace around them in strange ways.

> User names can overflow in at least two places:
> In a match candidate, names can wrap and overflow the element. Unfortunately, this is very easy to trigger with even short names on mobile if the @ happens to be near the right of the input box because the match candidates are horizontally shifted to the cursor.

Hmmm, I'll try making this so that the completion candidates container's `left` is at least 0 but no greater than `viewport width - completion box width`.  That should keep things on-screen for mobile.  I think the overflowing to the right is still the right thing for desktop widths.  Thanks for pointing this one out.

> In a mention, a long name will wrap, and similarly escape the styled mention.

I was thinking about this one in chat; Evan suggested I just make the `MentionSpan` have `display: inline`, which looks fairly reasonable.

> I have mixed feelings about pre-wrap. While preserving newlines feels natural, preserving extra spaces (and tabs, unless they start a line) outside of a code block feels strange to me. But that's probably a discussion for another day.

I also have somewhat mixed feelings about it, but I think without it you're back to having the rendered result look **so** different from the preview that the preview feels misleading.  (FWIW, I am also not thrilled with how blockquotes also cause this to happen, but I think those are less likely to be stumbled upon by accident than generic paragraph-wrapping).

But yeah, let's table that one for the time being.

zarvox:
Okay, I think this is ready for another round of review.

Of note:

* I tried the `overflow-x:` hidden idea, but if there was any appreciable content in the 18-nested blockquote, then each letter in the quoted text would get wrapped to its own line.  I think it's probably not worth trying to handle that case.  There will always be some inputs that we can't render well, and the overflowing content nonetheless didn't seem to harm the appearance of any of the other lines of content.  All the same, having the horizontal scroll bar is annoying, so even though this doesn't really solve `blockquote` nesting, it does at least prevent the scroll bar that would result from getting in the way of more recent chat messages, or confusing sticky-bottom scroll computations. 
* I made the autocomplete container placement smarter in 63dab75c132b639b07861f3ef475194b1c924090.  Now, it should behave well at the right edge on mobile, and also handle excessively-long display names.
* It turns out making mentions `display: inline` can cause an unfortunate appearance wart.  Instead, I removed the height restriction on MentionSpan, allowing the inline block to spread over multiple lines as needed.  This results in a thicker mention box, but it's contiguous and everything looks pretty reasonable otherwise.

imports/client/components/PuzzlePage.tsx
zarvox: Filtered down to users in the same hunt with a `displayName` set.

ebroder (approved):
Cool. I did another pass through and I can't find any other issues, so let's do this.

imports/server/hooks/ChatNotificationHooks.ts
```diff
@@ -0,0 +1,105 @@
...
+          if (matches) {
+            await ChatNotifications.insertAsync({
+              user: u._id,
+              sender: chatMessage.sender,
+              puzzle: chatMessage.puzzle,
+              hunt: chatMessage.hunt,
+              text: chatMessage.text,
+              content: chatMessage.content,
+              timestamp: chatMessage.timestamp,
+            });
```
grantaelliott, line 80: Despite the comment above, this actually does generate two separate notifications in the case of a coincident mention and ding. Since it looks like the intention is for both types of the notification to be the same (that is, one style isn't supposed to overrule the other), I think you just meant:
```suggestion
            usersToNotify.add(u._id);
```

grantaelliott:
> Hmm. To be honest, blockquotes are a bit of a pain and lead to confusing behavior because they more-or-less require you to do two line breaks between them and the next thing you want to write, which is awkward for something that does live display but avoids tracking blocks statefully, which is why I don't do anything to preview them.
> 
> I can try the `overflow-x: hidden` suggestion. I only implemented blockquote rendering because the lexer was throwing out my whitespace around them in strange ways and Evan indicated he liked having the option.
This compromise looks fine. If for some reason someone tries to deeply blockquote, their resulting mess will just be pushed off the screen and not significantly affect the chat for everyone.

grantaelliott (changes requested):
I'm really happy with this. Nicely done!

Please fix the duplicate notification issue noted above. I believe all my other concerns have already been addressed.

imports/server/hooks/ChatNotificationHooks.ts
zarvox: Quite right.  I might have messed this up in one of the rebases I did?  or maybe it was just buggy from the get-go.  Either way, good catch, and thanks!  Fixed in 27fcf78486573dbb0bc215ada2594c63975e8fe3

grantaelliott (approved):

zarvox:
Ah, this still needs a rebase.  🔜

PR #1313, opened 2023-01-09
Rework logging

### LINKED ISSUE #1297: Difference in behavior between client and server with optional arguments

ebroder:
This is my third crack at #1297, based on feedback from #1298.

When I dug into our logging infrastructure, I realized a few things. First, while the original idea of the "Ansible" library was to be able to log messages on the client that would also get seamlessly replicated to the server for persistence (i.e. FTL communication across the galactic space of the internet). In practice, we almost never used Ansible in client-side code (there were only about 5 call sites). Instead, for client-side code, we tended to use fine-grained `console.log` messages, which often supported being turned on or off at runtime for specific debugging cases.

Additionally, with the introduction of Bugsnag, we saw several cases where we wanted to both log an error and report that error to Bugsnag, leading to a fair amount of duplication of code of the form

```javascript
Ansible.error('Something went wrong', { error });
if (Bugsnag.isStarted()) {
  Bugsnag.notify(error, ...)
}
```

So...I pulled in a real grown-up logging library, [winston](https://github.com/winstonjs/winston#winston), which gives us enough capabilities to do all the things we need from logging today:

- We can transparently inject additional information to log messages, like automatically capturing `Meteor.userId` during method/publication invocations and adding it as log context on the server
- We can connect multiple log "transports", like having a Bugsnag transport that reports logs at the "warn" or "error" level if there's an error object attached to them
- We can have verbose log levels which aren't rendered by default. (Technically, we could do that before with `console.debug`, at least on the client-side, as browsers tend to not show that unless you explicitly opt in, but the messages still get rendered and stored)
- We can create child loggers with independent configuration so that we can turn (e.g.) debug logging on and off for `useCallState` independently of other debug logging. (Right now there's no easy way to get at that child logger to change it's configuration, but that was equally true of the old `DEBUG_LOGGING` variable.)

One thing I didn't do is convert `imports/client/tracing.ts` to use winston loggers, but it should be conceptually fairly straightforward.

In terms of review, 375204060587e5a83f28fb8612e040b7827bcb0c is quite voluminous; sorry about that. It's mostly rote conversions from either `console` or `Ansible` logging methods, but I did take the opportunity to clean up some of the message strings and additional contextual parameters, so it's not an automated conversion (or even a human-driven-automated conversion).

I'm also interested in feedback about whether this seems sophisticated enough that we could use this going forward in favor of overriding the `no-console` lint rule. (I'm not actively crusading over it, but I'm generally interested in keeping consistent downward pressure on eslint overrides in the codebase, either via improving our code style or by improving our eslint configuration)

Fixes #1297.


imports/client/hooks/useCallState.ts
```diff
@@ -499,17 +493,17 @@ const useCallState = ({ huntId, puzzleId, tabId }: {
...
+        logger.info('Creating new Mediasoup device');
         const newDevice = new Device();
         await newDevice.load({
           routerRtpCapabilities: JSON.parse(router.rtpCapabilities),
         });
         dispatch({ type: 'set-device', device: newDevice });
       })();
     } else {
-      console.log('Clearing Mediasoup device');
+      Logger.info('Clearing Mediasoup device');
```
zarvox, line 506: little-l `logger`

zarvox (approved):
This looks pretty good, and I'm happy for us to drop `console.log` and friends in favor of something more fully-featured.  Thanks for being so receptive to various bits of feedback along the way!

I pointed out in chat that this appears to have triggered us pulling in the `meteor-node-stubs` package, which increases bundle size by over 200k (>10% of current total code size), and that's not counting the `winston` code size.  If it's unavoidable, I guess I can live with it, but if there's a way we can hold the imports a little differently to impact a smaller bundle here I'd certainly appreciate it.

ebroder:
Hmm. I don't think there's any way to hold the imports differently with winston - it doesn't seem to let you import specific functionality, only the whole module, and that pulls in all the default transports (which I think is the main - although not only - problem):

```
  "util" in /Users/evan/src/jolly-roger/node_modules/@colors/colors/lib/colors.js (web.browser.legacy)
  "os" in /Users/evan/src/jolly-roger/node_modules/@colors/colors/lib/system/supports-colors.js (web.browser.legacy)
  "fs" in /Users/evan/src/jolly-roger/node_modules/winston/dist/winston/transports/file.js (web.browser.legacy)
  "path" in /Users/evan/src/jolly-roger/node_modules/winston/dist/winston/transports/file.js (web.browser.legacy)
  "zlib" in /Users/evan/src/jolly-roger/node_modules/winston/dist/winston/transports/file.js (web.browser.legacy)
  "http" in /Users/evan/src/jolly-roger/node_modules/winston/dist/winston/transports/http.js (web.browser.legacy)
  "https" in /Users/evan/src/jolly-roger/node_modules/winston/dist/winston/transports/http.js (web.browser.legacy)
```

PR #1318, opened 2023-01-10
Eliminate reduce-over-promise pattern

ebroder:
This was an attempt to avoid the no-await-in-loop linter rule, but because of the chaining, it creates massive stack traces with each loop showing up as an "awaited here" line.

If we want serialism, we shouldn't sneak past the linter to get it.

### DIFF
--- a/imports/server/hooks/HooksRegistry.ts
+++ b/imports/server/hooks/HooksRegistry.ts
@@ -21,39 +21,35 @@ class HooksRegistry {
   }
 
   async runPuzzleCreatedHooks(puzzleId: string) {
-    await this.registeredHooks.reduce(async (p, hook) => {
-      await p;
+    for (const hook of this.registeredHooks) {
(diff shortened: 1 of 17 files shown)


imports/server/hooks/HooksRegistry.ts
```diff
@@ -21,39 +21,35 @@ class HooksRegistry {
   }
 
   async runPuzzleCreatedHooks(puzzleId: string) {
-    await this.registeredHooks.reduce(async (p, hook) => {
-      await p;
+    for (const hook of this.registeredHooks) {
```
zarvox, line 24: This is orthogonal to this particular patchset, but it occurs to me that this file is probably one place we should take the hint from the lint and fire off hooks in parallel and do an `await Promise.all` for each of these here.

zarvox (approved):

PR #1324, opened 2023-01-10
Automatically re-join calls after code-push-triggered reload

### LINKED ISSUE #1314: Re-join call after code push causes a reload
Today, if we push out code to Jolly Roger, everyone's client will refresh and anyone currently on a call will get disconnected.

My intuition is that if we were previously in a call, we should re-join the call after the page is refreshed. There are a few tools we could use to make that happen. One is [Meteor's `reload` package](https://github.com/meteor/meteor/blob/devel/packages/reload/reload.js#L118-L154), which specifically allows you to thread through data across a reload. Another option would be to use `use-persisted-state` along with `sessionStorage`, which is local to the tab and doesn't persist. (`sessionStorage` might make it tricky to tell if the we refreshed the page due to code updates, or if the user refreshed the page on their own, in which case we plausibly would not want to rejoin the call)

@zarvox proposed an alternative on #594:

> As another point in design space for avoiding data loss during Meteor reload: Sandstorm had a scheme where if you had any grains open at the time, it'd inhibit reload from refreshing until you closed all of them or clicked on the "update available -- click here to reload" banner: https://github.com/sandstorm-io/sandstorm/blob/master/shell/imports/sandstorm-ui-topbar/topbar.js#L27-L55 . That approach makes it a bit more likely there are still old clients running around in the wild, though.

ebroder:
If Meteor triggers a page reload due to out-of-date code, stash the current state of whatever call we're in in sessionStorage, and use that to automatically re-join the call afterwards. We do some shenanigans to ensure that this only happens when Meteor triggers a page reload (as opposed to the user refreshing the page on their own).

Fixes #1314 

This doesn't 100% work yet - it doesn't suppress `allowInitialPeerStateNotification`, so if you're muted and the page refreshes for a code push, you'll get the large call popup. It's also somewhat annoying to test in development mode, because hot-module-replacement completely prevents Meteor-drive page reloads. I temporarily ran `meteor remove hot-module-replacement` temporarily in my local copy for testing, although I think that completely stopping and restarting the Meteor server might also trigger a full-page refresh.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -56,7 +56,8 @@
         "allow": [
           "_id", "_this", "_dropIndex", "_redirectUri", "_postRequest",
           "_anyMethodsAreOutstanding", "_schemaKeys", "_schema", "_retrieveCredentialSecret",
-          "_loginStyle", "_stateParam", "_name", "_allSubscriptionsReady"
+          "_onMigrate", "_migrationData", "_loginStyle", "_stateParam", "_name",
+          "_allSubscriptionsReady"
         ]
       }
     ],
--- a/imports/client/components/ChatPeople.tsx
+++ b/imports/client/components/ChatPeople.tsx
@@ -229,49 +229,57 @@ const ChatPeople = ({
 
   const { muted, deafened } = audioControls;
 
-  const joinCall = useCallback(async () => {
+  const joinCall = useCallback(() => {
     trace('ChatPeople joinCall');
-    if (navigator.mediaDevices) {
-      callDispatch({ type: 'request-capture' });
-      const preferredAudioDeviceId = localStorage.getItem(PREFERRED_AUDIO_DEVICE_STORAGE_KEY) ??
-        undefined;
-      // Get the user media stream.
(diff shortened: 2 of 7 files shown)


zarvox:
One thing that I'm not totally sure about is if the goal of this is actually reliably achievable under certain browser-enforced policies about media capture and playback, particularly being able to obtain an AudioContext or microphone capture before any user interaction takes place in the document.  In the case where the user has granted persistent microphone capture to the origin, I think we're probably fine.  However, if the user granted a one-time permission grant (which I'll point out is Safari's default behavior, and that you have to poke around in the UI to find a way to grant permission persistently), I could totally believe that the user would get prompted or if there was no mouse or keyboard event on the page, that the `getUserMedia` request might be dropped entirely.

zarvox:
I suppose in those cases it's still probably a win for them to get re-prompted rather than just silently dropped from the call, but we might have to either temper our expectations or otherwise guide users toward enabling persistent audio access.

ebroder:
Closed in favor of #1329

PR #1330, opened 2023-01-11
Fix assorted subtle chat message renderer bugs

zarvox:
* Paragraphs are not allowed to be children of spans.  We had some unnecessary component nesting that caused all markdown ranges to be rendered beneath a span.  Reducing indirection here simplifies the component tree and as a bonus avoids invalid nestings.
* Paragraphs need to render as inline with pre-wrap, not inline-block. If they are inline blocks, then they will tend to refuse to share a line with a mention (which must also be an inline block) which is worse layout.  The initial version of this code rendered paragraphs as a span, which used inline layout, so this fixes a regression I introduced in cea778556903b8480edced0886a13d2e330323d0.
* Ensure that unsupported tag types are rendered as pre-wrap raw text. This helps us do a better job of not letting marked's list-handling cause us to surprise users with the rendered result -- without the pre-wrap treatment, a multi-line bulleted list (which we intentionally, explicitly do not turn into a `<ul>` and `<li>`s) would render on a single line.

### DIFF
--- a/imports/client/components/ChatMessageV2.tsx
+++ b/imports/client/components/ChatMessageV2.tsx
@@ -13,7 +13,7 @@ const PreWrapSpan = styled.span`
 `;
 
 const PreWrapParagraph = styled.p`
-  display: inline-block;
+  display: inline;
   white-space: pre-wrap;
   margin-bottom: 0;
 `;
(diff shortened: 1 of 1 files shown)


ebroder (approved):

ebroder:
(Looks like the build trigger webhook got dropped, so this probably needs a rebase to kick it)

PR #1332, opened 2023-01-11
Prevent rerendering entire chat history

ebroder:
Whenever the data returned by a useTracker changes, the entire data set is replaced. This means that even if (e.g.) the only change was adding a new element to an array, all other elements of that array will return new objects, which won't compare as equal to the old objects (even if they have the same contents).

This means in practice that any time (e.g.) a new chat message is created, the entire chat history has to be re-rendered.

By using the useFind hook instead, we can significantly decrease the amount of work required. In my testing, this reduced render time by factor of roughly 10.

### DIFF
--- a/imports/client/components/FirehosePage.tsx
+++ b/imports/client/components/FirehosePage.tsx
@@ -1,4 +1,4 @@
-import { useSubscribe, useTracker } from 'meteor/react-meteor-data';
+import { useFind, useSubscribe, useTracker } from 'meteor/react-meteor-data';
 import { faEraser } from '@fortawesome/free-solid-svg-icons/faEraser';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import React, {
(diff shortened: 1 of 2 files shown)


ebroder:
Hmm. I have vague memories of this, but I can't find anywhere we actually dropped `useFind` from our code, other than your rewrite of WebRTC into `useCallState`. However, there have been a [bunch of patches](https://github.com/meteor/react-packages/compare/07f607c5d0cdb5d4476c4ae4acdc5eb0d9dd2500...master#diff-2b179edecef592a6a279f89228af38b0be3d13eaee875efbea1f700e0d8ca35d) to `useFind` since it was originally released, many of which look like they could be connected to dropped updates.

I didn't notice any issues while I was testing this, and I did test with both existing messages and on a puzzle that had no messages yet (since transition from empty to non-empty seemed like the obvious place for this to fail)

zarvox (approved):
All right, let's try it and see if we find any issues in the next couple days, and if not, maybe we can use `useFind` more again.

PR #1337, opened 2023-01-12
Highlight messages that trip your dingwords

zarvox:
This is kinda a starting point for discussion.  I'm not totally sold on this specific appearance for this, but I do like the idea of having "the message that summoned me to a puzzle" having something that indicates the specific message in the log as noteworthy.

I'm open to feedback on the idea, including "this is ugly/a bad idea; don't do this".  Thoughts?

### DIFF
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -50,6 +50,7 @@ import styled, { css } from 'styled-components';
 import Flags from '../../Flags';
 import Logger from '../../Logger';
 import { calendarTimeFormat, shortCalendarTimeFormat } from '../../lib/calendarTimeFormat';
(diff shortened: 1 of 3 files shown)


ebroder (approved):
This seems fine to me. I think it would be nice if we also highlighted the specific substring that matched (especially to help with debugging why a particular dingword is firing), but I'm not sure I know how to integrate that offhand with the v2 content.

zarvox:
I can think of a way I might achieve that in `ChatMessageV2` -- basically, search for dingwords in each range of text that we would render raw, and wrap the matching substring range in a `background-color`'d subspan?  I'll give it a shot later today.

ebroder:
I guess you can...decorate the dingword, and then this basically just works?

zarvox:
Can't use `decorate` outside of the editor.

PR #1339, opened 2023-01-12
Rework FirehosePage for chatmessage v2 

zarvox:
While we're doing layout, let's render with a more readable grid layout.  Does actual formatting on each message and supports both v1 and v2 chat messages, which *is* a bit more expensive -- I profiled it as taking about 3x as much wall-clock time per formatted message on a (separate, smaller) dataset on my M1 MBA as the current v1 chat (which doesn't do formatting at all).  Which is a bit.

However, I think this is probably good enough -- the worst of the FirehosePage is the initial loading time and streaming all the ChatMessage records from the server, which today for the 2022 takes 18 seconds, whereas rendering takes about 3 seconds for ~30000 chat messages.  The improved readability of having a more tabular structure seems plausibly worthwhile?  Especially since I expect most re-renders to hit the React memoization, and the worst loading/rendering times only show up once we're 30000 messages into a hunt anyway?

### DIFF
--- a/imports/client/components/FirehosePage.tsx
+++ b/imports/client/components/FirehosePage.tsx
@@ -15,7 +15,7 @@ import { indexedById } from '../../lib/listUtils';
 import ChatMessages from '../../lib/models/ChatMessages';
 import { indexedDisplayNames } from '../../lib/models/MeteorUsers';
 import Puzzles from '../../lib/models/Puzzles';
-import { ChatMessageType } from '../../lib/schemas/ChatMessage';
+import { ChatMessageType, nodeIsMention } from '../../lib/schemas/ChatMessage';
 import { PuzzleType } from '../../lib/schemas/Puzzle';
 import { useBreadcrumb } from '../hooks/breadcrumb';
 import useSubscribeDisplayNames from '../hooks/useSubscribeDisplayNames';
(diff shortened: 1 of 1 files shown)


ebroder:
Hmm. I'm not sure I personally find the tabular layout to be more readable, especially once things start wrapping:

Screenshot 2023-01-12 at 7 38 58 AM

As a strawman, what do you think about using `asFlatString` for the actual displayed text, and spilling the message onto its own line (with `white-space: pre-wrap`) if there's a newline in the message. Something like:

Screenshot 2023-01-12 at 7 44 40 AM

zarvox:
In weak defense: we could tune some of the space allocations, and it's worth noting that the most recent timestamps (which are what you're most likely looking at at any given moment) are going to be the shortest, since they'll just name the day and the time, so while tailing the log the timestamp shouldn't wrap.  Puzzle title and message sender are less controllable though.

That said: I don't think I personally care deeply about the UX here, so I could be talked into just about anything, and the render-to-flat-string approach might also give better performance?  I'll probably give that a go this afternoon.

zarvox:
Okay, I switched this to the suggested visual style.

ebroder (approved):

PR #1340, opened 2023-01-12
Use Mediasoup to heartbeat between servers

### LINKED ISSUE #1205: Make heartbeating between servers less noisy
With our in-house handling of audio calls, it's important that servers realize relatively quickly if another server goes down (since if that server was handling any calls, we want another server to step in to take over that call as quickly as possible - it will never be quick enough to be invisible to users, but we can try to prevent it from being completely disruptive).

At the time of implementation, we handled this by [significantly increasing](https://github.com/deathandmayhem/jolly-roger/commit/85afb07da7ed48d7395f6b85385c085c8a748852#diff-7e45e0895c47f12c4d7c342a68e5bb21a648b2f88546acc68d1eadaedbf2b746) the frequency with which a running server process updated it's MongoDB database record, resulting in one database write per server every 5 seconds or so, which is really quite a lot of ambient traffic for an otherwise completely idle site.

Acknowledging that this is a Hard Problem™ in computer science, it would be ideal if we could come up with a better way of detecting when a server has gone down that doesn't require quite as much database back and forth. A few specific ideas that spring to mind:

* Just use a different piece of infrastructure for heartbeating, e.g. Redis (with expiring keys)
* Have each Meteor process open a DDP connection to each other process, and rely on DDP's built-in ping-ponging to measure liveness. (This presents challenges around how we address individual server processes, or even individual servers, though that's slightly easier)

However, my current favorite idea is to have the mediasoup workers connect to each other. Mediasoup already has a built-in `PipeTransport` designed for multiple mediasoup workers to connect to each other. The primary intended use-case seems to be supporting [large-scale one-to-many broadcast](https://mediasoup.org/documentation/v3/scalability/#one-to-many-broadcasting), but it looks like we could probably establish transport without actually connecting any producers/consumers. We could then monitor those transports for disconnections as being indicative of server failure/shutdown. (For a sense of how to do this, look at the part of mediasoup's [pipeToRouter](https://github.com/versatica/mediasoup/blob/34cb60fab43ff367ec64e8034ea199697540490a/node/src/Router.ts#L724-L860) method before it actually wires up the provided producer, although we'd also need to wire up the inter-server communication of tuples and so on.) A few reasons this might not work:

* Transports are connected to _routers_ not to _workers_, which means that this could be `O(rooms * servers)` connections if we're not careful, which is a lot. It might be sufficient to connect to one room per server, although we'd need to be sure we handle the fenceposting correctly for when a server currently has no rooms allocated to it.
* A smaller thing, but we'd need to be sure that just having an open transport is sufficient to get any heartbeating between the two ends of the connection. If not, we might need to manually wire up data producers/consumers to send ping-pongs back and forth.

ebroder:
I don't plan to merge this until after this weekend, but I'll open this as a draft for feedback in case you have time to read through it.

Instead of heartbeating through MongoDB writes, we can use Mediasoup data producers and consumers to heartbeat between servers directly.

Each server creates a dedicated router for handling heartbeats; a DirectTransport, which allows the Meteor process to send data packets back and forth to the Mediasoup worker without a full WebRTC negotiation; and a DataProducer over that DirectTransport, which emits a heartbeat packet every 100ms (+/- jitter).

Then, whenever a server notices another server in the list of servers, it establishes a PipeTransport connection to that other server and pipes the heartbeat producer over that transport. It creates a MonitorConnectRequest record with the details of its half of the transport. Finally, it sets a watchdog - if it does not receive a heartbeat from the other server, it will mark that server as dead (and trigger the garbage collection cleanup hooks). The watchdog has an extra grace period before receiving the first heartbeat to allow for any potential delays in setting up the connection.

When a server observes a new MonitorConnectRequest record, it creates and connects its half of the PipeTransport and writes out a MonitorConnectAck so that the initiating server can connect, too. Then it resets the watchdog every time it receives a heartbeat packet.

Two notes about how Mediasoup works in this configuration:

First, as a reminder, Mediasoup's use of "producer" and "consumer" is from the perspective of a client, which makes it especially confusing here, because the sender of the heartbeats creates a "consumer" on the pipe transport, and the receiver creates a "producer".

Second, since PipeTransports don't use ICE to negotiation a connection, so you have to provide a single IP address, rather than a list of potential IP addresses, so we just grab the first one. In production, since we only use publicly-routable addresses for ICE, this should be fine.

Since this should allow us to detect dying servers at an acceptable pace for recovering WebRTC calls, we can also significantly relax the MongoDB-based heartbeats back to their previous frequency (heartbeats every 15 seconds with a 120 second expiration window).

Fixes #1205.

### DIFF
--- a/imports/server/garbage-collection.ts
+++ b/imports/server/garbage-collection.ts
@@ -24,12 +24,11 @@ async function cleanup() {
     },
   });
 
-  // Servers disappearing should be a fairly rare occurrence, but a disappearing
-  // server that's hosting an audio call blocks that call from proceeding until
-  // it's garbage collected, so set the timeouts on the tighter side. 5 seconds
-  // should be quick enough to recover without users noticing too much
-  // interruption, but long enough to account for transient blocking.
-  const timeout = new Date(Date.now() - 5 * 1000);
+  // Servers disappearing should be a fairly rare occurrence, so it's
+  // OK for the timeouts here to be generous. Servers get 120 seconds
+  // to update before their records are GC'd. Should be long enough to
+  // account for transients
+  const timeout = new Date(Date.now() - 120 * 1000);
   const deadServers = await Servers.find({ updatedAt: { $lt: timeout } })
     .mapAsync((server) => server._id);
   if (deadServers.length === 0) {
@@ -45,8 +44,15 @@ async function cleanup() {
   await Servers.removeAsync({ _id: { $in: deadServers } });
 }
 
+export async function cleanupDeadServer(id: string) {
+  for (const f of globalGCHooks) {
+    await f([id]);
+  }
+  await Servers.removeAsync(id);
+}
+
 function periodic() {
-  Meteor.setTimeout(periodic, 500 + (1000 * Random.fraction()));
+  Meteor.setTimeout(periodic, 15000 + (15000 * Random.fraction()));
   void cleanup();
 }
 
--- a/imports/server/mediasoup.ts
+++ b/imports/server/mediasoup.ts
@@ -1,4 +1,5 @@
 import { promises as dns } from 'dns';
+import EventEmitter from 'events';
 import { networkInterfaces } from 'os';
 import { Meteor } from 'meteor/meteor';
 import { Address6 } from 'ip-address';
@@ -8,11 +9,14 @@ import Flags from '../Flags';
 import Logger from '../Logger';
 import { ACTIVITY_GRANULARITY } from '../lib/config/activityTracking';
 import { RECENT_ACTIVITY_TIME_WINDOW_MS } from '../lib/config/webrtc';
+import Servers from '../lib/models/Servers';
 import CallHistories from '../lib/models/mediasoup/CallHistories';
 import ConnectAcks from '../lib/models/mediasoup/ConnectAcks';
 import ConnectRequests from '../lib/models/mediasoup/ConnectRequests';
 import ConsumerAcks from '../lib/models/mediasoup/ConsumerAcks';
 import Consumers from '../lib/models/mediasoup/Consumers';
+import MonitorConnectAcks from '../lib/models/mediasoup/MonitorConnectAcks';
+import MonitorConnectRequests from '../lib/models/mediasoup/MonitorConnectRequests';
 import Peers from '../lib/models/mediasoup/Peers';
 import ProducerClients from '../lib/models/mediasoup/ProducerClients';
 import ProducerServers from '../lib/models/mediasoup/ProducerServers';
@@ -22,13 +26,16 @@ import TransportRequests from '../lib/models/mediasoup/TransportRequests';
 import TransportStates from '../lib/models/mediasoup/TransportStates';
 import Transports from '../lib/models/mediasoup/Transports';
 import roundedTime from '../lib/roundedTime';
+import type { ServerType } from '../lib/schemas/Server';
 import type { ConnectRequestType } from '../lib/schemas/mediasoup/ConnectRequest';
 import type { ConsumerAckType } from '../lib/schemas/mediasoup/ConsumerAck';
+import type { MonitorConnectAckType } from '../lib/schemas/mediasoup/MonitorConnectAck';
+import type { MonitorConnectRequestType } from '../lib/schemas/mediasoup/MonitorConnectRequest';
 import type { ProducerClientType } from '../lib/schemas/mediasoup/ProducerClient';
 import type { RoomType } from '../lib/schemas/mediasoup/Room';
 import type { TransportRequestType } from '../lib/schemas/mediasoup/TransportRequest';
 import throttle from '../lib/throttle';
-import { registerPeriodicCleanupHook, serverId } from './garbage-collection';
+import { cleanupDeadServer, registerPeriodicCleanupHook, serverId } from './garbage-collection';
 import ignoringDuplicateKeyErrors from './ignoringDuplicateKeyErrors';
 import CallActivities from './models/CallActivities';
 import onExit from './onExit';
@@ -41,6 +48,10 @@ const mediaCodecs: types.RtpCodecCapability[] = [
     channels: 2,
   },
 ];
+
+const heartbeatFrequency = 100;
+const heartbeatTimeout = heartbeatFrequency * 50;
+const heartbeatInitialTimeout = heartbeatTimeout * 2;
 
 type ConsumerAppData = {
   call: string;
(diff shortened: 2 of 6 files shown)


imports/server/mediasoup.ts
```diff
@@ -386,11 +538,17 @@ class SFU {
   }
 
   onTransportCreated(transport: types.Transport) {
+    const transportAppData = transport.appData as TransportAppData;
+
+    // We're not wiring onTransportCreated up to the monitor router, so this
+    // shouldn't happen
+    if (transportAppData.type !== 'webrtc') {
```
zarvox, line 545: If it's indicative of a programming error/unreachable state, should this assert or throw instead of just ignoring the event?

imports/server/mediasoup.ts
```diff
@@ -519,6 +677,99 @@ class SFU {
     });
   }
 
+  onMonitorTransportCreated(transport: types.Transport) {
+    const transportAppData = transport.appData as TransportAppData;
+
+    // We're not wiring onTransportCreated up to non-monitor routers, so this
+    // shouldn't happen
+    if (transportAppData.type === 'webrtc') {
```
zarvox, line 685: ditto -- should this be an assertion/throw an exception?

imports/server/mediasoup.ts
```diff
@@ -42,6 +49,10 @@ const mediaCodecs: types.RtpCodecCapability[] = [
   },
 ];
 
+const heartbeatFrequency = 100;
+const heartbeatTimeout = heartbeatFrequency * 10;
```
zarvox, line 53: 1 second is probably a bit aggressive, given that we run stuff in the cloud where you can just get random hypervisor stolen time, where latency can be variable, and where I'm not sure what our behavior is if a server gets GC'd while still running.  We should probably consider:

* loosening the timeout a bit or
* convincing ourselves we understand what would happen if we incorrectly GC'd a server that still thinks it's alive/may be still serving clients.  Would we want some sort of STONITH/fencing mechanism if a server observes that its Servers record got deleted by another server to just intentionally exit so that it'd get restarted and take a new server ID, rather than trying to carry on in an inconsistent state?

imports/server/garbage-collection.ts
```diff
@@ -45,8 +44,15 @@ async function cleanup() {
   await Servers.removeAsync({ _id: { $in: deadServers } });
 }
 
+export async function markServerDead(id: string) {
```
zarvox, line 47: nit: I'd call this `cleanupDeadServer` instead of `markServerDead` since we're not really doing mark/sweep but are cleaning up after this particular dead server.

imports/server/mediasoup.ts
ebroder: I'll say up front that I'm moderately opposed to introducing a STONITH mechanism, because I am always worried about the cascading failure of every node decides to shoot every other node. So I'd prefer to avoid that.

A server which has its server record deleted out from under it will actually mostly recover on its own - the server record is managed with an upsert, so it'll just get re-inserted the next time the server's periodic process runs. For any rooms that it was responsible for, it'll observe that the `Room` record was deleted out from under it and destroy the corresponding mediasoup router, which is fine - the room should have been re-assigned to a different server.

However, it won't re-insert any records tied to that `serverId` and created by active subscriptions, which currently is specifically `Peer` and `Subscriber` records, so those will disappear. Active peers will get stuck in a "no self peer" state, and active viewers will disappear from the viewer list. Not ideal.

I'm fine bumping the timeouts back up to 5s (and probably 10s for initial handshake). Either way, I would love if we could come up with some way to handle `Peer` and `Subscriber` records more gracefully, although nothing is really coming to mind at the moment.

zarvox: > I'll say up front that I'm moderately opposed to introducing a STONITH mechanism, because I am always worried about the cascading failure of every node decides to shoot every other node. So I'd prefer to avoid that.

Totally reasonable.

> However, it won't re-insert any records tied to that serverId and created by active subscriptions, which currently is specifically Peer and Subscriber records, so those will disappear. Active peers will get stuck in a "no self peer" state, and active viewers will disappear from the viewer list. Not ideal.
> 
> I'm fine bumping the timeouts back up to 5s (and probably 10s for initial handshake). Either way, I would love if we could come up with some way to handle Peer and Subscriber records more gracefully, although nothing is really coming to mind at the moment.

I suspect we'll have to involve the client, since the client will still expect terminating the subscription to trigger cleanup of their Peer or Subscriber record if we quietly fixed things up behind the scenes, so that would have to happen from the backend they're talking to.  On the bright side, for e.g. the Peer record, the client should be able to observe that their intended call state is still joined, that `joinSubRef.current` is still valid, and that they observed `selfPeer` transition from truthy to falsy.  I'm not certain exactly what the code would look like, but I suspect we can make the client notice that it got dropped and re-establish a new subscription.

With the switch to the 5s timeout/10s initial timeout, I'm no longer super concerned with this misfiring, so let's go ahead and merge this.

We normally only operate with one backend instance during the off-season, right?  So we'll probably not really see anything that goes wrong here until we scale up for 2024?

imports/server/mediasoup.ts
```diff
@@ -807,10 +1189,14 @@ Meteor.startup(async () => {
     getLocalIPAddresses() :
     await getPublicIPAddresses();
   Logger.verbose('Discovered announceable IPs', { ips: ips.map((ip) => ip.announcedIp ?? ip.ip) });
+  const [primaryIp, ...otherIps] = ips;
+  if (!primaryIp) {
+    throw new Meteor.Error('Unable to discover any IP addresses');
```
zarvox, line 1194: Does this mean we'll fail to start up the backend if you're, say, developing jolly-roger while disconnected from other networks?  That's probably a developer experience regression that I'd prefer to avoid.

zarvox (approved):

imports/server/mediasoup.ts
ebroder: In development, we get addresses from iterating over your network devices, so presumably you'll always have a loopback in the worst case

imports/server/mediasoup.ts
ebroder: I was hypothesizing that we could detect this in the `mediasoup:join` subscription - after creating the peer, create an observer on the peer and if it gets deleted but the subscription hasn't been stopped, just recreate it. (And do something similar in `subscribers.inc`).

I haven't decided if I think there's meaningful risk of a server being alive and thinking it still had a client connected to it, but the client has gone away - screw cases there include that the client is just gone, or that the client has reconnected to a different server and we're in the "clean up an old peer" section of the code.

That's correct that we only run with 1 backend, although I think realistically the bigger issue is that we just don't _use_ Jolly Roger in the off-season. (We could pretty affordably run with, like, 2 `t3a.micro` instead of just 1 if we would actually get good data out of it)

PR #1383, opened 2023-01-18
Remove io-ts and fp-ts from client-side bundle

ebroder:
As of a few versions ago, TypeScript has support for [type-only imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export). Type-only imports are fully erased from generated JS code, which means that the code imported is erased as well.

This is useful for us with schemas, since client-side code only depends on the type information. The runtime information (i.e. the generated schemas) is needed on the server, but not on the client.

Therefore, move our usage of the schemas into server-only code that doesn't make it into the client bundle and use type-only imports throughout client code. Additionally, add some linting configuration to prefer type-only imports where valid and ban non-type imports of schemas in client-side code. (This isn't perfect - client code could accidentally transitively import schemas via code in `imports/lib/` or something, but there's comparatively little library code, so this seems likely to be safe)

In my testing, this successfully drops `io-ts` and `fp-ts` from the client-side bundle, shrinking it by about 10% or 200K.


zarvox (approved):
Wow, this is great!  Thanks for looking into this.

One (nonblocking) thought: is there a way we can help ensure that when we create a new collection we remember to attach the schema in `imports/server/schemas.ts`?  I'm just imagining it being easy to accidentally forget to add the line there and be accidentally running with no schema enforcement/autoValue magic until it bites us.

ebroder:
Yeah, I agree that's a risk, but I wasn't able to think of any way to test for that - there's not (e.g.) a global registry of `Collection` objects we could iterate over.

We could potentially have `Base`'s constructor stash itself into a `Map`, and then iterate over the map to see if everything has a schema attached? That would catch anything that uses `Base`, although we've been leaning away from that more recently as we have more models that are purely server-managed.

ebroder:
Would you prefer I (a) merge this as-is (b) do the `Base` registry thing (c) wait until we come up with a better idea?

zarvox:
Go ahead and merge this as-is -- it's a significant code-size win and is likely to conflict with lots of things if we wait.

PR #1385, opened 2023-01-18
Add additional client-side Bugsnag reporting

ebroder:
This catches any Meteor method exception and reports it to Bugsnag. This will be duplicative with our server-side reporting, but we generally have significantly better contextual information on the client, which should make debugging easier.

Additionally, add breadcrumbs on method calls so we can better tell what was happening immediately prior to an error.

(I also dropped the places where we were only catching a method error so that we could report it to Bugsnag, since this will catch all of those now)

### DIFF
--- a/imports/methods/TypedMethod.ts
+++ b/imports/methods/TypedMethod.ts
@@ -1,6 +1,9 @@
 import { check } from 'meteor/check';
 import type { EJSONable, EJSONableProperty } from 'meteor/ejson';
+import { EJSON } from 'meteor/ejson';
 import { Meteor } from 'meteor/meteor';
+import Bugsnag from '@bugsnag/js';
+import Logger from '../Logger';
 import type ValidateShape from '../lib/ValidateShape';
 
 export type TypedMethodParam = EJSONable | EJSONableProperty;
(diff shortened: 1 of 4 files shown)


imports/methods/TypedMethod.ts
```diff
@@ -133,10 +133,60 @@ export default class TypedMethod<
...
-    return Meteor.call(this.name, ...args);
+    try {
+      if (Bugsnag.isStarted()) {
+        Bugsnag.leaveBreadcrumb('Meteor method call', {
+          method: this.name,
+          arguments: typeof args[0] === 'object' ? EJSON.stringify(args[0]) : undefined,
+        }, 'request');
+      }
+      return Meteor.call(this.name, ...args);
+    } catch (error) {
```
zarvox, line 144: Does `Meteor.call` throw?  I thought it took a final callback argument and signaled failure via passing an error as the first arg to the callback function.  I think if you want to catch errors here you'll need to wrap the callback function the user passed, or add one if they didn't pass one.

ebroder: Yes, you're right of course. (It does, theoretically, throw on the server, where if you don't pass a callback it's invoked synchronously, but we don't do that anymore and I'm fine dropping that functionality).

Now that I'm no longer on a plane with no wifi I've actually tested this - you can see an example error reported here: https://app.bugsnag.com/death-and-mayhem/jolly-roger-dev/errors/63c8133d2264e90008beed10

ebroder:
...I'm so confused why the tests are failing on this.

zarvox:
Huh.  I agree that's baffling; I also have no failures locally, though I admittedly didn't try running the build under docker.

My best guess would be that either:

1. something has gone awry on the GitHub Actions cache side or
2. something happens in the Docker environment that is different from what happens elsewhere

zarvox:
If I don't have a Bugsnag account configured locally, does this change mean I lose the in-browser logging for these failure modes?  Would you be up for adding a path to send call failures error to the Logger for the branch where Bugsnag isn't enabled?

ebroder:
Hmm. I guess I did go to all this trouble to make our logger configuration capable of transparently upgrading to Bugsnag reporting. How about this?

https://app.bugsnag.com/death-and-mayhem/jolly-roger-dev/errors/63c8133d2264e90008beed10?filters[event.since]=30d&filters[error.status]=open&event_id=63cacab200ac5f9214f80000 shows what it looks like

zarvox (approved):
Looks good, and thanks!

PR #1387, opened 2023-01-18
Show "leave call" button when in a bad state

ebroder:
We tell people they should leave and re-join the call, but we don't give them a way to do that.

Before:

Screenshot 2023-01-18 at 8 55 01 AM

After:

Screenshot 2023-01-18 at 8 54 05 AM

### DIFF
--- a/imports/client/components/CallSection.tsx
+++ b/imports/client/components/CallSection.tsx
@@ -481,40 +481,42 @@ const CallSection = ({
     setShowMutedBy('hidden');
   }, [callDispatch]);
 
+  let joiningCallAlert;
   if (!callState.device) {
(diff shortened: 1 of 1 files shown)


ebroder:
Actually I want to rework this some

ebroder:
OK. I'm happier with this approach; ready for review. (The appearance is the same, but the structure is different)

zarvox (approved):
This is definitely an improvement.

While testing this out, I admit I found the flicker of the rerenders as various bits of state trickle in a little bit distracting, but that's not a regression introduced here (and keeping the button row the same size is arguably an improvement!), so definitely no reason to hold this back.

PR #1394, opened 2023-01-21
Switch to zod for database schemas

ebroder:
I think the commit message from 46c9f0ba57a6a07ca967faefcdc3207a48a6c40e is a fairly good summary of the PR overall, so I'm going to reproduce it here:

> This is a new system for defining database schemas which exist both at the type layer and the runtime layer. Instead of using io-ts to define the type layer and compiling that into SimpleSchema (which enforces the runtime layer), we instead use Zod to define schemas and compile them into JSON Schema, which we can then attach directly to the Mongo database to enforce the runtime behavior.
> 
> Zod represents an improvement for our use cases over io-ts in a few ways:
> - Its syntax for expressing optional types is much more concise (no more `t.union([t.string, t.undefined])`)
> - It allows expressing additional non-purely-type constraints, such as regex matches or lengths (which we previously expressed in Overrides)
> - It is somewhat more pleasant to metaprogram over at runtime (e.g. to generate JSON Schema)
> 
> Meanwhile, JSON Schema is capable of expressing everything we currently want to support in our schema language (unlike SimpleSchema which can't support, e.g., tagged unions) and is relatively straightforward to recursively generate from the Zod schema object (unlike SimpleSchema, which makes it fairly challenging to express things like nested object types).
> 
> The biggest downside of this approach is that while SimpleSchema comes with a native understanding of Mongo operations, giving it built-in support for auto-populating fields (e.g. `createdAt` timestamps) and standardized transformations on values which worked across different Mongo operators (e.g. our `answerify` function for uppercasing answers), JSON Schema supports no runtime transformation, only validation, and Zod has no understanding of Mongo operations to apply transformations intelligently.
> 
> To work around this, we introduce logic which can take a Zod schema and manipulate it such that it's possible to use it to transform Mongo operations. There are three main moving parts to make this work:
> 1. We have to relax certain constraints on the schema (e.g. a required field may not be present in an update operation). This potentially means that our runtime validation will allow through an invalid operation, but since our primary goal is transformation and not validation, that's acceptable. (Ultimately, the JSON schema enforcement at Mongo itself ensures validity.) We can do this in advance in the `relaxSchema` function.
> 2. We have to account for Mongo's dot-separated addressing syntax, which can put nested fields at the top level of an update operation. To account for this, we group the fields of the update operation by their top-level key (or lack of key, for non-dot-separated fields). For each group, we generate a new schema which represents just that sub-object (or array element, etc.), and then apply that schema to the group, recursing as necessary. Because we can't enumerate all possible ways that nested fields can be addressed (e.g. arrays can be indexed with a number), we have to do this at runtime, although that should only have cost if we're actually updating a nested field. This is handled by `parseMongoOperationAsync` (with assistance from `getSchemaForField`)
> 3. We need to ensure that auto-populated fields are populated at the appropriate time (be that insert, upsert, or update). For this, we use Meteor's EnvironmentVariables (which are basically variables with dynamic scope) to flag what state we're currently in, and then use Zod's transform functions to populate fields at the appropriate times. We set the insert flag on inserts, the upsert flag when generating `$setOnInsert` in an update, and the update flag for the `$set` operation in an update. (This is handled sort of all over the place, but primarily through cooperation between `parseMongoModifierAsync` and `imports/lib/schemas/customTypes.ts`.)
> 
> On top of the environment variables, an auto-populated field requires us to make a distinction between the type that is returned from the schema's processing and the type that is guaranteed to be present in the database. Under normal circumstances, the type that is returned from the schema's processing would be Zod's "output" type (`z.output` or `z.infer`). However, that type isn't especially useful to us, so we repurpose the output type to instead represent the type that is guaranteed to be present in the database. This requires us to, frankly, lie to Zod and the type system by using transform functions to claim that certain fields are not nullable, when in fact they are. This happens in `imports/lib/schemas/customTypes.ts`.
> 
> However, we retain the characteristic of Zod that the input type (`z.input`) represents what the user is obligated to provide, at least for insertion of a new document. (As outlined above, user obligations around updates are a bit more...fuzzy.)
> 
> (Additionally, because Zod's transformations and other "effects" suffer from runtime type erasure, we need an additional annotation to generate the correct JSON schema.)
> 
> All of this functionality is wrapped up in a new `Model` class, intended to replace both the old `Base` class and direct usage of `Mongo.Collection`. Unlike `Base`, it wraps a Mongo collection, rather than extending it, giving us more control over the interfaces we choose to support. For the specific `Base` functionality of supporting soft-deletion, there's a separate `SoftDeletedModel` class.
> 
> Rather than requiring that schemas provided to them hew to a specific interface, both `Model` and `SoftDeletedModel` add fields (`_id`, `deleted`) that they require to be present on their schemas. This means that the Zod schema object can not on its own be used to derive a model's type, so we instead export a new `ModelType` helper, replacing the pattern of exporting `t.TypeOf<typeof MySchema>` in the schema file.
> 
> `Model` additionally includes a few additional affordances. First of all, it requires that all string types be non-empty (either via length constraint or via regex). This is an alternative to SimpleSchema's default behavior of replacing empty strings with undefined and has the benefit of being more explicit (and more likely to trip you up if you, e.g., try to `$set` a field to an empty string). And second, it includes `AllModels`, making it easier for us to write code which operates on all models (e.g. for future index management tooling).
> 
> There are a few smaller regressions in this approach relative to what we were doing previously. Specifically:
> - This will pull in Zod on the client-side. We might be able to fix that by pulling mutation methods out of Model, similar to what we're doing with `defineMethod`, but it seems annoying and messy so I'm holding off right now. Zod isn't that big, anyway, so the footprint impact isn't that bad (around 8KB or so).
> - There's no way of expressing `denyInsert` or `denyUpdate`. This seems fairly difficult to reintroduce, since it's not expressible in JSON schema. I'm not sure how useful these were in practice, so I don't feel bad about dropping them.
> - `Model` does not preserve our `allow` rules which allow admin clients to manipulate state. We could re-introduce them, but since those default Meteor methods are defined on the collection (and not our `Model`), they would completely bypass transformation and thus be somewhat unpredictable. I think that instead of supporting these, we should build out an admin UI for manipulating state, which would be much more predictable and easier to use.
> 
> Finally, we update `publishJoinedQuery` to accept either a `Mongo.Collection` or a `Model`.

In terms of how to review this, here's my recommended reading order:
- Start with `imports/lib/models/validateSchema.ts`. It's a good intro to how to manipulate Zod schema objects at runtime (and how to recurse over the various container types)
- Quickly skim through `imports/lib/schemas/generateJsonSchema.ts`, although it's not that exciting. The most important thing is the exported `MongoRecordZodType` type, which recurs often for "a Zod schema for a thing that could be a top-level Mongo record".
- Read through `imports/lib/models/Model.ts`. Use the tests in `tests/unit/imports/server/Model.ts`, starting at around line 217, to get an understanding of what the different functions (`relaxSchema`, `parseMongoOperationAsync`, `parseMongoModifierAsync`) are supposed to do.
- Hopefully from there everything should more or less make sense.


imports/lib/models/Model.ts
```diff
@@ -0,0 +1,187 @@
...
+  async upsertAsync(
+    selector: Selector<z.output<T>>,
+    modifier: Exclude<Mongo.Modifier<z.output<T>>, z.output<T>>,
+    options: {
+      multi?: boolean | undefined;
+    } = {},
+  ): Promise<{
+    numberAffected?: number | undefined;
+    insertedId?: z.output<T>['_id'] | undefined;
+  }> {
```
zarvox, line 109: This  implementation seems to result in an `updatedAt` field (using `updatedTimestamp`) being set on upserts-that-are-inserts, which is a behavioral difference from what we had before (which was that `createdAt` would be set on first insert, and `updatedAt` would only be set on subsequent updates).

imports/lib/models/Model.ts
```diff
@@ -0,0 +1,187 @@
...
+      return optionalPreservingPartial(this.schema).parseAsync({});
+    });
+    const setFields = await IsUpdate.withValue(true, async () => {
+      return optionalPreservingPartial(this.schema).parseAsync({});
+    });
+
+    const mergedModifier: Mongo.Modifier<z.output<T>> = {
+      ...rawModifier,
+      $setOnInsert: {
+        ...rawModifier.$setOnInsert,
```
zarvox, line 125: This appears to have `setOnInsertFields` overriding anything the user put in `$setOnInsert`.  I'd have expected the inverse behavior -- that if the user specifies a `createdAt` that we'd use the explicit value they gave, rather than inferring one.  Should we switch the order here?

Ditto `$set` and `setFields` below.

imports/lib/models/Model.ts
ebroder: Yeah, OK - I think I can get the right behavior if I (a) treat `IsUpsert` and `IsUpdate` as mutually exclusive (which matches the behavior of collection2) (b) if fields are set in both `$set` and `$setOnInsert`, drop them from `$setOnInsert`.

(I'll note that the `updatedAt` timestamp on `Servers` behaves differently than other `updatedAt` timestamps today - it _is_ written on insert as well as update)

imports/lib/models/Model.ts
ebroder: So I actually thought that we had written `{created,updated}{At,By}` to not be overridable by callers, but I just mis-remembered. I think I can make 3 changes to get the right behavior here:

- Define the various custom timestamp fields as `z.date().optional().transform(...)` instead of `z.undefined().optional()` (i.e. allow them to be explicitly passed in).
- Update the transforms to preserve a value if it's already set
- Pass the `$set` and `$setOnInsert` parameters to `update` into `schema.parse`.

Again, I'll write tests to cover all of this but I'll push up the behavior modifications for now.

ebroder:
Another thing I still need to think about:

Almost all of our uses of `autoValue` are pure "populate if unpopulated" operators (created/updated timestamps/users). However, there are two that do something slightly more complicated: `guess` on `Guess` and `answers` on `Puzzle`, both of which run their contents through `answerify`. I can think of a couple ways to handle these:

- We could bail on auto-processing the values at the schema layer and instead just specify a `/^[^a-z]+$/` regex constraint, which callers would have to respect.
- We could try to encode how to handle all of the various mongo operators besides `$set` and `$setOnInsert` (it looks like simpl-schema [looks at almost everything](https://github.com/longshotlabs/simpl-schema/blob/e8f91dcd5016b3998cbec425fa77f337b398fc96/src/doValidation.ts#L9-L12) except `$pull`, `$pullAll`, `$pop`, and `$slice`, all of which can only be used to remove existing values rather than add new ones). This is probably doable, but certainly somewhat complex.
- We could ban anything other than `$set`, `$setOnInsert`, and `$unset`. (Pretty dissatisfying, since we make use of `$addToSet`, `$push`, and `$pullAll`, but calling it out because it _does_ exist in the solution space)
- Do a mix of options 2 and 3 - support the operators we use today but no more (e.g. no `$inc`).

ebroder:
Another thing to figure out is how exactly we want to bypass the schemas for things like migrations. With `simpl-schema` we can pass `{ validate: false, getAutoValues: false }` to more or less fully suppress its behavior and interact with the database without intermediation. We can pass `bypassDocumentValidation: true` as part of the options object to prevent Mongo from applying the json-schema validation, but that won't prevent `zod` from validating what gets passed in. Maybe we want to bypass `Model` entirely and interact directly with the underlying `Mongo.Collection`? (Maybe if you pass `bypassDocumentValidation` into `Model` _it_ skips the schema entirely and just blindly passes through what it gets?)

imports/lib/models/Model.ts
zarvox: To be fair I also didn't remember the existing behavior, but the types at the `update` callsite permitted me to provide an explicit `updatedAt` field as either Date or undefined, so I expected that if I did provide a value it wouldn't go to `/dev/null`.  I think I don't have an opinion yet on whether we should differentiate between "explicitly provided a value which was undefined" and "did not set that key in the update object", but I think that's already the case with needing to provide `$set` vs `$unset`.  I could probably be convinced that we should forbid explicit values for our autoValue fields if you like that better, but I feel like being able to specify the value if you want a particular one is probably more flexible and just as usable in practice.

zarvox:
Re `answerify`: we do already call `toUpperCase` on each character entered by the user at https://github.com/deathandmayhem/jolly-roger/blob/5e3161cb99da5ae48dcf9ee1a2a852eeaa931a2d/imports/client/components/PuzzlePage.tsx#L1416 since we need to do some modifications client-side to test if a guess is a duplicate.  I think it would not be burdensome at all to simplify the behavior to just enforcing that the answer is uppercase, or even to tolerate mixed-case answers (as we saw exist in this year's hunt), so I think the first option you named is the one I lean towards.

Re skipping validation for migrations: I could buy that `bypassDocumentValidation` should imply skipping the schema entirely, but I could also believe that the "skip the schema entirely" flag should be separate (and we just only have use cases right now where we use both together).

ebroder:
It's worth noting that zod intends for transforms to be used for more...traditionally...transformational stuff (e.g. `z.string().transform(s => s.length)`, which takes a string as input and outputs a number). If we're going to move away from allowing the schema to transform data, then we _may_ want some guardrails to prevent ourselves from writing problematic schemas in the future.

The current implementation has some indirect guard-rails against doing that sort of thing (because `generateJsonSchema` will reject any "effect" - transform, refinement, or preprocessor - by default, mostly because effects suffer from runtime type erasure in a way that other zod types don't), but we may want something less indirect.

ebroder:
@zarvox: As an aside, now that we're starting to narrow in on details, I'm curious where this falls for you on the "good idea" to "batshit" scale, especially the usage of transforms to replicate `autoValue`. I'm feeling reasonably good about it right now, but it's possible that that's I'm feeling good about getting it working, rather than feeling good about it being a good design.

imports/lib/schemas/customTypes.ts
```diff
@@ -0,0 +1,43 @@
...
+export const IsUpdate = new Meteor.EnvironmentVariable<boolean>();
+export const IsUpsert = new Meteor.EnvironmentVariable<boolean>();
+
+// In several of these auto-generated fields, we lie to zod about the output
+// type of the field - with type assertions, we say that they are non-nullable.
+// This will be true in the database, even though the transform functions don't
+// always return values (e.g. createdTimestamp will always be a Date, it'll just
+// get set on insertion and not on subsequent updates).
+
+export const stringId = z.string().regex(Id).optional()
```
zarvox, line 17: I'm a little fuzzy on why `stringId` is optional -- is it because it's optional when you're providing a value to e.g. `insert`, even though it's obviously a required key in every document?  What exactly do the schema types represent?

imports/lib/schemas/customTypes.ts
```diff
@@ -0,0 +1,43 @@
...
+  .transform((v) => v as unknown as string);
```
zarvox, line 18: Do we want to generate the `Random.id()` here if it wasn't provided, or just continue delegating that to Meteor?

imports/lib/schemas/customTypes.ts
```diff
@@ -0,0 +1,43 @@
...
+(stringId as any).customJsonSchema = { bsonType: 'string', pattern: Id.source };
+
+export const createdTimestamp = z.date().optional()
+  .transform((v) => {
+    if (v) return v;
+    if (IsInsert.getOrNullIfOutsideFiber() || IsUpsert.getOrNullIfOutsideFiber()) return new Date();
+    return undefined as unknown as Date;
```
zarvox, line 25: Does reaching this `undefined` here indicate a programming error?  Should we just throw instead?

imports/lib/models/Model.ts
```diff
@@ -0,0 +1,187 @@
...
+    options: {
+      multi?: boolean | undefined;
+      upsert?: boolean | undefined;
+      arrayFilters?: { [identifier: string]: any }[] | undefined;
+    } = {},
+  ): Promise<number> {
+    // Types should prevent passing a full document as a modifier, but verify at
+    // runtime to be sure
+    const rawModifier: Mongo.Modifier<z.input<T>> = modifier as any;
+    if (modifierIsWholeDoc(rawModifier)) {
```
zarvox, line 74: I had not realized that we never use `update` with a full document, though it makes sense in hindsight.  I went through all our updateAsync calls to be sure, and yep, every one of them uses `$set` or `$addToSet` or some other operator.  Neat!

imports/lib/models/Model.ts
```diff
@@ -0,0 +1,187 @@
...
+    modifier: Exclude<Mongo.Modifier<z.output<T>>, z.output<T>>,
+    options: {
+      multi?: boolean | undefined;
+    } = {},
+  ): Promise<{
+    numberAffected?: number | undefined;
+    insertedId?: z.output<T>['_id'] | undefined;
+  }> {
+    const rawModifier: Mongo.Modifier<T> = modifier as any;
+    if (modifierIsWholeDoc(rawModifier)) {
```
zarvox, line 111: Note: I think we do have one `upsertAsync` call that uses an empty doc as the modifier: https://github.com/deathandmayhem/jolly-roger/blob/5e3161cb99da5ae48dcf9ee1a2a852eeaa931a2d/imports/server/addUsersToDiscordRole.ts#L65-L70

zarvox:
> It's worth noting that zod intends for transforms to be used for more...traditionally...transformational stuff (e.g. z.string().transform(s => s.length), which takes a string as input and outputs a number). If we're going to move away from allowing the schema to transform data, then we may want some guardrails to prevent ourselves from writing problematic schemas in the future.
>
> The current implementation has some indirect guard-rails against doing that sort of thing (because generateJsonSchema will reject any "effect" - transform, refinement, or preprocessor - by default, mostly because effects suffer from runtime type erasure in a way that other zod types don't), but we may want something less indirect.

+1 to generally "strongly constrain the ability to be clever/problematic".  I will probably achieve this by never spending the time to learn how to do dangerous/fancy things with zod beyond the examples given here, but if there's some other technical refinement to keep future us from trying to be too clever by half, then I'm happy with additional guardrails.

 > @zarvox: As an aside, now that we're starting to narrow in on details, I'm curious where this falls for you on the "good idea" to "batshit" scale, especially the usage of transforms to replicate autoValue. I'm feeling reasonably good about it right now, but it's possible that that's I'm feeling good about getting it working, rather than feeling good about it being a good design.

I'm broadly enthusiastic about this change.  At a high level: I don't fully understand how this all works, but I also didn't fully understand how the previous schema system worked either.  From a user-visible-surface-area perspective, the old and the new seem relatively comparable.  This approach seems no *more* spookily magical to me as a user than io-ts+simpl-schema+collection2.  I read it as probably a couple fewer layers of complexity (the cleaning/transformation layer is now owned by us), usage appears basically equivalent outside the schema/model decls, and I expect it will be more flexible with regard to supported types (like, I expect zod to more likely support optional arrays of enums or literals, and have a better story around tagged enums), smaller in code size, and equally able to address our needs.  This seems like a pretty big win to me!

While I'm always cautious about potential subtle behavioral changes, I think that's something we can de-risk with a reasonably comprehensive set of unit tests.  I'm not sure if you can have unit tests for "this is expected to generate a type error" but if you can, it would be really compelling if we had tests that demonstrated that the type checker rejects unknown keys in an object passed to `insertAsync` and requires all non-optional keys.

I haven't done a direct comparison, and I imagine it's hard to do so without doing the full conversion of all the collections, but it would be interesting to know if `tsc` (or just `meteor npm run lint`) is faster or slower with the old approach or the new one.

imports/lib/schemas/customTypes.ts
ebroder: If you look at the resulting type of `stringId`, it has an input type of `string | undefined` and an output type of `string`. The intended use of input vs. output types is connected to parsing (or validating) a schema:

```typescript
const cleaned: z.output<typeof schema> = schema.parse(raw as z.input<typeof schema>)
```

However, I'm bending those semantics a little bit, and instead using the semantic that the input type represents what must be present on a record that is passed to `insert` and the output type indicates what we expect to get back from the database. This means that `schema.parse` will not actually return the output type, but that's fine, because we are only using the return value of `schema.parse` as an intermediate anyway before sending it off to Mongo.

That means that if you're inserting a new document, you don't have to provide the `_id`, and passing the input through `parse` won't populate one, but every document you get back from the database will have one. The optional-ness of it gets stripped away by the `transform` function on the next line - the parameter `v` is of type `string | undefined`, so the casting is a lie to the type system in order to establish the input and output types as being different.

imports/lib/schemas/customTypes.ts
ebroder: I figured it was easiest to continue delegating to Meteor. Meteor has some complex behavior around ID generation in order to ensure that (e.g.) stub method implementations generate the same IDs as their corresponding server-side runs, and even if we're not relying on that behavior, I didn't see any reason to suppress it.

imports/lib/models/Model.ts
ebroder: Ah, OK, so `modifierIsWholeDoc` should just return true if the object is empty.

imports/lib/schemas/customTypes.ts
ebroder: No, it indicates that we're applying the schema in a non-insert, non-upsert (i.e. update) context. That means that we're only updating a subset of the schema (because we don't support whole-document updates), so it's OK if this field remains unpopulated.

imports/lib/schemas/Setting.ts
```diff
@@ -1,92 +1,78 @@
...
+    value: z.object({
+      sharedSecret: z.string(),
+      scriptId: z.string(),
+      contentHash: z.string(),
+      endpointUrl: z.string().optional(),
+    }).strict(),
+  }).strict(),
 ]);
-export type SettingType = t.TypeOf<typeof SettingCodec>;
+const Setting = z.object({
```
zarvox, line 62: Ah, I imagine this particular `z.object()` is probably worth extracting as `BaseSchema` and then each of the things that we have that are based on `Base` can just `.and()` in the relevant additional fields.

imports/lib/schemas/customTypes.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const updatedUser = z.string().regex(Id).optional()
+  .transform((v) => {
+    if (v) return v;
+    try {
+      if (IsUpsert.get()) return Meteor.userId();
+    } catch (e) {
+      /* ignore */
+    }
+    return undefined as unknown as string;
```
zarvox, line 63: Since `updatedUser` is generally expected to be optional when returned from the DB, do you need the `as unknown as string` here?

imports/lib/schemas/Setting.ts
ebroder: Yeah, I was going to double-back and make something more like `Base`, I just wanted to see if I could get the discriminated union type working. (I am somewhat pleased that it worked with no changes to application code)

imports/lib/schemas/customTypes.ts
ebroder: Yeah, that's right, I got copy-paste happy. (I think/hope I would have eventually caught this in my tests, but will fix)

tests/unit/imports/server/Model.ts
```diff
@@ -0,0 +1,392 @@
...
+        const id = Random.id();
+        await model.upsertAsync(id, {});
+        const record = (await model.findOneAsync(id))!;
+        assert.isOk(record);
+
+        assert.isUndefined(record.updatedAt);
+
+        // It's a bit odd that updatedAt doesn't get set on an upsert that updates
+        // an existing document (rather than inserting a new one), but there's no
+        // $setOnNotInsert operator and this matches our historical behavior
```
zarvox, line 145: Huh, I didn't realize that was how `updatedAt` behaved today, but yeah, it looks like FeatureFlags never gets an `updatedAt` set.

Maybe it makes sense for collections we plan to purely upsert into to use `lastWriteTimestamp` instead of `updatedTimestamp`? (and maybe something with similar behavior for `updatedBy`?)

ebroder: When I realized that this was actually how `Base`'s `updatedAt` behaves today, my conclusion was mostly that we should just replace it with something more like `lastWriteTimestamp` across the board. I don't think there's a meaningful information difference between "updatedAt is unset" and "updatedAt has the same value as createdAt", and I don't think there's anywhere we need to (e.g.) query for "updatedAt is unset because this record has been inserted but never updated"

However, there's enough going on that I don't want to add unnecessary behavioral change on top of everything else, so I've made myself a note to follow up on this after this PR.

imports/lib/models/SoftDeletedModel.ts
```diff
@@ -0,0 +1,200 @@
...
+      z.ZodIntersection<Schema, z.ZodObject<{ deleted: typeof deleted }>>
+  > {
+  constructor(name: string, schema: Schema) {
+    super(
+      name,
+      schema instanceof z.ZodObject ?
+        schema.merge(z.object({ deleted })) :
+        schema.and(z.object({ deleted })) as any
+    );
+  }
```
ebroder, line 70: This ends up causing a bit of a style change in how we export types for our models.

I tried to write this class so that the schema being passed in had to already have a `deleted` property on it, but I couldn't figure out how to express that as a type constraint on the `Schema` type, so instead I decided to add the `deleted` property as part of the construction of a `SoftDeletedModel`.

That in turn means that the schema you pass in isn't necessarily reflective of the actual type of the collection. Therefore, instead of exporting the record type from the schema file, I added a new `ModelType` helper, with the expectation that you export the type in the model declaration, not the schema declaration.

This will churn imports everywhere a little bit (look around for e.g. `SettingType`), but I couldn't figure out any other way to do it. I think the minor upside is that it lets us potentially write more powerful extensions of model functionality if we want to.

ebroder:
> * [ ]  Look into whether we need an equivalent to simpl-schema's empty string to undefined coercion.

OK. I think we should do something about this. The interactions with controlled form elements in React means that we're likely to end up with empty strings in the database if we do nothing. Meanwhile, the tendency towards testing nullish-ness over false-iness means that empty strings in the database are likely to lead to logic bugs. Therefore, IMO it's desirable to do what we can to prevent empty strings from getting into the database across the board.

I think my preferred strategy for dealing with this is to require that we declare all string types as `z.string().min(1)`. That, on its own, doesn't prevent us from accidentally using a normal `z.string()`, but we can catch the mistake by adding a validation step to creating a model (i.e. have the constructor for `Model` walk the schema and look for string fields that don't have a length constraint).

We could alternatively have (e.g.) a `nonEmptyString` in `customTypes.ts` that automatically coerces empty strings to undefined, but I like the explicitness of having a minimum length constraint. It pushes the burden of checking for empty strings to the method that's manipulating the database and (IMO more importantly) forces the caller to deal with if they're doing a `$set` or an `$unset` on mutations.

The downside is that we'll probably have to track down a new set of bugs where we're not properly doing that coercion today (and which would throw errors with the new style of constraint), but I think it eliminates the class of bug where we (e.g.) try to `$set: {url: undefined}`, which is of course a noop.

ebroder:
I guess an alternative would be to use `.nullable()` instead of `.optional()` and have a helper that coerces an empty string to `null`. That has the advantage that `null` can actually be stored in Mongo, so `$set` operations would continue to work normally, but Mongo doesn't distinguish between `null` and unset fields at query time. And then we just declare across the board that we don't use `.optional()` fields anywhere (and similarly enforce that with validation).

tests/unit/imports/server/Model.ts
zarvox: I agree we should make such changes later/orthogonally, and agree with your general logic about what we should want here.

There's arguably a (probably small, kinda dumb) benefit at the margin of having less data to serialize/push down from the server/parse (dates are comparatively expensive fields), but we can presumably reap that benefit nearly as easily by being more intentional about field projection when publishing, especially with all the nice infrastructure from #1410.

imports/lib/models/SoftDeletedModel.ts
zarvox: Seems fine.  This also has the potential upside of making app code only import from `models`, instead of importing the collection from `models` and the type from `schemas`, which feels like maybe a slightly cleaner layering?

zarvox:
> I think my preferred strategy for dealing with this is to require that we declare all string types as `z.string().min(1)`. That, on its own, doesn't prevent us from accidentally using a normal `z.string()`, but we can catch the mistake by adding a validation step to creating a model (i.e. have the constructor for Model walk the schema and look for string fields that don't have a length constraint).

This is also my favored approach.  I generally prefer `undefined` over `null`, and since neither Mongo nor EJSON have a way to represent a present-but-undefined-not-null value, keeping such a thing inexpressible seems consistent.

While there's still some risk we'll write bugs where we client or server code doesn't `undefined` out empty strings correctly, this seems like the sort of thing that being principled about at the data layer will help us get a grip on with more comprehensive integration tests, and I like that the explicit backpressure is applied on the code with the bug, rather than "solving" the problem by relying on some magical action-at-distance.

tests/unit/imports/server/Model.ts
ebroder: They are, for what it's worth, somewhat less expensive than they look. Chrome (at least) negotiates the `permessage-deflate` extension to websockets, and some very casual testing shows that we get pretty significant compression if the `updatedAt` timestamp is the same (or close to) the `createdAt` timestamp:

```javascript
import zlib from 'zlib';
const record = Models.FeatureFlags.findOne();

> record.updatedAt === undefined
true

> zlib.deflateSync(EJSON.stringify(record)).length
147
> zlib.deflateSync(EJSON.stringify(Object.assign({}, record, { updatedAt: record.createdAt }))).length
153
> zlib.deflateSync(EJSON.stringify(Object.assign({}, record, { updatedAt: new Date(record.createdAt.getTime() + 1) }))).length
154
> zlib.deflateSync(EJSON.stringify(Object.assign({}, record, { updatedAt: new Date(record.createdAt.getTime() + 2) }))).length
155
```

So as long as the differences in timestamps are relatively small (which we'd expect in the cases where we were introducing an `updatedAt` timestamp that previously wasn't present), we only see a 6-8 byte difference in length of the deflated string that actually gets sent over the wire (vs. roughly 36 bytes for an un-deflated `updateAt` field + value in EJSON encoding).

imports/server/hooks/ChatNotificationHooks.ts
```diff
@@ -71,7 +72,7 @@ const ChatNotificationHooks: Hookset = {
     await Promise.all(collected.map(async (userId: string) => {
       await ChatNotifications.insertAsync({
         user: userId,
-        sender: chatMessage.sender,
+        sender,
```
ebroder, line 75: For what it's worth, this is the result of slightly better type precision than we had previously, and I'm kind of pleased about it.

Because this is accessing an attribute in a callback, TypeScript assumes that the object could potentially be mutated, and thus ignores the truthiness check previously on line 22 - within this function, `chatMessage.sender` is of type `string | undefined`.

With the old `Base` logic, `insert` and `insertAsync` accepted a `Partial<T>`. We didn't have any way at the type layer to track whether a field would be auto-populated or not, but that meant that the type layer would accept missing or null fields that were required by the schema. (Validating the object against SimpleSchema would throw a runtime error, ultimately ensuring consistency)

However, the new `Model` implementation can tell the difference (by using Zod's "input" vs. "output" construct), and correctly flagged that, if `chatMessage.sender` is optional, then this is potentially invalid.

imports/lib/schemas/User.ts
```diff
@@ -14,56 +13,35 @@ declare module 'meteor/meteor' {
...
+  services: z.any().optional(),
+  profile: z.object({}).optional(),
+  roles: z.record(z.string(), z.string().array()).optional(),
+  hunts: z.string().regex(Id).array().optional(),
+  displayName: z.string().optional(),
+  googleAccount: z.string().optional(),
+  googleAccountId: z.string().optional(),
+  discordAccount: DiscordAccount.optional(),
+  phoneNumber: z.string().optional(),
+  dingwords: z.string().array().optional(),
```
ebroder, line 42: For now, I decided to preserve our current behavior, that `MeteorUsers` is just a wrapper around `Meteor.users`. We historically did that to ensure that the schema was applied, although that's less important now that that happens in a standalone file.

However, I think we _may_ want to consider turning `MeteorUsers` into a `Model` object that wraps `Meteor.users` instead. It has the advantage that our application code can only deal with `Model`s, and doesn't need to think about both (e.g. `publishJoinedQuery`, which currently would have to continue accepting both, since we join onto users).

I think that, so long as we guarantee the types lined up, this should be safe (since we'd always be passing `Meteor.User`-typed objects around), and I can't think of any way it would interfere with places that we touch the actual accounts APIs. But still - I think that's an opportunity for future experimentation.

eslint/rules/no-sync-mongo-methods.ts
```diff
@@ -1,5 +1,5 @@
 import { ESLintUtils } from '@typescript-eslint/utils';
```
ebroder, line 1: One note: because `Model`s wrap `Mongo.Collection` instead of inheriting from it, this will not catch synchronous reads against `Model`s.

I'm choosing to defer this for the time being because I want to rework this check a little bit. (Specifically: I want to broaden it to more than just Mongo methods, now that Meteor is increasingly supporting async versions of core APIs). Given the current state of our code and infra, I'm reasonably confident that it'll be easy to go through and clean up any accidental synchronous calls introduced between now and then.

imports/lib/schemas/ChatMessage.ts
```diff
@@ -1,29 +1,27 @@
...
+  type: z.literal('mention'),
+  userId: z.string().regex(Id),
 });
-export type ChatMessageMentionNodeType = t.TypeOf<typeof MentionBlock>;
+export type ChatMessageMentionNodeType = z.infer<typeof MentionBlock>;
 
-const TextBlock = t.type({
-  text: t.string,
+const TextBlock = z.object({
+  text: nonEmptyString,
```
zarvox, line 13: Due to some constraints from slate (and a desire to preserve the validity of `ChatMessageContent` as a slate document) the `text` here may actually be an empty string, and should be stored and retrieved as such.

imports/lib/models/validateSchema.ts
```diff
@@ -0,0 +1,65 @@
...
+    case z.ZodFirstPartyTypeKind.ZodBoolean:
+    case z.ZodFirstPartyTypeKind.ZodNever:
+      // No validation needed
+      break;
+
+    case z.ZodFirstPartyTypeKind.ZodString: {
+      // String fields must not accept empty strings
+      const result = schema.safeParse('');
+      if (result.success) {
+        throw new Error(`String fields must not accept empty strings (${path.join('.')})`);
```
zarvox, line 57: This exists to help keep us from writing a schema with a string field that permits empty strings, yes?  I guess the need for the possibly-empty string in `ChatMessage` implies that this validation needs some way to permit an exception?

imports/lib/models/Model.ts
```diff
@@ -0,0 +1,465 @@
...
+    O extends Omit<Mongo.Options<z.output<this['schema']>>, 'transform'>,
+  >(
+    selector?: S,
+    options?: O,
+  ) {
+    return this.collection.find(selector, options) as
+      Mongo.Cursor<SelectorToResultType<z.output<this['schema']>, S>>;
+  }
+
+  findOne<S extends Selector<z.output<this['schema']>>, O extends Omit<Mongo.Options<z.output<this['schema']>>, 'limit' | 'transform'>>(
```
zarvox, line 428: I guess we still need all the sync methods for now because we use them in client-side code?

I wonder what Meteor's plans are re: deprecating/removing the sync methods from the minimongo implementation or just leaving them in.  If they eliminate them, then we have a bunch of frontend code rewriting to do, and I'm not sure that will play well with React where we generally expect synchronous results.

imports/lib/schemas/ChatNotification.ts
```diff
@@ -1,72 +1,28 @@
...
-import type { Overrides } from './typedSchemas';
-import { buildSchema, inheritSchema } from './typedSchemas';
+import withCommon from './withCommon';
 
-const SharedFields = {
+// A notification triggered by a chat message sent by a user.
+const ChatNotification = withCommon(z.object({
   // The userId of the user for whom this notification targets.
-  user: t.string,
+  user: z.string().regex(Id),
```
zarvox, line 10: I notice we use this `z.string().regex(Id)` a bunch for all the foreign-key references.  Do you have an opinion on whether it would be good or counterproductive to have a `customType` for an id from another collection?  I could see it being neat to be able to potentially traverse referenced objects, but I also don't see it being super likely we write any bugs where we pull from the wrong model in the first place, and if we named the collection the foreign key referenced then we'd have schemas depending on model types, which seems kinda messy (we'd still have a DAG, but it'd be hairier).

I think my position is: writing this as `foreignKey` could be a neat shorthand to have (where `foreignKey` is just `z.string().regex(Id)`, but we should punt for now on referencing other models, and also this is fine as written.

imports/lib/schemas/User.ts
```diff
@@ -14,56 +13,35 @@ declare module 'meteor/meteor' {
...
+  _id: stringId,
+  username: z.string().regex(/^[a-z0-9A-Z_]{3,15}$/).optional(),
+  emails: z.object({ address: z.string().regex(Email), verified: z.boolean() }).array().optional(),
+  createdAt: z.date().optional(),
+  lastLogin: z.date().optional(),
+  services: z.any().optional(),
+  profile: z.object({}).optional(),
+  roles: z.record(z.string(), z.string().array()).optional(),
+  hunts: z.string().regex(Id).array().optional(),
+  displayName: z.string().optional(),
```
zarvox, line 37: Do we want these `z.string()`s to be `nonEmptyString`s?

imports/lib/schemas/customTypes.ts
```diff
@@ -0,0 +1,91 @@
...
+      if (IsUpdate.get()) return Meteor.userId() ?? undefined;
+    } catch (e) {
+      /* ignore */
+    }
+    return undefined;
+  });
+attachCustomJsonSchema(updatedUser, { bsonType: 'string', pattern: Id.source });
+
+export const answer = z.string().transform((v) => answerify(v));
+attachCustomJsonSchema(answer, { bsonType: 'string', pattern: '^[^a-z]+$' });
```
zarvox, line 91: Doesn't have to happen in this PR, but I would prefer to move away from being prescriptive about what an answer can be.  I think the reality is that we have tried to be more strict about this than several writing teams, with the result being us having to on multiple occasions adapt on the fly out-of-band during hunt to try to express novel answer contents because we tried to enforce a rule that never really existed.

I feel like we'd get just as much value out of just uppercasing things by default in the UI, but letting the serialization/DB layer become more flexible here.  We could even allow operators to override the UI-imposed constraints.

We can keep the `answerify` transform for now for minimizing behavioral change in this PR, but I think the particular custom JSON schema on this line isn't needed.

imports/lib/schemas/withCommon.ts
```diff
@@ -0,0 +1,19 @@
+import { z } from 'zod';
+import type { MongoRecordZodType } from './generateJsonSchema';
+import { TimestampFields } from './withTimestamps';
+import { UserFields } from './withUsers';
+
+// We could compose withTimestamps and withUsers, but in the case that the
+// passed in schema is not a ZodObject, we'd end up creating multiple nested
```
zarvox, line 7: Ahhh, and `Setting` in particular, with a discriminated union at the root, is not a `ZodObject`

imports/lib/schemas/withCommon.ts
```diff
@@ -0,0 +1,19 @@
...
+// intersections, so this makes the resulting schema (and type) slightly (though
+// only slightly) cleaner.
+export default function withCommon<
+  T extends MongoRecordZodType
+>(schema: T):
+T extends z.ZodObject<infer Shape, infer UnknownKeys, infer Catchall> ?
+  z.ZodObject<Shape & typeof TimestampFields & typeof UserFields, UnknownKeys, Catchall> :
+  z.ZodIntersection<T, z.ZodObject<typeof TimestampFields & typeof UserFields>> {
+  return schema instanceof z.ZodObject ?
+    schema.extend({ ...TimestampFields, ...UserFields }) :
```
zarvox, line 17: Since `.extend()` can clobber fields, should we do anything to forbid `schema` from containing any of the `TimestampFields`/`UserFields` already?  If we encounter competing definitions, that's probably a bug.

I'd say this is low priority -- no need to work too hard to defend against that failure mode; it seems pretty unlikely that we'd reuse the field names.

tests/unit/imports/server/generateJsonSchema.ts
```diff
@@ -0,0 +1,416 @@
...
+  });
+
+  describe('number fields', function () {
+    it('respects checks on numbers', async function () {
+      const schema = z.object({
+        _id: z.string(),
+        min: z.number().gt(3).optional(),
+        max: z.number().lt(3).optional(),
+        exclusiveMin: z.number().gte(3).optional(),
+        exclusiveMax: z.number().lte(3).optional(),
```
zarvox, line 210: I think the behaviors tested here are correct for the field definitions, but I think the names for `min` and `exclusiveMin` are backwards, as are the names for `max` and `exclusiveMax`?  e.g. greater-than 3 means the value must be exclusive-minimum 3

tests/unit/imports/server/generateJsonSchema.ts
```diff
@@ -0,0 +1,416 @@
...
+        integer: z.number().int().optional(),
+        multipleOf: z.number().multipleOf(3).optional(),
+      });
+      const collection = await createTestCollection(schema);
+
+      assert.throws(() => collection.insert({ min: 3 }), /Document failed validation/);
```
zarvox, line 216: this is the epitome of nits, but I think I'd probably find the assertions here slightly more readable if they were grouped by behavior for each field rather than "accepted vs rejected", since I think it'd be easier to see the cutover points and the behavior at the boundary conditions.

zarvox:
This is all really, really good stuff!  Thanks so much for iterating on this, refining it, and generating some thorough tests.  I'm chuffed to have a more modern take on schemas and data structures that is both more flexible (e.g. discriminated unions) and more strict (e.g. distinguishing the input and output type, and forcing a re-check within the callback)!

I have read through this all and the only comment I consider blocking is making it possible to have that empty string in the `ChatMessage` text block, because (due to contenteditable selection range constraints) Slate does not allow non-inline blocks (like mentions) to start or end a Slate document, so a chat message that appears to start with a mention actually starts with an empty text block.

imports/lib/models/Model.ts
ebroder: Yeah, that's right. I haven't seen any discussion about what the expected client-side behavior is, but my assumption is that minimongo will continue to be synchronous (there's no real reason for it not to be)

imports/lib/schemas/ChatNotification.ts
ebroder: I actually kind of want to play with using [branded types](https://zod.dev/?id=brand) for foreign keys, which would allow us to capture at a type level what type of foreign key we're interacting with. (And potentially make it a type error to, e.g., pass a `Puzzle` ID when we're expecting a `Hunt` ID, or whatever)

But yeah, I think it makes sense to have a `foreignKey` helper for the time being. I'll add that.

imports/lib/schemas/customTypes.ts
ebroder: I'm fine with that, I'll make the change.

I have some vague ideas swirling around for how we should be handling answers. I agree that our prescriptiveness has gotten in our way in the last few years, but the flip side is that there's a reason for the prescriptiveness. Being able to require (e.g.) that answers match `/^[A-Z]+/` means that it's easier to do things like de-duplication and uniqueness checking, etc.

My not fully formed idea is that we should have a flag somewhere for "traditional" vs. "non-traditional" answers (terminology is a WIP), where traditional answers have to match `/^[A-Z]+/` (or at least be comparable if you upcase and strip non-alpha), and non-traditional answers are, like, literally byte strings or something. I'm still not sure entirely how to fit it together, but throwing it out in case it gets any ideas churning in your head.

imports/lib/schemas/ChatMessage.ts
ebroder: Ah, fair. I had forgotten about that quirk of slate documents. I've introduced an `allowedEmptyString` that can be used for this sort of thing.

imports/lib/schemas/User.ts
ebroder: Yeah, I think we do. I added a manual call to `validateSchema` here so we can make sure this meets whatever our standards are. (Though if we ever decided to wrap `Meteor.users` into a `Model`, we could drop it)

imports/lib/schemas/withCommon.ts
ebroder: Yeah, that's reasonable. Added a check. (I only added it on the `ZodObject` branch because if we're intersecting in a new type, we wouldn't be clobbering any existing configuration.)

tests/unit/imports/server/generateJsonSchema.ts
ebroder: Yeah, that's right. I think I originally wrote these with `z.number.min(3)` and then changed them when I realized that `.gt`/`.lt` were the only syntax for writing exclusive bounds. Fixed.

tests/unit/imports/server/generateJsonSchema.ts
ebroder: Quite reasonable 🙂. Fixed

imports/lib/schemas/Guess.ts
```diff
@@ -1,87 +1,45 @@
...
+const Guess = withCommon(z.object({
   // Denormalized in so subscriptions can filter on hunt without having to join on Puzzles
-  hunt: t.string,
+  hunt: nonEmptyString.regex(Id),
   // The puzzle this guess is for.
-  puzzle: t.string,
+  puzzle: nonEmptyString.regex(Id),
   // The text of this guess.
-  guess: t.string,
+  guess: nonEmptyString.transform(answerify),
```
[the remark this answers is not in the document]
ebroder: I think my preference for now is to preserve existing behavior, and do a broader revisit of how we handle answers later.

zarvox (approved):
Awesome work.  Let's ship it (and then follow up by migrating the rest of the schemas and getting that sweet sweet bundle size reduction)!

PR #1403, opened 2023-01-25
Remove last users of MeteorPromise

ebroder:
As of Meteor 2.9, the OAuth package supports async functions (and publications have for a while).

(Last non-test override for jolly-roger/no-sync-mongo-methods is in `imports/server/accounts.ts`, for generating email templates, which still has to be synchronous)


zarvox (approved):
LGTM

We might consider sending upstream a patch to fix `Accounts` to `await` the various email-generator functions; not sure it's on their radar

ebroder:
I started to sketch it out but it cascades pretty badly into a lot of the account email internals that I didn't feel like touching

PR #1410, opened 2023-01-28
Remove and replace Base.publish

ebroder:
So...I fell into a bit of a yakshave while working on #1394. But I think the result is pretty good.

This _started_ with not wanting to support an equivalent to `Base.publish`, both because it's kind of a messy implementation, and because I don't really believe in giving clients the ability to issue arbitrary backend database queries.

So I set out to remove the places where we were still depending on it (which, as it turned out, were not very numerous), but found that our toolkit for doing sophisticated things with Meteor publications was not quite where I wanted it...so I added some more. ✂️ 🧶 

One philosophical note: I believe that my stylistic preference is to have omnibus subscriptions that are scoped to (e.g.) a particular page or component, and which deliver all of the data for that component, rather than requiring the component to fetch multiple subscriptions. We're not there yet, but this PR moves us closer. (E.g. in the long term, we should join display names into a component's omnibus data fetch instead of having it separately call `useSubscribeDisplayNames()`)

I needed a few new helpers to make that feasible:
- I added support for array-of-string foreign key fields and for including deleted objects in `JoinPublisher`, as well as making its lifecycle a little more self-contained (and making it a function, `publishJoinedQuery`, instead of a class). I also added some tests to make sure that (at minimum) the basics were working correctly.
- I added `publishCursor`, which is basically the standard Meteor publish mechanism, but can be composed with other tools (like `publishJoinedQuery`)
- I added `PublicationMerger`, which basically wholesale steals the logic Meteor uses to merge multiple independent subscriptions to instead merge multiple independent queries within a single subscription.
- Once I had `PublicationMerger` and `publishCursor`, I realized that composing those two primitives was extremely close to `SwappableCursorPublisher`, so I went the extra mile to replace the latter with the former. (The initial missing functionality was mostly the ability to remove a sub-subscription from `PublicationMerger` without spinning down the entire machinery)

And finally, I replaced the `'mongo.*'` publications from `Base.publish` with new `TypedPublication`s, attempting as best as I could to consolidate the functionality into my preferred omnibus-style subscriptions.

For review, this should review reasonably well commit-by-commit. A few callouts on specific commits:

- *Add some basic tests to publishJoinedQuery* (268736d04035e55f9affc4d0f72981e2fbc39cd9 if I don't rebase): let me know if you have ideas for additional tests that would increase confidence. I skipped testing lingering because I didn't want to wire up clock stubs, but if you feel it's important I can have a go.
- *Add a utility for merging overlapping publications* (dfcb70a1156b5f71cfc4f5994766556d7ef0e501): Like I said, the implementation is more or less wholesale lifted from [Meteor upstream](https://github.com/meteor/meteor/blob/devel/packages/ddp-server/livedata_server.js), if you want a reference.
- *Replace Base.publish with TypedPublications* (fafb1c8de534e60cde7dab0e60a854ba7d4a5151): Things that may be particularly worth verifying - that I didn't drop models entirely, and that I carried over when queries pulled deleted records (note that in a few places I consolidated where we had two "non-deleted" + "deleted" queries in favor of a single "allowing deleted" one.)
  
  Additionally, I did run into a little trouble with `PuzzlePage`, because previously we evaluated it as two different partially-overlapping subsets of subscriptions. I'll copy what I wrote in the commit message:
  
  > Most of these are comparatively straightforward; the most complicated was on PuzzlePage, where we previously recombined multiple subscriptions into multiple loading states without any being a clean subset of the other. However, the only reason for crossover was to determine if the puzzle had been deleted (so that we could disable chat and voice). Instead, I restructured the code to _assume_ that the puzzle was deleted until we confirmed that it wasn't, which removed the explicit dependency on loading state.


imports/server/publications/pendingGuessesForSelf.ts
```diff
@@ -49,7 +49,7 @@ class PendingGuessWatcher {
 
         Object.entries(roles).forEach(([huntId, huntRoles]) => {
           if (huntId === GLOBAL_SCOPE || !huntRoles.includes('operator')) return;
-          this.huntGuessWatchers[huntId] ||= new JoinPublisher(this.sub, huntGuessSpec, { state: 'pending', hunt: huntId });
+          this.huntGuessWatchers[huntId] ||= publishJoinedQuery(this.sub, huntGuessSpec, { state: 'pending', hunt: huntId });
```
zarvox, line 52: Hmmm, this might be a little pedantic, and I don't think it will actually negatively impact anything currently, but this doesn't seem quite right -- this is going to mark the whole `pendingGuessesForSelf` sub as `ready()` as soon as a single `Hunt` worth of guess watchers has run, which is too soon.

I think it'd be more consistent for readiness to be signaled at the top-level of the subscription -- it's a single bit that's associated with the subscription as a whole, and the top-level is the only place where there's sufficient context to know if this is the only `publishJoinedQuery` that will be run.

imports/server/publishJoinedQuery.ts
```diff
@@ -247,46 +247,47 @@ const addObservers = (
...
+    },
+    removed: (id) => {
+      if (spec.lingerTime !== undefined) {
+        Meteor.setTimeout(() => { observer.decref(id); }, spec.lingerTime);
+      } else {
+        observer.decref(id);
+      }
+    },
+  });
+  sub.onStop(() => stop());
```
zarvox, line 322: Since this will now be called on subscription shutdown, as well as manually, that means that when we call `stop` manually, the sub infrastructure will also call it a second time when the client unsubs.  We should probably make `stop()` explicitly tolerate being called multiple times (and do nothing after the first invocation).

tests/unit/imports/server/publishJoinedQuery.ts
```diff
@@ -0,0 +1,174 @@
...
+      throw new Error(`Collection ${collection} not found`);
+    }
+    if (!collectionMap.delete(id)) {
+      throw new Error(`Document ${id} not found in collection ${collection}`);
+    }
+  }
+
+  ready() { /* noop */ }
+
+  stop() { /* noop */ }
```
zarvox, line 59: Arguably this should call each of the `stopHooks`?  Then maybe the `after()` calls would just call `stop()` here.

tests/unit/imports/server/publishJoinedQuery.ts
```diff
@@ -0,0 +1,174 @@
...
+        field: 'tags',
+        join: {
+          model: Tags,
+        },
+      }],
+    }, { _id: puzzleId });
+
+    await Puzzles.updateAsync(puzzleId, { $set: { tags: newTagIds } });
+    // no better way to wait for update to propagate than sleeping
+    await new Promise<void>((r) => { Meteor.setTimeout(r, 50); });
```
zarvox, line 164: If this winds up being flaky in CI, a possibly-more-resilient approach might be to poll every 50ms until the condition you wish to observe holds, up to a (much higher than 50ms) timeout.  Happy to cross my fingers for now, but my experience is that losing races tends to be pretty endemic when using shared infrastructure (or in general, where you don't have control over work-scheduling).

imports/server/PublicationMerger.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  constructor(merger: PublicationMerger) {
+    this.merger = merger;
+    this.connection = merger.mainSub.connection;
+    this.userId = merger.mainSub.userId;
+  }
+
+  ready!: () => void;
+
+  readyPromise: Promise<void> = new Promise<void>((resolve) => {
```
zarvox, line 173: This appears to only work with the most-recently-created `readyPromise`.  If that's what you want, maybe document that behavior, but I'd find it more well-behaved if `ready()` armed *all* the `readyPromise`s that have been created so far.

Also: if I'm reading this right, this appears to only work if you create the Promise before `ready()` gets called?  I'd have expected a later call to `readyPromise` to get a promise that is immediately resolved if the sub is ready already, rather than one that hangs potentially indefinitely.

imports/server/publications/pendingGuessesForSelf.ts
ebroder: Mmm yeah, I think you're right. This should actually be using `PublicationMerger`

imports/server/PublicationMerger.ts
ebroder: This is an initializer, so the promise gets created once when the class is instantiated.

zarvox: Oh.  Right, sorry for the noise 😅

imports/server/publishJoinedQuery.ts
ebroder: Now that `pendingGuessesForSelf` doesn't need a handle to the shutdown function, we can actually ensure that it only gets called once, so I think we can side-step this specific issue.

tests/unit/imports/server/publishJoinedQuery.ts
ebroder: Yeah, reasonable. Fixed.

tests/unit/imports/server/publishJoinedQuery.ts
ebroder: This prompted me to think about how to word this in a way that doesn't require a timeout, and I realized that since this is an entirely server-side test, I could just set my own `observeChanges` watcher and block until that fires (which indicates that the change has propagated through the oplog)

imports/server/publications/puzzlesForPuzzleList.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    publishJoinedQuery(this, {
+      model: Puzzles,
+      allowDeleted: includeDeleted,
+      foreignKeys: [{
+        field: 'tags',
+        join: {
+          model: Tags,
+        },
+      }],
```
ebroder, line 36: Oh, we actually need to make sure we're publishing all tags anywhere that we allow editing the tag list.

zarvox (approved):
Seems neat on the whole!

As a non-blocking comment: occasionally I've done `Meteor.subscribe("mongo.chatmessages")` (or similarly with other collections) from the browser's JS console to either:

1. explore how much time it takes to load all the data for a hunt in prod,
2. perform generic ad-hoc queries in the JS console, or
3. extract the data so I can iterate on some analysis locally with a file on a filesystem

While most of these can (at least now) be addressed by visiting other places in the site that perform similar subscribes, maybe it's worth thinking about how to support some of the ad-hoc DB-access workflows.  Maybe we can add an admin DB explorer page that has an option to just fetch everything ever?

ebroder:
> Maybe we can add an admin DB explorer page that has an option to just fetch everything ever?

Yeah, I think something like that could be useful. Potential options include a `debug` subscription that ships everything (sort of similar to `mediasoup:debug` today) or a Django-style DB admin UI (probably more useful; could support edits as well as views; more work to build).

I don't feel a ton of sadness about losing some of the access we have today, though, because my experience recently has been that it's kind of spotty and not always what you want (e.g. some of the publications that would silently only publish partial data). I've in general been leaning a lot more on `meteor mongo` locally and the MongoDB Atlas web console in prod for data investigations. (You should have a login for the latter, for the record)

ebroder:
> Yeah, I think something like that could be useful. Potential options include a `debug` subscription that ships everything (sort of similar to `mediasoup:debug` today) or a Django-style DB admin UI (probably more useful; could support edits as well as views; more work to build).

Oh, although I'll note that either of those is much more straightforward and less toilsome once #1394 lands and we have a registry of all models that we've instantiated.

PR #1426, opened 2023-02-02
Convert remaining schemas to Zod

ebroder:
I tried to pull everything that wasn't rote "switch to Zod" changes into separate commits for ease of review.

A few specific oddities that I noticed while I was doing the migration:
- `UploadTokens` was previously using `Base`, but the schema declaration didn't include `BaseSchema`. I decided to add `withCommon` to the schema, so it's now the equivalent of a full `Base` model under io-ts/SimpleSchema. Since `UploadToken`s are short lived, there shouldn't be any issues with historical data.
- Previously, we used a load-time transformation function on the `Mongo.Collection` for `Puzzles` to unique-ify tags coming out of the database. I decided to replace that with the `uniqueItems` constraint in JSON schema. We shouldn't fall afoul of that in normal usage, because we always add new tags with `$addToSet` (or by `$set`ing the entire array to a unique set), but we do have some puzzles with duplicate tags in prod, so I added a migration to do a one-time cleanup of historical state.
- Both `createHunt` and `updateHunt` accept basically all non-common fields on `Hunt` as a parameter, and we previously expressed that as `Omit<HuntType, keyof BaseType>`. That's a little harder to express now, so I instead exported `EditableHuntType`, which represents the fields unique to `Hunt` (i.e. not `_id`, `deleted`, `{created,updated}{At,By}`)
- For reasons that I don't entirely understand, the typechecker was very unhappy with the locking code living inside of the model class declaration, so I decided that the easiest thing to do was to pull `withLock` out into a separate file, which seemed to solve the issue.

Otherwise, if you're in the mood to do a careful review, there are 3 specific things worth double-checking on each of the conversions:
- In the schema file, if we previously were using `inheritSchema`, we should now be using `withCommon`
- In the model file, if we previously were using `Base`, we should now be using `SoftDeletedModel`
- In the model file, if we previously were using `Base`, we should now have the `jr_` prefix on the collection name

I haven't run into any bugs in casual testing (and the test suite hasn't either), but I'm sure they're out there. My tentative plan is to follow up this PR with adding test coverage for Meteor methods, since I think that's our most porous interface boundary, and it's fairly easy to enumerate over them


imports/lib/schemas/customTypes.ts
```diff
@@ -38,7 +38,7 @@ export const allowedEmptyString = z.string();
 // always return values (e.g. createdTimestamp will always be a Date, it'll just
 // get set on insertion and not on subsequent updates).
 
-export const stringId = z.string().regex(Id).optional()
+export const stringId = nonEmptyString.optional()
   .transform((v) => v as unknown as string);
 attachCustomJsonSchema(stringId, { bsonType: 'string', pattern: Id.source }, true);
```
zarvox, line 43: Seems to me like this `attachCustomJsonSchema` call will still attempt to enforce that the id matches `Id` at the DB layer, which going by the commit message, is not what you were aiming for?

imports/lib/schemas/customTypes.ts
```diff
@@ -38,7 +38,7 @@ export const allowedEmptyString = z.string();
 // always return values (e.g. createdTimestamp will always be a Date, it'll just
 // get set on insertion and not on subsequent updates).
 
-export const stringId = z.string().regex(Id).optional()
+export const stringId = nonEmptyString.optional()
```
zarvox, line 41: I wonder if it makes sense to have `Model` continue assuming & enforcing Meteor-style `_id` keys, but to also have an `ExplicitlyKeyedModel` where the `_id` field is not optional.  Then we could enforce the "does `_id` look like a `Random.id()`" restriction in the relevant places for the majority of collections, while allowing things like `BlobMappings` or singleton collections to demand the caller specify the `_id` everywhere we want to interact with the collection.

Thoughts?

ebroder: Yeah, I had been kind of thinking about this, because it would be nice if (e.g.) `DriveActivityLatest` had its `_id` type declared as a `z.literal`, since it's a fixed value, and I came up with a way to do it. Pushing the new code in a sec, but the basic idea is that `stringId` remains a strict `z.string().regex(Id)`, but instead allow manually passing in a different schema for `_id` to the `Model` constructor.

imports/server/migrations/48-make-puzzle-tags-unique.ts
```diff
@@ -0,0 +1,16 @@
+import Puzzles from '../../lib/models/Puzzles';
+import Migrations from './Migrations';
+
+Migrations.add({
+  version: 48,
+  name: 'Make puzzle tags unique',
+  async up() {
+    for (const puzzle of Puzzles.find()) {
+      await Puzzles.updateAsync(puzzle._id, {
```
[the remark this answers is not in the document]
ebroder: Sure, fixed.

imports/lib/models/DiscordRoleGrants.ts
```diff
@@ -1,6 +1,8 @@
-import type { DiscordRoleGrantType } from '../schemas/DiscordRoleGrant';
-import Base from './Base';
+import DiscordRoleGrant from '../schemas/DiscordRoleGrant';
+import type { ModelType } from './Model';
+import Model from './Model';
 
-const DiscordRoleGrants = new Base<DiscordRoleGrantType>('discord_role_grants');
+const DiscordRoleGrants = new Model('jr_discord_role_grants', DiscordRoleGrant);
```
zarvox, line 5: This was previously a child of `Base`, which I think means the direct migration should be to a `SoftDeletedModel` here, rather than just `Model`?

ebroder: Yep, good catch. Fixed

zarvox (approved):

zarvox (commented):
Actually hold up -- I think we likely need to rewrite some `validate: false`, `clean: false` stuff from some of our older migrations, and to use the new `bypassSchema` bits?

imports/lib/schemas/mediasoup/CallHistory.ts
```diff
@@ -1,28 +1,12 @@
...
 // manipulated by users, so many of the fields don't make sense
-const CallHistoryCodec = t.type({
-  _id: t.string,
-  hunt: t.string,
-  call: t.string,
-  lastActivity: t.union([date, t.undefined]),
+const CallHistory = z.object({
+  hunt: foreignKey,
+  call: foreignKey,
+  lastActivity: z.date().optional(),
```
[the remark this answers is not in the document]
ebroder: Hmm. Yeah, I agree - I think it should be required (and the current prod database contents agree)

imports/lib/schemas/Puzzle.ts
```diff
@@ -1,60 +1,23 @@
...
-export type PuzzleType = t.TypeOf<typeof PuzzleCodec>;
-
-const Puzzle = buildSchema(PuzzleCodec, PuzzleOverrides);
+const Puzzle = withCommon(z.object({
+  hunt: foreignKey,
+  tags: tagList,
+  title: nonEmptyString,
+  url: z.string().url().optional(),
+  answers: answer.array(),
+  expectedAnswerCount: z.number().nonnegative(),
```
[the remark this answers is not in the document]
ebroder: They're very chainable. Pushed a fix

imports/lib/schemas/mediasoup/MonitorConnectRequest.ts
```diff
@@ -1,50 +1,17 @@
...
-  producerId: t.string,
-  producerSctpStreamParameters: t.union([t.undefined, t.string]), /* JSON-serialized if present */
-  producerLabel: t.union([t.undefined, t.string]),
-  producerProtocol: t.union([t.undefined, t.string]),
+const MonitorConnectRequest = z.object({
+  initiatingServer: foreignKey,
+  receivingServer: foreignKey,
+  transportId: z.string().uuid(),
+  ip: nonEmptyString,
+  port: z.number().positive(),
```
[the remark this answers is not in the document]
ebroder: Yeah, reasonable. Pulled into a `portNumber` helper and added the extra constraints.

zarvox (approved):
Very nice!

PR #1427, opened 2023-02-03
Switch Mongo indexes to be declarative

ebroder:
Instead of creating indexes in migrations, declare them as part of the model. Then, at startup, iterate through all declared indexes and indexes actually present in the database and reconcile them.

### DIFF
--- a/imports/lib/models/Model.ts
+++ b/imports/lib/models/Model.ts
@@ -1,5 +1,7 @@
 import { Mongo, MongoInternals } from 'meteor/mongo';
-import type { Document, IndexSpecification, CreateIndexesOptions } from 'mongodb';
+import type {
+  Document, IndexDirection, IndexSpecification, CreateIndexesOptions,
+} from 'mongodb';
 import { z } from 'zod';
 import {
   IsInsert, IsUpdate, IsUpsert, stringId,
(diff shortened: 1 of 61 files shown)


imports/lib/models/Model.ts
```diff
@@ -280,6 +282,70 @@ export function formatValidationError(error: unknown) {
...
+  } else if (indexIsObject(index)) {
+    return Object.entries(index);
+  } else if (indexIsMap(index)) {
+    return [...index];
+  } else {
+    return index.map(normalizeIndexSpecification).reduce((acc, val) => acc.concat(val), []);
+  }
+}
+
+export function normalizeIndexOptions(options: CreateIndexesOptions): NormalizedIndexOptions {
```
zarvox, line 341: Rather than filter out unsupported options here, I'd prefer you threw an exception to make it loud and clear that the option is unsupported.

zarvox: Ahhh, but now that I'm reading the part that retrieves current indexes from the DB, I see that you have to handle whatever is in the database.  Maybe have a flag or a separate function for the two behaviors, so as to explicitly disallow unsupported `addIndex` calls, but be tolerant of whatever's in the DB?

zarvox: I guess the type declaration prevents incorrect usage of `addIndex` already.  Never mind!

zarvox (approved):
One thing that I'm a little wary of is how this behaves in a clustered environment with potentially out-of-date nodes.  If we are running a heterogeneous cluster of app versions, then it's possible that if one of them were to restart, we'd revert indexes from e.g. 2023-02-02's release to 2023-01-31, and last-writer wins, whereas with migrations the older code would see that the migration marker is past what it wanted to set it to, and thoughtfully takes no action.

Is there a way we can avoid a mix of V0131 and V0202 potentially fighting for the DB state?  Or do you see that as more of a desired outcome (where rolling back to an older version creates/drops indexes to match what that version expected)?

I guess indexes are "just" a performance optimization, and we generally don't change them all that much, so it's probably okay?

ebroder:
Hmm. We do depend on unique indexes for correctness in a few cases, but I'm not sure how to avoid the conflicting index applications here.

My gut says that it's mostly not a big deal in practice (we don't see a lot of heterogenous deployments), and that easy inspectability of indexes is worth the risk?

zarvox:
> Hmm. We do depend on unique indexes for correctness in a few cases, but I'm not sure how to avoid the conflicting index applications here.
>
> My gut says that it's mostly not a big deal in practice (we don't see a lot of heterogenous deployments), and that easy inspectability of indexes is worth the risk?

I can think of a few possible approaches, any of which I'd probably be at peace with:

* Leave this PR as-is, and accept the risk.  It's only the intermediate state of "subset of the cluster attempting to take an update" that can hit the potentially-bad sequence I described, and that's generally 1) transient 2) converges to the desired system and 3) is unlikely to be a change we make during Hunt or anything like that.  We aim to minimize production changes during Hunt proper and in the off-season and while developing we generally have a node count of 1, so it's pretty unlikely to come up.  Upsides: simple, already implemented, nice developer ergonomics. Downsides: accepts theoretical risk
* Track the commit timestamp of each build in the build process.  Upon completion of applying index changes, update a record in a DB with that commit timestamp.  Only attempt to apply index changes if the build timestamp is greater than the DB timestamp (or `Meteor.isDevelopment`).  Upsides: low developer friction, avoids unintended rollbacks in production environments.  Downsides: somewhat more complicated to implement
* Reuse the migration system for "run-once" functionality: don't attempt to converge indexes on startup, but instead provide a function to do so and add a migration each time we change indexes.  (As a developer ergonomics thing, we could warn or error on startup that a migration is needed if the declared migrations do not match the ones observed in the DB.)  Upsides: avoids rollbacks.  Downsides: probably still a pain to iterate on index changes locally.  Potentially easy to miss that a new "please converge again" migration is needed.

Is the primary goal of making these indexes declarative that you can iterate on them more easily without having to carefully poke at migrations?  And maybe laying the groundwork for a future with a registry of the known indexes so that we can require all queries to have indexes?

ebroder:
Hmm. I'm not sure I actually want to require indexes for queries, but I do want to make it easier to tell if a query is well-covered. Today if you're writing a query and want to tell if there's an index to cover it, you either need to inspect the mongo shell or walk through all of the migrations (and keep track of which indexes were subsequently dropped). I think that making that better is my primary goal. Ease of development is definitely a secondary goal, though.

I think the second option you propose is reasonable, although I think I'd rather use build timestamp instead of commit timestamp, since commits can potentially be weird, but we do (I think) always want the most recent build to win. In which case, I'm not even sure we need a new build step to capture the timestamp; I think we can instead just look at the lastmod time of some file in the build output - something like:

```typescript
import { promises as fs } from 'fs'

const ts = (await fs.stat(process.argv[1])).mtime
```

zarvox:
> I think the second option you propose is reasonable, although I think I'd rather use build timestamp instead of commit timestamp, since commits can potentially be weird

My instinct from reproducible builds was to use something that's actually tracked in the repo rather than from the build environment and corresponds to "when did we decide this was a thing that we wanted to exist" rather than "when did the infrastructure catch up".  I'm not sure if GHA ever runs branch builds (e.g. for `main`) in parallel, but if you can imagine them completing in a different order than the branch order, then that's a potential hazard (it probably still would be for tagging purposes, which is orthogonal, but w/e).  Determinism in built artifacts is handy for ensuring predictable and consistent behavior.  If we use the git committer timestamp (rather than the author timestamp), I believe we should expect a branch to have monotonically increasing timestamps.

All that said, your proposal for a build timestamp is probably also fine for our purposes.

ebroder:
One specific thing I like about build timestamps is that they work in both development and production, and they work regardless of the mechanism you're using to run the Meteor app in production (i.e. if you're running the app but not using our Docker build), so we don't need to special-case anything based on environment. I like that that means we get to exercise the mechanism in development (even if we expect it to allow through all changes)

zarvox:
I'm convinced.  Let's do build timestamps.

PR #1428, opened 2023-02-03
Fix modifiers for arrays with defaults

ebroder:
Even if a schema has a default, we still need to relax the inner type to make sure we handle arrays correctly.


zarvox (approved):
This fixes the `addTag` method.

PR #1432, opened 2023-02-04
Migrate historical ChatMessage data to V2 and remove chat v1

zarvox:
This will result in some differences in rendering of historical data, but we broadly consider this acceptable, as we believe the new format is better-behaved for most practical purposes, and we expect vanishingly few people to care deeply about the precise preservation of old formatting of the chat messages from more than a year ago.  Also, we've broken chat formatting in the past when we switched from Slack-ish formatting to slightly more-standard GitHub Flavored Markdown.


ebroder (approved):
Seems believable to me. Can I request a followup to rename away anything that still has "v2" in the name?

zarvox:
> Can I request a followup to rename away anything that still has "v2" in the name?

Will do.

PR #1434, opened 2023-02-04
Change all updated timestamps to last-write timestamps

ebroder:
This cleans up an odd previous behavior around updated timestamps and upserts. Previously, an upsert which updated an existing document wouldn't update an updatedTimestamp (or our standard definition of updatedAt with SimpleSchema).


imports/client/components/NotificationCenter.tsx
```diff
@@ -237,8 +237,8 @@ const GuessMessage = React.memo(({
             className="ms-3"
             width={16}
             height={16}
-            startTime={guess.updatedAt!.getTime()}
-            endTime={guess.updatedAt!.getTime() + LINGER_PERIOD}
+            startTime={guess.updatedAt.getTime()}
```
zarvox, line 240: This throws a red flag for me -- if the type system is claiming that `updatedAt` is always present, then we probably need a migration to go backfill the value (from `createdAt`?) onto all records of all types that don't have one yet.

zarvox (changes requested):

imports/client/components/NotificationCenter.tsx
ebroder: Yeah, that's reasonable. Added a migration to iterate over everything and backfill the `updatedAt` property.

zarvox (approved):

PR #1436, opened 2023-02-04
Use permission stubs to simplify pendingGuessesForSelf

ebroder:
By changing our permission stubs to accept the relevant partials of users and hunts, we can take advantage of those stubs in places where we don't have complete objects, such as pendingGuessesForSelf, where we're not guaranteed to see a full object (but where any update _will_ have the data we need to use the stub functions).

(This came up while I was doing some unrelated experimentation, but I think it's a good improvement independent of that)


zarvox (approved):

zarvox:
This might need a rebase to pick up the fix from #1437

PR #1438, opened 2023-02-05
Experiment with using branded types for hunt IDs

ebroder:
Types in TypeScript are structural by default, meaning that an object of one type can be assigned to another if those types have the same shape. Branded types are a form of nominal typing, meaning that an object can't be assigned to a variable of a branded type unless it is explicitly the same exact type.

This is useful for types which look the same but should be treated differently - for instance, a hunt ID and a puzzle ID. They're both strings, but they're not the same string.

This switches hunt IDs throughout the database to be branded, using Zod's built-in support for branded types. Zod's branding requires a brand name; we use the collection name.

This also introduces an assortment of helpers to deal with branded types:
* `ForeignKeyType<M extends Model>`: The branded type of a foreign key to a model. This is used in place of `string` for foreign keys. (Most of the changes in this commit are to replace `string` with `ForeignKeyType<typeof Hunts>`.)
* `makeForeignKey<M extends Model>(id: string)`: A helper to create a foreign key of the correct type. This is used to "encode" a string which we know to be a valid foreign key, and should be used very sparingly. In practice, we only need this in tests and for the hard-coded fixture hunt ID.
* `matchForeignKey<M extends Model>()`: Returns a matcher which can be passed to `check` to assert that a value is a foreign key. This is actually equivalent to just verifying that the value is any string (because we can't actually check that a string is a valid foreign key without querying the database), but it's necessary for our handling of typed methods and publications.

Note that none of this should introduce any runtime overhead; type brands exist at the type layer only, and do not have any runtime footprint (other than values passing through a few extra identity functions).

### DIFF
--- a/imports/lib/models/Model.ts
+++ b/imports/lib/models/Model.ts
@@ -1,3 +1,4 @@
+import { Match } from 'meteor/check';
 import { Mongo, MongoInternals } from 'meteor/mongo';
 import type {
   Document, IndexDirection, IndexSpecification, CreateIndexesOptions,
@@ -9,7 +10,7 @@ import {
 import type { MongoRecordZodType } from './generateJsonSchema';
 import validateSchema from './validateSchema';
 
-export type Selector<T extends Document> = Mongo.Selector<T> | string | Mongo.ObjectID;
+export type Selector<T extends Document> = Mongo.Selector<T> | T['_id'];
 export type SelectorToResultType<T extends Document, S extends Selector<T>> =
   S extends string ? T & { _id: S } :
   S extends Mongo.ObjectID ? T & { _id: S } :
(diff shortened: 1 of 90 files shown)


imports/lib/models/Model.ts
```diff
@@ -525,6 +530,20 @@ class Model<Schema extends MongoRecordZodType, IdSchema extends z.ZodTypeAny = t
...
 
-export type ModelType<M extends Model<any, any>> = z.output<M['schema']>;
+export type ModelType<M extends Model<any, any, any>> = z.output<M['schema']>;
+
+export type ForeignKeyType<M extends Model<any, any, any>> = string & z.BRAND<M['name']>;
+export const makeForeignKey = <M extends Model<any, any, any>>(id: string): ForeignKeyType<M> => {
+  return id as ForeignKeyType<M>;
+};
+
+export const matchForeignKey = <
```
zarvox, line 540: This is a function that returns a matcher function, not the matcher function itself, so I'd call this makeForeignKeyMatcher<M>() or something so it's more apparent that you're supposed to invoke it in your `match()` call when you wouldn't do the same for e.g. `String`

ebroder:
FWIW I'm going to go ahead and close this - I don't think there's enough value here to be worth the conversion effort. We can always reevaluate if we find ourselves writing foreign key confusion bugs.

PR #1442, opened 2023-02-05
Make heartbeats less reliable and hopefully more robust

ebroder:
Mediasoup's default configuration attempts to make SCTP delivery ordered and reliable. However, for our heartbeat system, which sends a new message every 100ms, that's not actually what we want - if one heartbeat gets dropped, we don't want to wait for retransmission of that heartbeat before delivering subsequent ones to the receiving application.

Setting ordered: false will allow packets to be delivered out of order, and setting maxRetransmits: 1 should allow the sending side to be more aggressive about dropping missed packets, instead of trying to retransmit them.

### DIFF
--- a/imports/server/mediasoup.ts
+++ b/imports/server/mediasoup.ts
@@ -122,9 +122,7 @@ class Watchdog extends EventEmitter {
 
   resetTimeout() {
     this.lastActivity = new Date();
-    if (this.timeoutHandle) {
-      clearTimeout(this.timeoutHandle);
-    }
+    Meteor.clearTimeout(this.timeoutHandle);
     this.timeoutHandle = Meteor.setTimeout(() => {
       this.emit('timeout');
     }, this.timeout);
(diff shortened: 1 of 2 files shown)


ebroder:
For the record, I give this about a 65% chance of fixing the issues we're seeing in prod. I think the odds are good but not perfect.

imports/server/mediasoup.ts
```diff
@@ -122,9 +122,7 @@ class Watchdog extends EventEmitter {
 
   resetTimeout() {
     this.lastActivity = new Date();
-    if (this.timeoutHandle) {
-      clearTimeout(this.timeoutHandle);
-    }
+    Meteor.clearTimeout(this.timeoutHandle);
```
zarvox, line 125: Are we intentionally dropping the `if (this.timeoutHandle) {` guard?

ebroder: It was never necessary - `timeoutHandle` is created in the constructor and is never set to undefined.

zarvox (approved):

PR #1443, opened 2023-02-05
Log improvements

ebroder:


imports/server/configureLogger.ts
```diff
@@ -26,8 +28,15 @@ logger.format = format.combine(
     level, label, message, [userIdSymbol]: userId, error,
     ...rest
   }) => {
+    const ctx: string[] = [];
+    if (workersCount > 1) {
```
zarvox, line 32: I wouldn't be sad about including this regardless

zarvox (approved):

imports/server/configureLogger.ts
ebroder: Eh. It feels like it's just noisy in local dev

imports/server/loadBalance.ts
```diff
@@ -39,14 +39,16 @@ const getWorkersCount = () => {
...
+
 WebApp.onListening(() => {
-  const workersCount = getWorkersCount();
   if (workersCount > 0) {
     http.globalAgent.maxSockets = 99999;
     const proxy = HttpProxy.createProxyServer({
       xfwd: true,
     });
-    const workers = WorkerPool.create(workersCount);
+    workers = WorkerPool.create(workersCount);
```
zarvox, line 51: not sure why this is no longer `const workers`?

ebroder: Oh whoops, wilfix

PR #1445, opened 2023-02-05
Fix find with no arguments

ebroder:
Meteor's find method has the slightly confusing property that passing an `undefined` query explicitly behaves differently than passing no query (since the query is an optional parameter). Passing `undefined` explicitly returns an empty result set.

We want to allow the query to be optional, as with Meteor, but that resulted in us explicitly passing undefined to Meteor, which caused the empty result set behavior. So instead make sure we pass an empty query. (This means that we can no longer access the explicit-undefined behavior from Model, but I don't think we've ever wanted that anyway.)

### DIFF
--- a/imports/lib/models/Model.ts
+++ b/imports/lib/models/Model.ts
@@ -492,7 +492,7 @@ class Model<Schema extends MongoRecordZodType, IdSchema extends z.ZodTypeAny = t
     selector?: S,
     options?: O,
   ) {
-    return this.collection.find(selector, options) as
+    return this.collection.find(selector ?? {}, options) as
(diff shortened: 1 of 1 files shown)


ebroder:
As a note, this only came up on `Model` classes - not `SoftDeletedModel`, since those always generated a rewritten query (except for `findAllowingDeleted`, which could trigger this behavior)

zarvox (approved):

zarvox:
I have no particular attachment to preserving the behavior of `find()` with no args, and I would also find requiring at least the selector be passed to find like `find({})` to be a perfectly acceptable solution here.

PR #1488, opened 2023-02-22
Unwind Google Drive fetcher if lock is preempted

ebroder:
When we have a lock to fetch Google Drive activity, we attempt to continually renew it. But if, for some reason, that renewal fails, we should bail and attempt to reacquire the lock. Previously, we would blithely go on about our business and continue trying to renew the lock that we had already lost, resulting in both two processes attempting to fetch activity history and a lot of error noise from the failed renewals.


ebroder:
(This can be seen in production error logs at https://app.bugsnag.com/death-and-mayhem/jolly-roger/errors/63f4eab5d594900008c319a5)

zarvox (approved):

PR #1498, opened 2023-02-24
Use createPortal for modal popups

### LINKED ISSUE #1317: Migrate modals to use React portals
We've had problems in the past with where modals insert into the DOM (e.g. activating tooltips for the element that cause them to pop up, like in #1267). [React Portals](https://reactjs.org/docs/portals.html) are an easy solution to this - we can just hoist the modal up to the top-level of the DOM.

We should go through and migrate our modals to live inside portals.

ebroder:
React's new-ish createPortal allows you to place a component outside of its rendering tree, which makes sense for modals which are effectively untethered from where they would appear by default in the DOM.

This is a more stable fix for problems we've had in the past with, e.g., overlay triggers, where the modal is inside of the overlay trigger and thus triggers the tooltip.

For some reason, the structural changes to the components which used forwardRef caused the react/prop-types eslint rule to start complaining. Since it's not really useful in a TypeScript world, I just turned it off.

Fixes #1317.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -115,6 +115,11 @@
     // default prop values of `undefined`
     "react/require-default-props": "off",
 
+    // This seems to interact poorly with HOCs like forwardRef. TypeScript means
+    // that we don't really have to worry about missing types on our component
+    // props
+    "react/prop-types": "off",
+
     // We use spread props a lot for HOCs or other wrapper-type components
     "react/jsx-props-no-spreading": "off",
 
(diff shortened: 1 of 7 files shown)


zarvox (approved):
In my testing, I found that the "We've left your mic muted for now given the number of people on the call. You can unmute yourself at any time." popover appears to jump between the top-left corner and the middle of the screen a couple times.

However, it also appears to do that with the current git `main`, so I guess that's not a regression in this change.  Probably the ref that we're trying to point at is getting transiently nulled out or something?

Issue #1279, opened 2023-01-03
Collect documentation on more complex inner workings


ebroder:
Jolly Roger has accumulated a fair amount of complex systems with subtle interdependencies split across a lot of files which are mostly undocumented. It would be good if we had some documentation for how those things work.

As a proposal, I think we should have a `docs` folder with a collection of Markdown design docs for some of these systems. I'm cognizant that it'd be pretty easy for these docs to get out of date, and I'm not sure I have a solution for that, but my intuition is that something would be better than nothing.

(One strawman idea: each documentation file has a front matter with a structured list of source code files that it references + a "last updated" revision. We write a GitHub Action that fails such that if those referenced source files are modified but the documentation file's last updated revision is not.)

Here are some potential areas that I think might benefit from documentation:

- [ ] The handshake protocol for Mediasoup calls between the client and server and between servers (including API methods, database records, etc.)
- [ ] Google Drive document and permissions management
- [ ] The custom asset pipeline

PR #1513, opened 2023-03-03
Add a semi-simple documentation system

ebroder:
Some higher-level information about a codebase are hard to document inline in code. For systems that span multiple files, functions, etc., it's useful to have a more centralized source of documentation. However, the inherent risk of any such centralized documentation is that it quickly becomes out of date.

Introduce a `docs` folder, containing such higher-level documentation. However, to mitigate against docs becoming out of date, additionally introduce some validation in the form of a new, homegrown linter. This linter requires that some metadata be included with each documentation file: specifically, which files the documentation covers and when the documentation was last updated. If the files are ever updated without a corresponding update to the docs, the linter will fail.

Here's what an error message looks like if a documentation file is out of date:

```
> jolly-roger@0.0.1 lint:docs /Users/evan/src/jolly-roger
> ts-node-esm tests/check_docs.mts

Documentation errors:

  assets.md is out of date
    imports/lib/models/BlobMappings.ts was updated in 4a64684c3dfe9cbd94ff30a018541f0c430c8b6b
    imports/server/models/Blobs.ts was updated in 4a64684c3dfe9cbd94ff30a018541f0c430c8b6b
  To fix this, either update assets.md or, if no changes are needed, change the updated field to match the most recent commit to those files.
```

This introduces the system for #1279, although I'm not going to claim it fixes it until we have a bit more of a library of documentation.

(As an aside, I experimented with using a third-party service - swimm.io - which does something similar to my linter. I ultimately rejected it in favor of this approach because (a) their documentation files are committed to the repo, but in a `.swm` hidden directory making them less discoverable (b) their files can't be edited directly but have to be edited through their web app, limiting ability of others to contribute (c) it's not entirely clear how their detection of out-of-date documentation works and thus how likely it would be to false-negative when updates are necessary. You can see what their committed docs look like on [`evan/custom-asset-docs`](https://github.com/deathandmayhem/jolly-roger/compare/evan/custom-asset-docs?expand=1) if you're curious.)

### DIFF
--- a/tests/check_docs.mts
+++ b/tests/check_docs.mts
@@ -1,0 +1,170 @@
+/**
+ * Require that documentation is kept up to date as the referenced files change.
+ *
+ * To do this, for each file in the docs directory, we look at the set of files
+ * listed in the front matter and check when each of them last changed. If any
+ * of them is more recent than the last commit to the doc file, then the doc is
+ * out of date. (If the files have changed but the documentation doesn't require
+ * changing, then you can update the "updated" date in the documentation, which
+ * will bring the latest commit to the doc up to date with the changes to the
+ * files it documents)
+ */
+import child from "node:child_process";
+import { promises as fs } from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { promisify } from "node:util";
+import matter from "gray-matter";
+
+const execFile = promisify(child.execFile);
+
+const filename = fileURLToPath(import.meta.url);
+const dirname = path.dirname(filename);
+
+interface NewerFile {
+  file: string;
+  commit: string;
+}
+
+type DocError =
+  | {
+      type: "out-of-date";
+      doc: string;
+      newerFiles: NewerFile[];
+    }
+  | {
+      type: "missing-front-matter";
+      doc: string;
+    }
+  | {
+      type: "missing-updated-field";
+      doc: string;
+    }
+  | {
+      type: "missing-files";
+      doc: string;
+    };
+
+const checkDoc = async (doc: string): Promise<DocError | undefined> => {
+  const docMatter = matter(await fs.readFile(doc));
+
+  if (!docMatter.data.updated) {
+    return {
+      type: "missing-updated-field",
+      doc,
+    };
+  }
+
+  // First: figure out how recent the documentation is by looking at the last
+  // commit to the file.
+  const { stdout: fileRevision } = await execFile("git", [
+    "rev-list",
+    "-1",
+    "HEAD",
+    "--",
+    doc,
+  ]);
+  const docRevision = fileRevision.trim();
+
+  // If the doc file has no revision, then it's probably a new file being
+  // written, which is fine (definitionally newer than antyhing that's been
+  // committed)
+  if (!docRevision) return undefined;
+
+  // Next: figure out if any of the referenced files are more recent than our
+  // doc revision
+  const files: string[] = docMatter.data.files;
+  if (
+    !files ||
+    !Array.isArray(files) ||
+    !files.every((file) => typeof file === "string")
+  ) {
+    return {
+      type: "missing-files",
+      doc,
+    };
+  }
+
+  const fileChecks = await Promise.all(
+    files.map(async (file) => {
+      const { stdout: updated } = await execFile("git", [
+        "rev-list",
+        "-1",
+        "HEAD",
+        `^${docRevision}`,
+        "--",
+        file,
+      ]);
+      if (updated.trim().length > 0) {
+        return { file, commit: updated.trim() };
+      }
+
+      return undefined;
+    }),
+  );
+  const newerFiles = fileChecks.filter(
+    (fc): fc is NewerFile => fc !== undefined,
+  );
+  if (newerFiles.length > 0) {
+    return {
+      type: "out-of-date",
+      doc,
+      newerFiles,
+    };
+  }
+
+  return undefined;
+};
+
+const main = async () => {
+  const docsDir = path.join(dirname, "..", "docs");
+  const docs = await fs.readdir(docsDir);
+
+  const results = await Promise.all(
+    docs.map((doc) => checkDoc(path.join(docsDir, doc))),
+  );
+  const errors = results.filter(
+    (result): result is DocError => result !== undefined,
+  );
+
+  if (errors.length > 0) {
+    process.stderr.write("Documentation errors:\n\n");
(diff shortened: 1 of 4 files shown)


tests/check_docs.mts
```diff
@@ -0,0 +1,130 @@
...
+    return {
+      type: 'missing-updated-field',
+      doc,
+    };
+  }
+
+  const { stdout: fileRevision } = await execFile('git', ['rev-list', '-1', 'HEAD', `^${docRevision}`, '--', doc]);
+  if (fileRevision.trim().length > 0) {
+    docRevision = fileRevision.trim();
+  }
```
ebroder, line 59: For the record, this logic is kind of weird, and I'm not sure if it's right. The commit in the `updated` frontmatter field is always going to be less recent than the most recent commit to the file itself, because :sparkles:merkle trees:sparkles:.

I do want it to be possible to flag "this documentation file wasn't changed but is still up to date", but I also would prefer to not have, like, a "generation number" in the documentation file or whatever that just gets monotonically incremented. A commit ID at least feels like it has _some_ semantic meaning.

zarvox: A commit ID has some meaning, but yeah, it definitely feels less than ergonomic, especially if the linter isn't going to tell you what the actual commit ID you should be updating the file to is.  Maybe we should have a helper script `mark-up-to-date` that does the appropriate git lookup & find/replace in the front matter?  or outputs a diff that you can pipe into `patch` or something?

Something I've seen done in other documentation systems is to have a "last-reviewed-at" date, but that's typically more oriented for process/team-level documentation rather than doc documentation (and skew between authored/committed dates during code review could be troublesome) and is more targeted at an async "audit any docs that haven't been updated in 3 months to make sure they're still accurate" flow rather than a blocking "if you touched the code, you must also touch the docs with a fresher sha1".

A thing that I've found tricky in the past is when the logic for something moves between locations, but the docs still reference the old home for that logic.  I guess if we check for changes in all the files referenced, then any time you extract logic from one of them you're modifying that one, so your scheme covers that case.  I suspect this may wind up adding a chunk of friction to refactorings, though?  I don't really have a great answer for how to get around that, though; it seems intrinsic to making the docs know about what files their content covers.

To be honest, while I certainly value having up-to-date docs, I'm cautious about making doc-freshness checks something that blocks merges.  One possible negative effect would be that it may incentivize us to bump shas blindly without actually doing meaningful review of the docs, but on the flipside: they'd probably force us to at least see them/be reminded they exist in code review, and if semantics are indeed changing, that'd be a useful prompt, and also we're generally under less schedule pressure than the other places I've seen this sort of thing break down.

ebroder: For the record, I never followed up on this because you correctly called out a thing that I was already unhappy with and I didn't have a better option. However, helping @egnor with getting JR setup has reminded me that I want to do more here (especially because _I've_ forgotten how some of these systems fit together, unsurprisingly), so I'm going to re-raise it.

I mostly agree that this type of enforcement system can be problematic, but I think it's more problematic to not have anything at all and allow the docs to drift. I'm curious to see what the actual false-negative rate on this test is, but my gut says that it's going to be relatively low. The exception will be large multi-purpose files that lead to unrelated changes (e.g. heaven forbid we document anything in `PuzzlePage.tsx`), but I think that's a good motivation to split out those files and make them more manageable?

I am unhappy with the git revision thing, though, for the reasons I mentioned above. But maybe a "last updated date" would actually work. With a last updated date, we don't actually need to look at the value of the date, we just look and see if the docs are at-least-as-fresh as the files they document.

ebroder: OK. Reworked this so that `updated` is now a date, and the actual value is ignored by the linter. If you update files that are referenced by documentation but don't need to update the documentation itself, you can bump the `updated` timestamp, which will cause the file to appear in the git history as being updated at the same time or later than the files it documents. I'm happier with this.

tests/check_docs.mts
```diff
@@ -0,0 +1,163 @@
+/**
+ * Require that documentation is kept up to date as the referenced files change.
+ *
+ * To do this, for each file in the docs directory, we look at the set of files
+ * listed in the front matter and check when each of them last changed. If any
+ * of them is more recent than either the last commit to the doc file or the
+ * commit listed in the "updated" field (whichever is more recent), then the doc
```
zarvox, line 7: > commit listed in the "updated" field

This is a datestamp now, right?

tests/check_docs.mts
```diff
@@ -0,0 +1,163 @@
...
+    }
+  | {
+      type: "missing-files";
+      doc: string;
+    };
+
+const checkDoc = async (doc: string): Promise<DocError | undefined> => {
+  const docMatter = matter(await fs.readFile(doc));
+
+  if (!docMatter.data.updated) {
```
zarvox, line 49: We don't appear to use the value of `docMatter.data.updated` anywhere, besides verifying that it exists.  I guess it's really just a field to prompt the developer to touch the file in the commit (or a later commit in the branch), so the `rev-list` logic excludes it?

tests/check_docs.mts
ebroder: Oh yeah, I'll update that

tests/check_docs.mts
```diff
@@ -0,0 +1,163 @@
...
+          );
+          for (const newerFile of error.newerFiles) {
+            process.stderr.write(
+              `    ${newerFile.file} was updated in ${newerFile.commit}\n`,
+            );
+          }
+          process.stderr.write(
+            `  To fix this, either update ${path.basename(
+              error.doc,
+            )} or, if no changes are needed, change the updated field to match the most recent commit to those files.\n`,
```
zarvox, line 139: I presume we don't actually intend to list commit IDs in the updated field any more?  Maybe update the guidance given?

tests/check_docs.mts
ebroder: Right. We could, like, verify that it is actually a timestamp if we wanted.

zarvox: Nah, I'm just verifying my understanding.

I think I'd rather leave it unchecked, to better handle the case there are two commits that update the same file in a day, which would make it hard to update the field if we verify that it's (purely) a date.

And if we made it a full date+time, that feels like it's asking each file for more specificity than is merited.  I think I'm fine leaving this as-is; just making sense of it.

zarvox (approved):

zarvox:
Looks like this might need a rebase, for `package.json`?

PR #1782, opened 2023-08-11
Update stylelint and address complaints

ebroder:
The old stylelint support for styled-components had been somewhat hacked in and didn't work well. With v15 of stylelint, the integration is much more solid. As a side effect, it also lets us lint the top-level SCSS files at the same time.

Due to a mix of the integration being more solid and other lint changes in v15, there were a handful of changes to catch up on:

- stylelint is much pickier about `//` comments - apparently they aren't valid in standard CSS, so they now tend to throw errors. This hasn't been a problem to date, but it doesn't seem like styled-components is pre-processing them away, so it's easier to just switch to `/* */` comments.

- The declaration-block-no-redundant-longhand-properties rule has become pickier about using top+bottom+left+right in a single declaration, similar to how it was previously picky about using margin-{top,left,bottom,right}. This introduced me to the `inset` property, which I hadn't seen before. I've ported the relevant styles to use it.

- Media queries now have a syntax allowing for `<` and `>`, where previously `max-width` and `min-width` were always inclusive ranges. This lets us simplify a few places where we were intentionally compensating for inability to express an open range.


imports/client/components/styling/FixedLayout.tsx
```diff
@@ -3,8 +3,9 @@ import { NavBarHeight } from './constants';
 
 export default styled.div`
   position: fixed;
-  top: calc(env(safe-area-inset-top, 0) + ${NavBarHeight});
-  bottom: 0;
-  left: env(safe-area-inset-left, 0);
-  right: env(safe-area-inset-right, 0);
+  inset: /* top right bottom left */
```
zarvox, line 6: Thank you for this comment; I will probably never manage to remember the ordering

zarvox (approved):
Today I learned about the `inset` property.

ebroder:
> Today I learned about the `inset` property.

Same, to be clear. It seems like it's new-ish - browsers started reliably supporting it in 2021 or so.

PR #1814, opened 2023-08-31
Replace try/finally blocks with new "using" syntax

ebroder:
(Note that this definitely looks best with a whitespace-insensitive diff)

Maybe not a huge improvement, but I like these for putting the cleanup at the same place as the resource acquisition.

(At some point in the future, I might put together a disposable version of `Meteor.setTimeout` / `Meteor.setInterval` since that seems to be a pattern)

### DIFF
--- a/client/main.ts
+++ b/client/main.ts
@@ -2,6 +2,9 @@
 import '../imports/client/bugsnag';
 import '../imports/client/configureLogger';
 
+// Polyfills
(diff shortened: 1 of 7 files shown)


ebroder:
Bah I thought that this worked when I tested it locally, but I think this is effectively blocked behind a Node upgrade, because we need a newer version of typescript-eslint. Oh well.

PR #1817, opened 2023-09-01
Add bookmarking for puzzles

### LINKED ISSUE #1368: Consider making it easier to return to puzzles you were working on previously
From a user in Discord post-hunt:

> This is one of those "maybe it already exists" things, but is there any kind of bookmarking/favoriting/recently-opened functionality in JR? Whenever I go to sleep, I usually just leave the tabs of the puzzles I was looking open to see how things are going, but this year I had a couple of times where my browser crashed or I had to restart my computer, and I ended up digging the puzzles out of my browser history to see if they'd been solved or still needed eyes because I couldn't remember their names. There are ways I could get around this via external bookmarking (browser bookmarking, dumping all of the puzzles I'm interested in into a markdown file, whatever) but within JR would be pretty cool.

Maybe this winds up being a separate sort order option, or a "my recent activity" view option or something?  I could believe this would work either with the server-side activity tracking data from our various activity sources, or possibly with just client-side localStorage records that we write (which would probably be able to easily consider navigations but less likely document or call activity).

ebroder:
Some of our team members have asked for a way to more easily keep track of puzzles they care about, even when they're not currently looking at them.

Bookmarked puzzles are shown in a section at the top of the puzzle list page (in addition to showing up in their normal place in the list):

Screenshot 2023-09-01 at 12 01 37 PM

When a puzzle is solved, users that have that puzzle bookmarked get a notification:

Screenshot 2023-09-01 at 12 00 23 PM

One thing that's missing right now is a UI element on the individual puzzle page for bookmarking a puzzle, as I'm not entirely sure where to put it.

@Apophenia I'm curious if this seems like it would work for your use case, since you were one of the people who requested this.

Fixes #1368.

### DIFF
--- a/imports/client/components/NotificationCenter.tsx
+++ b/imports/client/components/NotificationCenter.tsx
@@ -24,6 +24,8 @@ import isAdmin from '../../lib/isAdmin';
 import { indexedById } from '../../lib/listUtils';
 import Announcements from '../../lib/models/Announcements';
 import type { AnnouncementType } from '../../lib/models/Announcements';
+import type { BookmarkNotificationType } from '../../lib/models/BookmarkNotifications';
+import BookmarkNotifications from '../../lib/models/BookmarkNotifications';
 import type { ChatNotificationType } from '../../lib/models/ChatNotifications';
 import ChatNotifications from '../../lib/models/ChatNotifications';
 import Guesses from '../../lib/models/Guesses';
@@ -34,9 +36,11 @@ import PendingAnnouncements from '../../lib/models/PendingAnnouncements';
 import Puzzles from '../../lib/models/Puzzles';
 import type { PuzzleType } from '../../lib/models/Puzzles';
 import { huntsUserIsOperatorFor } from '../../lib/permission_stubs';
+import bookmarkNotificationsForSelf from '../../lib/publications/bookmarkNotificationsForSelf';
 import pendingAnnouncementsForSelf from '../../lib/publications/pendingAnnouncementsForSelf';
 import pendingGuessesForSelf from '../../lib/publications/pendingGuessesForSelf';
 import configureEnsureGoogleScript from '../../methods/configureEnsureGoogleScript';
+import dismissBookmarkNotification from '../../methods/dismissBookmarkNotification';
 import dismissChatNotification from '../../methods/dismissChatNotification';
 import dismissPendingAnnouncement from '../../methods/dismissPendingAnnouncement';
 import linkUserDiscordAccount from '../../methods/linkUserDiscordAccount';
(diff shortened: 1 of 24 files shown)


Apophenia:
speaking only for myself, I think this 100% achieves what I was hoping for :) (or will once there's a way to bookmark from the puzzle page)!

ebroder:
I guess the easiest option would be to throw it in with the various control buttons we already have on the page. I'm not wild about that particular section because it feels like sort of a hodge podge, but it does somewhat fit with the other things that are already there:

Screenshot 2023-09-01 at 12 55 43 PM

ebroder:
(Ignore the fact that managing Google docs is broken on my local build at the moment)

zarvox (commented):
The backend implementation all looks solid!

UI-wise: I admit this is kinda bikesheddy but I feel surprisingly strongly about it: Can we change the iconography (and at your option, the language around the name of the action itself) to stars instead of bookmarks?  I feel like the bookmark ribbon icon is neither particularly evocative nor easily recognized.  In my experience people find stars to be immediately recognizable and widely used, to the point of often not needing any further explanation/label.  Even browsers, despite having used the term "bookmark" for ages, use stars as the iconography for bookmarked pages nowadays.

One other nice thing about stars is that star buttons often need no additional outline, so they have the potential to feel lighter-weight in the UI than an equivalent outlined button does.

I feel like the button with the long text is too large/saturated for something of as minimal importance on the page as a bookmark button.  Removing the text label on the star helps, but I think the way I'd really like for such a toggle to feel is that it is an outline when the object it references is not bookmarked, and when it is, the star is filled/saturated (could be an orangish yellow, could be black/grey, could even be blue).  This could be as simple as U+2606 and U+2605 if we're feeling unicody or `fa-star` (regular and solid) if not.  I'd like the star to have a tooltip indicating what it will do if clicked (bookmark this puzzle/unbookmark this puzzle).

As for placement on the page: I'd be interested in seeing how it'd feel to place a textless star toggle in the navbar, either directly to the left of the puzzle title or to the right of the puzzle title.  I know I've been quite reticent about putting anything else in that space before and I'm admittedly still wary of how it'll behave on mobile around long puzzle titles, but I think it's worth exploring.  It feels semantically closest to a bit associated with "the puzzle", and the place on the page that the puzzle is named is in the navbar.

The other possible placement that comes to mind for me is putting the star toggle just to the left (or maybe right) of the puzzle link -- in the row, but not between buttons, and on the lighter-weight side of the row.

ebroder:
Yeah I'm totally fine using a star instead of a bookmark. I mostly used the bookmark because we called the feature bookmarking, but I waffled between bookmark and star when I started on implementation. I'll switch it over.

ebroder:
@zarvox Did you have any other thoughts on this? (I just pushed a rebase, but it was fairly uneventful other than `package-lock.json` conflicts)

zarvox (approved):
No further comments -- looks good, thanks!

PR #1820, opened 2023-09-08
Allow declaring methods and publications under models

ebroder:
@zarvox - I know you've expressed concern in the past about how `TypedMethod` and `TypedPublication` were leading to a very flat code structure, so I decided to see if I could come up with a better alternative. I'm not entirely sure if I think this is an improvement, but I think it does have some nice properties.

> Almost all of our publications and models are associated (or at least primarily associated) with a model. This is reflected in the way we name our methods ("verbNoun"). However, we previously placed all methods in a single flat directory and did the same with publications.
> 
> Instead, allow declaring methods and publications as part of creating a new model class. This has the convenient side-effect of reducing the imports you need to operate on a model to just the model class itself. It also makes the internal identifier for a method or publication match how we refer to it in code (e.g. `ChatMessages.methods.send`).
> 
> For now, the implementations of the methods and publications are still in the same files as before; that's potentially something to reevaluate in the future. Additionally, it arguably makes it harder to enumerate all methods and publications, but that might be an acceptable tradeoff.

### DIFF
--- a/imports/client/components/FirehosePage.tsx
+++ b/imports/client/components/FirehosePage.tsx
@@ -18,7 +18,6 @@ import ChatMessages from '../../lib/models/ChatMessages';
 import Puzzles from '../../lib/models/Puzzles';
 import type { PuzzleType } from '../../lib/models/Puzzles';
 import nodeIsMention from '../../lib/nodeIsMention';
-import chatMessagesForFirehose from '../../lib/publications/chatMessagesForFirehose';
 import { useBreadcrumb } from '../hooks/breadcrumb';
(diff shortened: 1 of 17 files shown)


ebroder:
Oh another note: we can probably _find_ a model to house all of our methods/publications under, but it's not always going to be clean. E.g. `teamName` and `mediasoup:debug` are both kind of messy. We can either leave those as flat-declarations or jam them under a model (e.g. `Settings` or try and do a better job of materializing the `TeamName` model for the former; the latter is...tricky).

Methods tend to be a little cleaner, but e.g. some of the setup page methods are tricky.

PR #1894, opened 2023-11-23
Take all updates that we can

ebroder:
The latest versions of mediasoup use the `??=` operator, which isn't supported on Node 14, so we need to block dependabot updates. Additionally, our block on `@types/node` didn't allow us to take updates to the v14 types, so relax that slightly.

Additionally, drop `p-throttle` as a dependency, as we have not used it since c2d07b3485e21e329926d8c0dbd89db4caa20990.

Finally, some update-related housekeeping: add aria-labels to components in `GuessQueuePage` and `Puzzle` which have tooltips (which aren't typically used as labels by screenreaders). These aren't required with our current version of `eslint-plugin-jsx-a11y`, but it seems like they will likely be required in the future. (The latest verison does require them, although it also has a bug with table elements, so we are deferring that update for now.)


zarvox (approved):
Thanks for sorting all this out!

zarvox:
Looks like maybe the `@typescript-eslint` or `typescript` updates might not be happy?

```
#23 39.56 ../node_modules/@typescript-eslint/typescript-estree/dist/ts-estree/ts-nodes.d.ts(11,15): error TS2300: Duplicate identifier 'AssertClause'.
#23 39.56 ../node_modules/@typescript-eslint/typescript-estree/dist/ts-estree/ts-nodes.d.ts(13,15): error TS2300: Duplicate identifier 'AssertEntry'.
#23 39.57 ../node_modules/typescript/lib/typescript.d.ts(6021,10): error TS2300: Duplicate identifier 'AssertEntry'.
#23 39.57 ../node_modules/typescript/lib/typescript.d.ts(6023,10): error TS2300: Duplicate identifier 'AssertClause'.
```

ebroder:
Ugh yeah, I noticed this in #1876 and wasn't sure why I didn't notice it locally, but I think I didn't mash my way through enough of the install/lint/test commands because I'm seeing it now. I'll poke around a bit.

ebroder:
Looks like this is an incompatibility between TypeScript and typescript-eslint. (You can see a reference to it here: https://github.com/typescript-eslint/typescript-eslint/pull/7968/files#r1399942500). It'll probably get fixed in the latest release of typescript-eslint, but we can't pick that up due to Node v14 incompatibilities. So we can work around it by skipping validation of the type declarations used in the eslint plugin with the `skipLibCheck` option to tsconfig.

zarvox:
> Looks like this is an incompatibility between TypeScript and typescript-eslint. (You can see a reference to it here: https://github.com/typescript-eslint/typescript-eslint/pull/7968/files#r1399942500). It'll probably get fixed in the latest release of typescript-eslint, but we can't pick that up due to Node v14 incompatibilities. So we can work around it by skipping validation of the type declarations used in the eslint plugin with the `skipLibCheck` option to tsconfig.

Was the force-push after this comment supposed to include the change you described (adding `skipLibCheck` to the tsconfig)?  It looks like it was just a rebase

ebroder:
> Was the force-push after this comment supposed to include the change you described (adding `skipLibCheck` to the tsconfig)? It looks like it was just a rebase

Ah yes it was, whoops. Should be there now.

PR #1928, opened 2023-12-30
Use transient props to styled-components

zarvox:
By default, styled-components will pass on whatever props you pass a styled component on to the underlying component.  In the cases where we're specifying props to adjust the styling, we don't actually want those props to be propagated, since e.g. `<div>` doesn't know what to do with a `isAdministrivia` property being set on it.

Since this is a common use case, styled-components has multiple mechanisms for controlling propagation of props, but the tersest and most immediately legible is that it treats props that start with a `$` as "transient props", and will not propagate any of them.

We should prefer transient props any time we are passing props to a `styled()` component purely for the purpose of controlling the styles applied.

This should remove a large volume of warnings and errors in the JS console during development.

### DIFF
--- a/imports/client/components/Avatar.tsx
+++ b/imports/client/components/Avatar.tsx
@@ -55,12 +55,12 @@ const DefaultAvatarInner = ({
   return <AvatarInitial style={style}>{initial}</AvatarInitial>;
 };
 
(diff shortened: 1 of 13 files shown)


ebroder (approved):
Oh nice, I've been meaning to go and do this.

PR #1941, opened 2024-01-07
Prepare to switch to prettier for code formatting

ebroder:
This adds prettier as a dependency and lint step and reconfigures eslint with the prettier config package (so that it stops linting for formatting that is managed by prettier), and drops a handful of eslint disables that are no longer needed (since prettier will supercede them).

And finally - in a bit of inside baseball - it does some preparatory formatting of the `tests/unit/imports/server/Model.ts` file. Once prettier reformats this file, it would move the object members onto the next line, thus moving them _out_ of line with the @ts-ignore comment. By pre-splitting them now, we keep them in sync.

Before merging, we need to pick a point where the repo is relatively idle (to prevent too many merge conflicts), and then run `meteor npx prettier --write .` to actually format the repository. Finally, create a `.git-blame-ignore-revs` file that includes the commit with the prettier formatting.


.prettierrc
```diff
@@ -0,0 +1 @@
+{}
```
ebroder, line 1: The VSCode extension that integrates with prettier requires there to be a config file in order to pull prettier out of the repo (instead of looking for a globally installed version), but we don't actually want to apply any configuration options, so we can just leave it empty.

zarvox (approved):
:shipit:

PR #1942, opened 2024-01-07
Fix rendering of post-username migration Discord handles

ebroder:
It looks like we still see a mix of Discord account references that both have and lack discriminators. At some point in the future once all usernames have been migrated, we may no longer need to track the discriminator at all, but for the time being only display it if it represents a non-migrated value (i.e. is not "0").

Before:

Screenshot 2024-01-07 at 1 36 37 PM

After:

Screenshot 2024-01-07 at 1 34 22 PM


ebroder:
(Rebased for #1941)

zarvox (approved):

PR #1943, opened 2024-01-08
Audio call styling changes

aldeka:
- Background color is peachy instead of pink
- Border and blue glow on "self" box
- Initial letter is bold
- Tweaks to mute/deafen icons/colors/sizes
- Fix weird alignment of volume bars

This is on top of @ebroder 's prettier branch so there's a bunch of changes associated with that (or prettier auto-fixes in the files I touched >.<). I could go through it by hand after the fact to take those out but I wanted to get this pushed up already :P

### DIFF
--- a/imports/client/components/Avatar.tsx
+++ b/imports/client/components/Avatar.tsx
@@ -30,10 +30,8 @@ const DiscordAvatarInner = ({
 };
 
 const AvatarInitial = styled.div`
-  width: 90%;
-  height: 90%;
-  margin: 5%;
-  border-radius: 50%;
+  width: 100%;
+  height: 100%;
   display: flex;
   justify-content: center;
   align-items: center;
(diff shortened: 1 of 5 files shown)


ebroder:
Went ahead and rebased this on top of #1941 since I had some practice doing it for my own branches. Diff looks much better now :)

Will try to take a look at the actual changes shortly.

imports/client/components/Avatar.tsx
```diff
@@ -99,13 +105,15 @@ const Avatar = React.memo(
     displayName,
     discordAccount,
     className,
+    isSelf,
```
ebroder, line 108: I think that `$isSelf` either needs to be optional on `AvatarContainer` or this needs to default to false, a la

```suggestion
    isSelf = false,
```

Mild preference for the latter I think?

imports/client/components/styling/PeopleComponents.tsx
```diff
@@ -36,10 +36,15 @@ export const AVButton: FC<ComponentPropsWithRef<typeof Button>> = styled(
...
-  justify-content: flex-end;
+  justify-content: flex-start;
+  padding-right: 12px;
+  max-height: 120px;
+  overflow-y: auto;
+  overflow-x: hidden;
 
   &:not(:empty) {
-    margin: -4px -4px 0 0;
+    /* margin: -4px -4px 0 -9px; */
```
ebroder, line 46: Was this supposed to get deleted?

```suggestion
```

ebroder:
These changes all seem fine to me, other than the small issues I mentioned.

imports/client/components/CallSection.tsx
```diff
@@ -54,13 +54,20 @@ const CallStateIcon = styled.span`
...
-  font-size: 8px;
+  border: 0.5px solid #0d6efd;
+  border-left: 0.5px solid #bbb;
+  border-top: 1px solid #bbb;
   text-align: right;
+  color: black;
+  font-size: 9px;
 `;
 
 const RemoteMuteButton = styled.div`
```
zarvox, line 73: It's not in the diff, so I can't give you the inline fix, but a few lines below this, you also need to increase the `width` & `height` of `RemoteMuteButton` from 40px to 44px

PR #1952, opened 2024-01-08
Add per-hunt "terms of use"

ebroder:
This is a way to surface rules of engagement for a hunt and require hunters to agree to them before interacting with the hunt.

Here's what it looks like:

Screenshot 2024-01-08 at 12 36 02 PM

The modal is pretty aggressively blocking - there's no way to interact with background elements without clicking the "agree" button. Once you agree, that state is stored on your user object.

There's also a preview button on the hunt edit page that you can use to preview the rendering. (We should plausibly go back and add something similar to the signup message field)

Screenshot 2024-01-08 at 12 37 28 PM
Screenshot 2024-01-08 at 12 37 35 PM

### DIFF
--- a/imports/client/components/HuntApp.tsx
+++ b/imports/client/components/HuntApp.tsx
@@ -1,9 +1,11 @@
 import { Meteor } from "meteor/meteor";
 import { useTracker } from "meteor/react-meteor-data";
 import React, { useCallback, useMemo } from "react";
+import { Modal, ModalBody, ModalFooter } from "react-bootstrap";
 import Alert from "react-bootstrap/Alert";
 import Button from "react-bootstrap/Button";
 import ButtonToolbar from "react-bootstrap/ButtonToolbar";
+import { createPortal } from "react-dom";
 import { Outlet, useNavigate, useParams } from "react-router-dom";
 import Hunts from "../../lib/models/Hunts";
 import type { HuntType } from "../../lib/models/Hunts";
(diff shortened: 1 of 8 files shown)


imports/client/components/HuntApp.tsx
```diff
@@ -144,7 +155,29 @@ const HuntApp = React.memo(() => {
...
-  return <Outlet />;
+  let termsModal = null;
+  if (mustAcceptTerms) {
+    termsModal = createPortal(
+      <Modal show size="lg">
+        <ModalBody>
+          <Markdown text={hunt.termsOfUse!} />
+        </ModalBody>
+        <ModalFooter>
+          <Button variant="primary" onClick={acceptTerms}>
```
zarvox, line 166: How annoyed do you think people would be if we made this button stay disabled for a few seconds after being shown, before enabling, to encourage people to actually take the time to read the reminders?

zarvox: Ehhh, I changed my mind, that's probably overkill.

ebroder: Not necessarily opposed but yeah seems like it's probably overkill

imports/lib/models/User.ts
```diff
@@ -40,6 +41,7 @@ export const User = z.object({
   profile: z.object({}).optional(),
   roles: z.record(z.string(), nonEmptyString.array()).optional(),
   hunts: foreignKey.array().optional(),
+  huntTermsAccepted: z.record(z.string(), z.boolean()).optional(),
```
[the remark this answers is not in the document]
ebroder: Yeah, reasonable. Pushed that change.

zarvox (approved):
Looks good!

I am amused that as soon as you save the Hunt with the terms of use, you immediately get smacked with the terms modal 😂

ebroder:
Yeah I couldn't decide if that was suboptimal or deeply appropriate 🤷🏻‍♂️

PR #1956, opened 2024-01-09
Update DEVELOPMENT.md (merging with OPERATION.md) based on what I've learned

egnor:
Doc-only update. Hopefully I got most of this right? Let me know if there's some linter or style guide or something that would apply here.


DEVELOPMENT.md
```diff
@@ -1,10 +1,17 @@
...
+[Meteor web framework](https://www.meteor.com/) version 2. To run Jolly Roger locally, you'll
+need to first [install Meteor](https://www.meteor.com/install).
 
-## Run the server
+Meteor v2 (and therefore Jolly Roger) requires [Node.js](https://nodejs.org/)
+v14 for both development and production. This is an old and deprecated version
+of Node, so you will probably want to use a version manager such as 
+[nvm](https://github.com/nvm-sh/nvm), [Volta](https://volta.sh/),
+[asdf](https://asdf-vm.com/) or (my favorite) [mise](https://mise.jdx.dev/) so
+you're not stuck with Node v14 on your whole system.
```
ebroder, line 12: Not sure whether or not this is worth actually specifying, but you only need to install node yourself if you're running a production build. In development mode, the meteor CLI comes with its own node.

DEVELOPMENT.md
```diff
@@ -28,29 +35,95 @@ meteor
...
+You will also need some dependent services, at the very least a
+Mongo DB instance (for `$MONGO_URL`), some way to send email (for `$MAIL_URL`),
+a web server that can forward to your server and `$PORT`, and some domain
+name pointing to that web server (hopefully with an SSL certificate).
+
+As a "canned configuration" for all of that,
+[`cloudformation/jolly-roger.yaml`](cloudformation/jolly-roger.yaml) is
+a template for [AWS CloudFormation](https://aws.amazon.com/cloudformation/).
+You will need to edit it for your use (change the domain name, etc) but then
+it can spin up a full production system including Jolly Roger, MongoDB,
```
ebroder, line 55: Small nit: mongodb still has to come from a 3rd party - we're using MongoDB's hosted service (Atlas)

DEVELOPMENT.md
egnor: Ooh I thought it was just its own `npm`! Weird, when I got the version wrong I was getting build issues...

egnor: Ah, that's because I was using "npm" instead of "meteor npm"

egnor: Doc fixed!

DEVELOPMENT.md
egnor: Fixed!

ebroder (approved):
LGTM! Thanks for putting this together.

PR #1959, opened 2024-01-09
Move JR nav items into top nav bar

aldeka:
I can never ever see the blue items menu (hunt website + JR utility links) even when looking for it:
- partly because it's not in an expected place for navigation,
- partly because it gets stretched super wide on non-small windows such that it doesn't even look like a set of buttons.

Now they're in the nav bar (if a hunt ID is set; they don't appear on the hunts index page).

This also improves contrast on the top nav bg a little and adds a bottom border to the nav bar (in the same color as the vertical panes separator).

In small windows the menu appears in the old spot and looks more or less like the old one. Not idea having it move, but I'm still trying to decide whether to have multiple dropdown menus in the nav on the smallest screens or something else. Definitely more fiddling possible to refine the flex/collapse behavior but I think this is workable for now; it aggressively clips so the new menu should never cause the navbar to break out in height.

Screenshot 2024-01-09 at 07 31 49

Screenshot 2024-01-09 at 07 16 02

Screenshot 2024-01-09 at 07 15 39

Screenshot 2024-01-09 at 07 15 18

### DIFF
--- a/imports/client/components/App.tsx
+++ b/imports/client/components/App.tsx
@@ -20,7 +20,7 @@ import Container from "react-bootstrap/esm/Container";
 import type { FallbackProps } from "react-error-boundary";
 import { ErrorBoundary as ReactErrorBoundary } from "react-error-boundary";
 import * as RRBS from "react-router-bootstrap";
-import { useNavigate, Link } from "react-router-dom";
+import { useNavigate, Link, useParams } from "react-router-dom";
 import type { StackFrame } from "stacktrace-js";
 import StackTrace from "stacktrace-js";
 import styled, { css } from "styled-components";
@@ -32,12 +32,13 @@ import Loading from "./Loading";
 import NotificationCenter from "./NotificationCenter";
 import { NavBarHeight } from "./styling/constants";
 import { mediaBreakpointDown } from "./styling/responsive";
+import HuntNav from "./HuntNav";
 
 const Breadcrumb = styled.nav`
   display: flex;
   align-items: center;
   height: ${NavBarHeight};
-  flex: 1;
+  flex: 1 1 auto;
   min-width: 0;
 `;
 
@@ -97,6 +98,15 @@ const NavUsername = styled.span`
 const Brand = styled.img`
   width: ${NavBarHeight};
   height: ${NavBarHeight};
+`;
+
+const HuntNavWrapper = styled.div`
+  ${mediaBreakpointDown(
+    "sm",
+    css`
+      display: none;
+    `,
+  )}
 `;
 
 const ErrorFallback = ({
@@ -180,6 +190,7 @@ const ReactErrorBoundaryFallback = ({
 
 const AppNavbar = () => {
   const userId = useTracker(() => Meteor.userId()!, []);
+  const huntId = useParams<"huntId">().huntId!;
 
   const displayName = useTracker(
     () => Meteor.user()?.displayName ?? "<no name given>",
@@ -229,7 +240,15 @@ const AppNavbar = () => {
   // correct amount of space in the top bar even if we haven't actually picked
   // a nonempty source for it yet.
   return (
-    <NavbarInset fixed="top" bg="light" variant="light" className="py-0">
+    <NavbarInset
+      fixed="top"
+      variant="light"
+      style={{
+        backgroundColor: "#f0f0f0",
+        borderBottom: "1px solid #6c757d",
+      }}
+      className="py-0"
+    >
       <NavbarBrand className="p-0">
         <Link to="/">
           <Brand
@@ -240,6 +259,11 @@ const AppNavbar = () => {
         </Link>
       </NavbarBrand>
       {breadcrumbsComponent}
+      {huntId && (
+        <HuntNavWrapper>
+          <HuntNav />
+        </HuntNavWrapper>
+      )}
       <Nav className="ml-auto">
         <Dropdown as={NavItem}>
           <DropdownToggle id="profileDropdown" as={NavLink}>
--- a/imports/client/components/App.tsx
+++ b/imports/client/components/App.tsx
@@ -28,11 +28,11 @@ import isAdmin from "../../lib/isAdmin";
 import { useBreadcrumbItems } from "../hooks/breadcrumb";
 import lookupUrl from "../lookupUrl";
 import ConnectionStatus from "./ConnectionStatus";
+import HuntNav from "./HuntNav";
 import Loading from "./Loading";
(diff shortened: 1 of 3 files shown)


imports/client/components/Avatar.tsx
```diff
@@ -30,10 +30,8 @@ const DiscordAvatarInner = ({
 };
 
 const AvatarInitial = styled.div`
-  width: 90%;
-  height: 90%;
-  margin: 5%;
-  border-radius: 50%;
+  width: 100%;
+  height: 100%;
```
aldeka, line 34: I'm so confused how this is in here, this is already in main??

grantaelliott (commented):
I like this; it’s a workflow change I’ve thought about for a while and this is a good implementation.

Could you please add screenshots showing this with pathologically long puzzle titles at different breakpoints to confirm that wrap and truncation still work as expected?

A couple of requested tweaks / points of discussion:
- Vertical dividers are inconsistent between breakpoints. I don’t think they’re necessary anyway.
- The hunt site button overflows vertically in the second screenshot. Perhaps the simplest option is the best one: Drop the button presentation and just style it as a link (as with the puzzle icon and word “Puzzle” in the metadata pane). That would be cleaner and I think should still be sufficient to make the differing behavior evident.
- I prefer the vertical icon/text layout and it leaves more room for the title breadcrumbs. Can we use it even at the widest breakpoints?
- The user dropdown should be adjusted to space and line break consistently with the other items.
- The fallback to the old style on narrow viewports is reasonable. Long term, I think it’d be best at this breakpoint to replace the user dropdown with a hamburger dropdown containing both navigation and user controls (presumably with a divider). Depending on how much time you have for this, I’m happy with landing it with the existing fallback.

imports/client/components/HuntNav.tsx
```diff
@@ -0,0 +1,216 @@
...
+    css`
+      justify-content: space-between;
+      border: 1px solid #0d6efd;
+      margin-right: 0;
+      padding-right: 0;
+    `,
+  )}
+`;
+
+const StyledPuzzleListLinkAnchor = styled(Nav.Link)`
```
zarvox, line 80: I'm noticing that after this change, all the nav items are triggering full page reloads (loading html, code, & assets, opening a new websocket, etc.), rather than doing in-app navigation.

I think in the old behavior this was a styled `react-router-dom` `Link`, which overrides some handlers to avoid triggering navigation.  Can you make sure that element is in the component hierarchy, and verify that following these links trigger only in-app navigation, rather than browser-level navigation?

imports/client/components/HuntNav.tsx
```diff
@@ -0,0 +1,216 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import React from "react";
+import Button from "react-bootstrap/Button";
+import { useParams } from "react-router-dom";
+import styled, { css } from "styled-components";
+import Hunts from "../../lib/models/Hunts";
+import { userMayWritePuzzlesForHunt } from "../../lib/permission_stubs";
+import { mediaBreakpointDown } from "./styling/responsive";
+import { Nav } from "react-bootstrap";
+
```
zarvox, line 17: Add an import here for use down below:

```javascript
import { LinkContainer } from "react-router-bootstrap";
```

imports/client/components/HuntNav.tsx
zarvox: update: looks like the tool of choice (since you want to provide the `<a>` element itself) is `react-router-bootstrap`'s `<LinkContainer>`

ebroder:
(FYI, I think I've changed the repository settings so that CI will run on your PRs going forward)

aldeka:
Screenshot 2024-01-09 at 21 41 43

Screenshot 2024-01-09 at 21 42 10

Screenshot 2024-01-09 at 21 42 38

Screenshot 2024-01-09 at 21 42 49

Screenshot 2024-01-09 at 21 43 02

imports/client/components/App.tsx
```diff
@@ -229,7 +240,15 @@ const AppNavbar = () => {
   // correct amount of space in the top bar even if we haven't actually picked
   // a nonempty source for it yet.
   return (
-    <NavbarInset fixed="top" bg="light" variant="light" className="py-0">
+    <NavbarInset
+      fixed="top"
+      variant="light"
+      style={{
+        backgroundColor: "#f0f0f0",
```
zarvox, line 247: How attached are you to this darker `backgroundColor`?  I ask because the medium-grey on light grey is worse for text readability in the topbar than the previous `#f8f8f8` (see https://webaim.org/resources/contrastchecker/ ) and I found it a little harder to read the active breadcrumb, which is arguably the most important thing to be able to read in the navbar, since it's how you tell which puzzle you're looking at.

Can we keep this at `#f8f8f8`, or even go full `#fff` (which would pass WCAG AA for normal text)?  Alternately: could we make the active breadcrumb foreground color darker?

If the overall goal here was making the navbar feel sufficiently "separate" from the rest of the page that it's not getting in the way of readability of stuff farther down on e.g. the puzzle list page, another thing we could try to make the navbar easier to ignore/distinguish would be to (probably in a separate PR, the layout stuff is somewhat finicky) make the non-fixed-layout puzzle pages at desktop widths be actually able to scroll away the navbar, rather than leaving it on top at all times, which I think is the difference in behavior that makes our navbar feel more awkwardl than those on some other sites.

imports/client/components/HuntNav.tsx
zarvox: Looks like you solved this with `react-router-dom`'s `NavItem`

imports/client/components/HuntNav.tsx
```diff
@@ -0,0 +1,216 @@
...
+          <HuntHomeButtonContents>
+            <MenuIcon icon={faMap} />
+            <StyledPuzzleListLinkLabel>Hunt site</StyledPuzzleListLinkLabel>
+          </HuntHomeButtonContents>
+        </Button>
+      </HuntLinkWrapper>
+    );
+    return (
+      <JRLinkList>
+        <StyledPuzzleListLinkAnchor
```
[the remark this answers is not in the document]
zarvox: Looks like you solved this with `react-router-dom`'s `NavLink`

zarvox:
Broadly looks and feels good!

I intend to approve this once you:

* address the breadcrumb contrast ratio by making the navbar lighter or the text darker
* fix up the presubmit check errors

zarvox (approved):
My feedback is addressed!

PR #1960, opened 2024-01-09
Fix breadcrumbs when looking at a profile from a hunt

ebroder:
Historically, we've only had a single route for viewing a user profile under `/users`, but that meant that breadcrumbs broke when you went from a hunt-specific profile list to an individual profile.

This mounts the profile view under both `/users/:userId` and `/hunts/:huntId/hunters/:userId` and makes sure that breadcrumbs behave correctly regardless of how you get there.

🎁 @flipdog


flipdog:
My hero <3

zarvox (approved):

PR #1962, opened 2024-01-10
Try to fix haproxy config again

ebroder:
The biggest improvement seems to have come from pinning the older version of haproxy, as I think that 2.9.1 has some sort of issue, at least with our configuration. While I was debugging, I went and enabled some additional timeouts to make sure we were GCing connections successfully - even if those weren't the proximate issue, they still seem useful to prevent resource leaks.

### DIFF
--- a/cloudformation/jolly-roger.yaml
+++ b/cloudformation/jolly-roger.yaml
@@ -698,6 +698,9 @@ Resources:
                     timeout connect 5s
                     timeout client 50s
                     timeout server 50s
+                    timeout http-request 20s
+                    timeout client-fin 20s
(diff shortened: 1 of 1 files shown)


cloudformation/jolly-roger.yaml
```diff
@@ -698,6 +698,9 @@ Resources:
                     timeout connect 5s
                     timeout client 50s
                     timeout server 50s
+                    timeout http-request 20s
```
zarvox, line 701: `haproxy` knows about websockets, right?  Just making sure this won't force us to fall back to XHR

zarvox (approved):

PR #1968, opened 2024-01-11
Allow top navbar to be scrolled away on most pages

zarvox:
The always-visible topbar is actually not the best use of space on most pages:

* the topbar is easily found by scrolling back to the top of the page
* not having it take up 50 pixels of vertical space means 50 more rows for more-useful content like the puzzle list or guess queue
* having the top bar scroll away means we don't have to worry as much about having it be strongly-contrasting in color with other content that would slide under the fixed navbar.
* while some SPAs do have a more fixed layout, most other websites that have a navbar like this either do not make it sticky at all, or at least hide it when scrolling down

So let's just make the topbar be scrollable like most of the rest of the web, and give a little more space for content.

The view on `PuzzlePage` will not change at all, since we use a more fixed layout there anyway to position/size the chat and iframe.

### DIFF
--- a/imports/client/components/App.tsx
+++ b/imports/client/components/App.tsx
@@ -43,10 +43,10 @@ const Breadcrumb = styled.nav`
 `;
 
 const ContentContainer = styled.div`
-  margin-top: calc(env(safe-area-inset-top, 0px) + ${NavBarHeight});
-  padding-bottom: max(env(safe-area-inset-bottom, 0px), 20px);
-  padding-left: max(env(safe-area-inset-left, 0px), 15px);
-  padding-right: max(env(safe-area-inset-right, 0px), 15px);
+  padding: /* top right bottom left */ max(env(safe-area-inset-top, 0px), 15px)
+    max(env(safe-area-inset-right, 0px), 15px)
+    max(env(safe-area-inset-bottom, 0px), 20px)
+    max(env(safe-area-inset-left, 0px), 15px);
 `;
 
 /* Using some prefixed styles with widespread support and graceful failure */
@@ -241,7 +241,6 @@ const AppNavbar = () => {
   // a nonempty source for it yet.
   return (
     <NavbarInset
-      fixed="top"
       variant="light"
       style={{
         backgroundColor: "#f0f0f0",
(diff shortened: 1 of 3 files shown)


imports/client/components/App.tsx
```diff
@@ -319,12 +324,15 @@ const App = ({ children }: { children: React.ReactNode }) => {
       </ReactErrorBoundary>
     );
   }
+
+  const onPuzzlePage = useParams<"puzzleId">().puzzleId !== undefined;
```
ebroder, line 328: I don't really love this? It feels fragile and only tangentially connected to the question we're actually trying to answer. (What if we add another page with a puzzleId?)

On the other hand, I'm not sure we need it? I see two issues:

First, the border renders under the sidebar:

Screenshot 2024-01-11 at 9 19 26 AM

I think this is just a fencepost error.

And second, the navbar is still scrollable even when the rest of the page content isn't:

https://github.com/deathandmayhem/jolly-roger/assets/28167/a4e73595-c92f-4025-846a-6c8338425ee2

And I think through mild shenanigans we can fix both of these in the `FixedLayout` component:

```TSX
const FixedLayoutDiv = styled.div`
  position: fixed;
  inset: /* top right bottom left */ calc(
      env(safe-area-inset-top, 0) + ${NavBarHeight} + 1px
    )
    env(safe-area-inset-right, 0) 0 env(safe-area-inset-left, 0);
`;

const FixedLayoutGlobal = createGlobalStyle`
  body {
    overscroll-behavior: none;
  }
`;

const FixedLayout = React.forwardRef<
  HTMLDivElement,
  React.HTMLAttributes<HTMLDivElement>
>((props, ref) => {
  return (
    <>
      <FixedLayoutGlobal />
      <FixedLayoutDiv ref={ref} {...props} />
    </>
  );
});

export default FixedLayout;
```

zarvox: Ah, neat -- if we do the `FixedLayout` shenanigans with disabling `overscroll-behavior`, we can get away with just always using the non-fixed navbar, even though it's the only part of the body whose content contributes to the flow (vs having fixed position)?  That works for me!

ebroder (approved):

PR #1969, opened 2024-01-11
Swap port allocations between mediasoup and coturn

ebroder:
MIT's network is documented to allow UDP connections on ports 50000+. By moving mediasoup into that port range, we can hopefully avoid needing a TURN relay on this network.

The deployment of these two changes will be a bit messy, as they can't be pushed out atomically with each other. That should be fine as long as they deploy when Jolly Roger isn't being heavily used.

### DIFF
--- a/cloudformation/jolly-roger.yaml
+++ b/cloudformation/jolly-roger.yaml
@@ -349,18 +349,18 @@ Resources:
         - Description: Mediasoup RTC ports
           CidrIp: 0.0.0.0/0
           IpProtocol: udp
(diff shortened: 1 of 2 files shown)


ebroder:
Oh, I can actually make this deploy quasi atomic:

- `ssh` into the current Jolly Roger instance and stop the `watchtower` Docker container, preventing updates from going out
- Wait for the build to complete
- Update the CloudFormation template. When the replacement instance spins up, it'll get the new container

zarvox (approved):

PR #1992, opened 2024-01-15
Send puzzle-solved notifications only to the solved-puzzle's hunt

### LINKED ISSUE #1991: Only send on-puzzle-solved chat message to puzzles with matching group tag in the same hunt as the solved puzzle
Otherwise, reusing a group tag's name (like `group:runaround`) across different hunts and then solving that puzzle produces notifications on all metas for that group

zarvox:
Fix an unindexed Puzzles find query that also probably wants to limit to the current hunt while I'm here.

Fixes #1991.

### DIFF
--- a/imports/server/hooks/ChatHooks.ts
+++ b/imports/server/hooks/ChatHooks.ts
@@ -16,9 +16,13 @@ const ChatHooks: Hookset = {
       group.name.substring("group:".length),
     );
     const metaTags = await Tags.find({
-      name: { $in: groupNames.map((name) => `meta-for:${name}`) },
+      hunt: puzzle.hunt,
+      name: {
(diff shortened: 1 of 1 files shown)


ebroder (approved):
LGTM. Are you sure that there's not also a bug with tagging a puzzle from the wrong hunt?

zarvox:
Yes, the logic here is sufficient to explain the observed behavior (we happened to have at least two `group:runaround` tags in the DB, and the puzzle we saw the notification for was `meta-for:runaround`, and we didn't have a `hunt` selector on this query).  I checked all the places we fetch tags in the client components, and these are the only ones that didn't specify either an `_id` or a `hunt`.

PR #2006, opened 2024-01-20
Fix eslint/tsconfig for Windows

ebroder:
On Windows, npm explore eslint-plugin-jolly-roger seems to result in the pwd being node_modules/eslint-plugin-jolly-roger, instead of eslint (like we see on POSIX platforms), which causes the extend to break.

There's no philosophical motivation for having that inheritance in the first place - the two projects are fairly disconnected and I only used the "extends" the first time out of laziness, but this at least gets the Windows install a little further along.

Thanks to jpd for the bug report!


zarvox (approved):
You might need to rebase this to pick up the `prettier` change (which I believe underlies the presubmit check complaint).

PR #2015, opened 2024-01-26
Make SSH user list configurable in CloudFormation config.

### LINKED ISSUE #2013: Remove hardcoded user list from CloudFormation config
Currently, IIUC, the CloudFormation configuration (specifically the EC2 instance) has a hardcoded user list allowing the D&M tech team to log in with their GitHub credentials [here](https://github.com/deathandmayhem/jolly-roger/blob/main/cloudformation/jolly-roger.yaml#L546). Naturally, this doesn't generalize for other use cases.

I'm not yet familiar with this configuration format, but I imagine doing one of the following:

- Add a parameter which is a list of user configs, where a user config is something like `<name>:<github account>`, and expand that to the user config here. This would be the bare minimum amount of generality that would work for the existing set of users while also being usable by anyone else as long as they don't mind using their GitHub public key to login.
- Make the entire user section a parameter for full control.
- Some other way of configuring users with a bit more structure?

jpd236:
We add a parameter to specify a CSV of the users and their SSH keys, and we use a lambda function to convert it into the "users:" stanza for the EC2 config.

An alternative could be to specify two parallel parameters, one for the usernames, and one for the SSH keys. We could then use the AWS::LanguageExtensions function Fn::ForEach to loop over one list while using Fn::Select to pull the corresponding object from the other list. This would simplify the implementation, but make the "API" (the parameters) arguably more complex.

Fixes #2013

### DIFF
--- a/cloudformation/jolly-roger.yaml
+++ b/cloudformation/jolly-roger.yaml
@@ -113,13 +113,17 @@ Parameters:
     Type: String
     Default: "true"
     AllowedValues: ["true", "false"]
+  SshUsers:
+    Description: Comma-separated list of SSH users, each of the form <username>=<ssh_import_id>
+    Type: String
+    Default: "evan=gh:ebroder,zarvox=gh:zarvox"
 
 Conditions:
   HavePapertrail: !Not [!Equals [!Ref PapertrailHost, ""]]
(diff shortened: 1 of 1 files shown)


ebroder (approved):
This is amazing 🙂

PR #2020, opened 2024-01-31
Enable AWS IAM authentication with MongoDB.

jpd236:
- Use a fixed name for the role used to run EC2 instances so it can be referenced in the MongoDB config. Note that if you try to set up JR instances across multiple regions, it will now fail due to the role name being duplicated. If this is a concern, we could append the region name to the role (but this would require one user per region in MongoDB).
- Include aws4 in dependencies as required for AWS authentication.

### DIFF
--- a/cloudformation/jolly-roger.yaml
+++ b/cloudformation/jolly-roger.yaml
@@ -47,7 +47,13 @@
 #   be configured as the CertificateArn paramater.
 #
 # - Store the MongoDB URL using `credstash put mongo @<path>`. The free account
-#   from mongodb.com (for instance) should be sufficient.
+#   from mongodb.com (for instance) should be sufficient. You can configure a
+#   user using the "AWS IAM" authentication method, the type "IAM Role", and the
+#   ARN "arn:aws:iam::<AWS Account ID>:role/JollyRogerAppInstanceRole"; then you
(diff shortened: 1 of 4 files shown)


ebroder:
Hmm. If we do this, should we also switch to accepting MONGO_URL (or `MONGO_HOST` maybe?) via parameter and just constructing it instead of needing credstash at all for that piece?

jpd236:
Sure - we now can take MongoUrl and MongoOplogUrl parameters from CF, or else fall back to credstash as before.

I was a little more hesitant to just take the host and synthesize the URL as there are a few parameters there (particularly the database name) that one might want to customize. I can do this though if preferred. FWIW, here are the URLs I've been using (just taken from what Atlas produced for my instance):

```
MONGO_URL=mongodb+srv://<host>/jollyroger?authSource=%24external&authMechanism=MONGODB-AWS&retryWrites=true&w=majority
MONGO_OPLOG_URL=mongodb+srv://host>/local?authSource=%24external&authMechanism=MONGODB-AWS&retryWrites=true&w=majority
```

ebroder:
Yeah agreed with not synthesizing the URL from hostname. LGTM!

ebroder (approved):

PR #2022, opened 2024-02-02
Provide a CloudFormation parameter allowing stack shutdown.

jpd236:
If this parameter is false (to turn off serving), we only keep around static entities - e.g. roles, instance functions, and the S3 bucket for image uploads. This allows us to more easily flip serving off and on without having to fully delete the stack, which has side effects like deleting the S3 bucket and requiring reentry of all parameters on recreation.


ebroder (approved):
Seems fine to me

Issue #2021, opened 2024-02-01
Support for Google Account required authentication model


jpd236:
Currently, Jolly Roger has its own independent authentication system and set of accounts. In contrast, Ange Management currently requires that all team members have a Google Account. This enables the following:

- Authentication by checking membership in the group. Unfortunately, there's no direct way I'm aware of to check membership in a regular consumer Google Group. Instead, we check for access to the hunt's drive folder, which is shared with the Google Group. To avoid having to request full Drive access, which is a scary OAuth scope, we only ask for per-file access and we use the [Drive picker](https://developers.google.com/drive/picker/guides/overview) and ask users to select this year's hunt folder, which gives us access to that specific folder. This was implemented a few years ago - maybe there is a better way now. But it works.
- No need to share spreadsheets with all users and risk running into limits. You can just share directly with the group, and users will show up as themselves (not anonymous) in the sheet.
- No need for an email system since all of the existing emails relate to account creation and maintenance and being added to hunts, AFAICT.

FWIW, we also use a service account to create the sheets and perform other Google operations. IIRC the current system asks for a dedicated actual user account for this purpose. Service accounts feel better suited to this purpose.

Meteor has a built-in Google Account integration which might be useful (our current system is DIY). At a high level, I imagine offering this as a configurable option that replaces the existing log-in system with a "Sign in with Google" option; after signing in, we create a user entry in JR including any necessary credentials for authenticated requests. Would have to figure out things like the initial user creation flow for the admin, hunt creation creating folders and doing the per-hunt auth dance as needed.

I'm planning to take a look at this myself but concerns, thoughts and ideas would be welcome! I'd like to upstream this to avoid drift, and I think that Palindrome might find at least parts of this useful.

Issue #698, opened 2022-02-11
Audit API calls and subscriptions for hunt membership checks


ebroder:
Today, in some cases we rely on the information entropy in IDs as a de facto permission check. E.g. in the `addTagToPuzzle` method, we don't verify that the caller is a member of the hunt of the puzzle to which they're attempting to add a tag. This isn't a huge deal with mostly trusted users, but as we contemplate more open JR membership, we should make sure all of these methods do explicit permission checks.

(This also means we should finally address the lack of permissioning around viewer subscirptions)

Issue #631, opened 2022-01-20
Support unlisted, private hunts


ebroder:
Today all hunts are public to any logged in user. In an "any user can create hunts" world, we likely want to have publicly-listed hunts (for big events like Mystery Hunt) and unlisted, invite-only hunts, likely defaulting to the latter. This should likely be a toggle for the operators of the hunt to set.

Given that our permissions are already oriented around hunts, this should hopefully not be too challenging; it mostly means that when we publish hunts to clients, we need to make sure that we don't publish hunts that are both (a) unlisted (b) not in the user's membership list.

Issue #296, opened 2020-01-21
Merging accounts


ebroder:
We end up with a lot of people that have signed up multiple times under different email addresses. It would be great if it was somehow possible to merge those to consolidate history and clean up cruft.

We've historically ignored Meteor's ability to have multiple email addresses associated with the same user account, but that's probably one part of the solution here. We'd also have to think about what to do with historical data - e.g. do we update the `createdBy` on everything that the user has touched? Seems potentially expensive, but maybe still worthwhile.

As an alternative, maybe we leave behind a tombstone account that gets credited for past actions, but future actions get associated with the new user.

ebroder:
OK. I think I have mostly come up with a design for this that works. @zarvox take a read through and let me know if you can spot any issues

I'll use "target user" throughout as the user that will continue to exist after the merge completes and "source user" as the one that is going away. Here's the rundown:

## Data model changes

We introduce a new model, `MergeOperation`, which is used as a tracker record for merges, with a schema that looks roughly like this:

```ts
const MergeOperation = withCommon(z.object({
  // A tuple of source user and target user
  participants: z.tuple([foreignKey, foreignKey]),
  // The background job that created this operation
  job: foreignKey,
  // Whether or not this merge has been successfully completed
  completedAt: z.date().optional(),
  // A snapshot of fields with unique constraints on users, to use as the source
  // for populating the target user after the source has been scrubbed
  snapshot: z.object({
    emails: z.object({
      address: z.string(),
      verified: z.boolean(),
    }).array(),
    googleAccount: z.string().optional(),
    googleAccountId: z.string().optional(),
    googleProfilePicture: z.string().optional(),
    discordAccount: DiscordAccount.optional(),
  }).optional(),
});
```

and with the following index declaration:

```ts
MergeOperations.addIndex({ participants: 1 }, {
  unique: true,
  partialFilterExpression: { completedAt: null },
});
```

Unique indexes on array fields in MongoDB index all elements of the array, so taken together, this index enforces that no user ID can be involved (on either side) in more than one incomplete merge. This gives us concurrency safety.

(For now, we don't expire merge operations, but we could consider adding a TTL index on `completedAt` with some delay.)

Additionally, we add a new field to users:

```ts
  mergedInto: foreignKey.optional(),
```

When populated, we prevent the user from logging in.

And although `services` is defined as `z.any().optional()`, we will be storing a new key on it:

```
  mergeConfirmation: z.object({
    token: z.string(),
[...]

PR #2040, opened 2024-02-18
Add support for signing in with a Google Account.

jpd236:
Show a "Sign in with Google" button on the login page. This sign in flow first attempts to match against existing users who have linked that same Google Account. If there is no such user, a new account is created.

Icon taken from:
https://developers.google.com/identity/branding-guidelines

Notes / issues:

- This allows anyone to create a Jolly Roger account as long as they have a Jolly Roger account; previously, access was invite only. This seems reasonable in that hunts have their own ACLs, but if this is a concern, we could reject new account creation until a follow-up change which optionally configures access based on Drive folder access.

- A user can unlink their Google Account; if the user was originally created through this flow, they will be left with an account with no password or method of entry. However, they can follow the forgotten password flow to create a new password (if an email server is configured).

- At least on local instances, a console warning appears upon showing the Google Account popup: "Cross-Origin-Opener-Policy policy would block the window.closed call". This also occurs for the existing Google Account integration so isn't exclusive to this flow.

- If the popup is closed without completing sign-in, the button remains disabled because we don't appear to get a failure callback. I'm unsure why this is happening, though it's easy enough to refresh the page.

- Sometimes, when you click Sign Out, the login page renders with a blank "Sign in with Google" button, though the button appears to work as expected.

See #2021


ebroder (commented):
High level feedback: I think the implementation here is good, and I'm generally supportive of the change. As you called out, it does let anybody create a JR account, and I think that that runs into a couple other specific open issues.

In particular, I think that both #631 and #698 are potentially cases where we surface sensitive (I mean, sure, "sensitive") information to users that shouldn't be able to see it, which has mostly been OK historically. I also think that having a one-click account creation button is potentially going to exacerbate #296.

I would love if we could make progress on those issues, but I don't want to put a bunch of roadblocks in front of you. As a compromise, could we hide the Google account login behind a feature flag? That way it can be off by default to avoid those interactions, but you can turn it on for your deployment. And if #296, #631, and #698 ever get fixed, I'd be very happy removing the feature flag and making this behavior the default.

imports/client/components/AccountForm.tsx
```diff
@@ -246,6 +256,51 @@ const AccountForm = (props: AccountFormProps) => {
     return <div>loading...</div>;
   }
 
+  const SignInWithGoogleBlock = () => {
```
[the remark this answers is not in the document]
jpd236: It uses a couple of things from the form: the submit state (to disable widgets while the popup is open) and the error message (to show an error if sign-in fails).

imports/server/accounts.ts
```diff
@@ -45,6 +45,53 @@ const summaryFromLoginInfo = function (info: LoginInfo) {
...
+    interface ExternalLoginInfo {
+      serviceName: string;
+      serviceData: Meteor.GoogleServiceInfo;
+    }
+
+    function setAdditionalFindUserOnExternalLogin(
+      func: (info: ExternalLoginInfo) => Meteor.User | undefined
+    ): void;
+  }
+}
```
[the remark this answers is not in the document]
jpd236: Thanks, didn't realize there was a central place for these. Mov

imports/client/components/AccountForm.tsx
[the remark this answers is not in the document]
ebroder: Ah sure. I think I'd still prefer to see this live at top-level - defining a component inside of another component strikes me as kind of weird. (I know that it'll have slightly weird behavior on render, since each render will re-create the function and require swapping the component out)

imports/server/accounts.ts
```diff
@@ -45,6 +45,32 @@ const summaryFromLoginInfo = function (info: LoginInfo) {
...
+  return user;
+});
+
+Accounts.setAdditionalFindUserOnExternalLogin(
+  (info: Accounts.ExternalLoginInfo) => {
+    if (info.serviceName === "google") {
+      // Attempt to match existing Google users by their account ID.
+      // We can't use the async method since Meteor's API only takes a sync one.
+      // eslint-disable-next-line jolly-roger/no-disallowed-sync-methods
+      return MeteorUsers.findOne({ googleAccountId: info.serviceData.id });
```
ebroder, line 68: Oh we probably need to make sure there's only one user with that Google ID, because we don't enforce uniqueness. So, like, find with limit=2 and make sure you get exactly 1 back

ebroder:
Yeah I'm fine with "throw an error if the account doesn't exist already"

Just for the record, I know that your Drive folder thing is coming, and I haven't decided how I feel about it yet 🙂. I guess I need to get on that.

imports/server/accounts.ts
jpd236: On digging further, this might be a bit more involved to get right.

Meteor is tracking the Google association in its own field, services.google.id ([ref](https://github.com/meteor/meteor/blob/c26569d224eb081b9aef950f44066c555ca9b6f9/packages/accounts-base/accounts_server.js#L1296)). While the current PR does a one-time link between these two fields, looking for JR users by existing googleAccountId, and (in the commented-out code) populating the googleAccountId from the service ID on new account creation, after this point, nothing is keeping these in sync. So you can still run into weirdness by e.g. making an association with a unique account at the time, and then associating more accounts with that Google account via the JR profile page.

Do you have any opinions on how we should handle this? I feel conceptually like having two different fields for this will lead to pain; maybe we should just migrate our existing fields to the standard Meteor ones, or will that cause other problems?

ebroder: Ugh yes that is annoying. I agree that if we're going to support logging in with Google account, then we should also migrate our existing fields onto the Meteor standard ones.

Here are the reasons I can think of that it's going to suck:
- We now need to care about uniqueness. Unfortunately we *do* actually need to explicitly handle this somehow, at least for the Death and Mayhem deployment. I just checked and we have about 25 Google accounts that are associated with more than one JR account. (Realistically, I'm pretty sure all of these should be merged into a single account, but that's a more complicated problem)
- We probably need to migrate incrementally (as people log in or re-link their accounts), rather than being able to do it all at once. The `googleAccount` and `googleAccountId` fields aren't enough to populate Meteor's service record. I guess it's possible that we can fudge it (and just not have an accessToken or any of the other fields), but I'm not sure if that will break things down the line. My personal preference would be to lazily migrate the data, either when someone logs in using an existing Google account or when they re-link their Google account, and then at some point in the future we can deprecate that and write a database migration to remove any users that still have the old fields set.

jpd236: After thinking about it more, I'm less convinced that it's worth using the standard mechanism at this time. Now that we're moving towards invitation links rather than other Google-based methods, we don't have a particular use case for it right now, and a custom login handler that uses our existing fields nets out to be simpler than the standard one, given that most of the complexity is already in the google-oauth module anyway, and that we don't have to deal with any migration questions at the moment.

So I've pivoted this PR to using a custom login handler instead, and we could punt migrating to the standard handler unless/until there's a more compelling reason for it.

I did find a bug along the way - unlinking a Google Account only cleared the linked email but not the ID. I've fixed that here, but if this PR doesn't get merged for any reason we should make sure not to lose track of that.

Let me know what you think!

ebroder:
Sorry again for not getting a chance to look at this sooner. This seems like a good compromise approach for now, and the implementation seems good to me.

I did have one question around your specific requirements. If I'm reading this correctly, it's not possible to create a new JR account via Google sign-in, right? Only log into an existing account? And looking at #2056, it looks like the invite links currently require you to already have an account to apply an invite. Do you have a flow in mind for how your users create their accounts initially?

ebroder:
Ah. I just read your comment in the description of #2056 as reserving that for future work. In that case, I'm happy with this

ebroder (approved):

Issue #2047, opened 2024-02-24
Support durable invitation links to hunts


jpd236:
Allow members of a hunt to generate invitation links to that hunt. Invitation links would allow existing users to sign-in to get access to that hunt, or allow new users to create an account and then join that hunt.

I think the most reasonable way to start here would be to allow only admins to (re)generate invitation links for a hunt, but allow all hunt members to view the link once generated. We could go further like Discord and track invitation URLs on a per-user basis, with an admin dashboard to view invitation usage and block individual invitation URLs in case of abuse, but this feels excessive to me for now.

ebroder:
@jpd236 What's the still pending work for this issue? Or can we close it out now that #2498 has landed?

PR #2056, opened 2024-03-02
Add initial support for durable invitation links.

jpd236:
- Add invitation codes to the hunt schema. Codes are randomly generated IDs.
- Allow admins and hunt operators to (re)generate and clear invitation links from the hunt profile list page.
- Allow any user with invitation permissions to see the current invitation link (if any) from the hunt profile list page.
- Add an authenticated /join/:invitationCode endpoint which adds the current user to the hunt with that code and redirects to that hunt page. (If a user is unauthenticated when they open this link, they will be redirected here after signing in.)

To be considered for this PR:

- [x] ~~I believe users can access invitation IDs from the raw hunt data, even if not made visible, since they are part
  of the hunt schema and are not filtered from the client. Is this resolvable without a broader change?~~ Resolved by separating invitations out of hunts and into their own model
- [ ] UX polish. Possible ideas include:
  - [x] Confirmation dialog when generating/regenerating/clearing invitation links
  - [ ] Confirmation message upon successfully joining via an invitation link
  - [ ] Better UI for the invitation code
- [ ] Unit tests

To be handled in a future PR:

- When an unauthenticated user opens an invitation link and is redirected to /login, provide the option to
  create an account. This can be detected by checking the history state since it is stored on redirect. If the invitation link is verified upon form submission, create the new user and add it to the hunt.

See #2047


imports/lib/models/Hunts.ts
```diff
@@ -68,6 +71,7 @@ export const HuntPattern = {
   puzzleHooksDiscordChannel: Match.Optional(SavedDiscordObjectPattern),
   firehoseDiscordChannel: Match.Optional(SavedDiscordObjectPattern),
   memberDiscordRole: Match.Optional(SavedDiscordObjectPattern),
+  invitationCode: Match.Optional(String),
```
ebroder, line 74: `HuntPattern` is used by the `createHunt` and `updateHunt` methods to determine what parameters users are allowed to set. I think we want users going through the new methods, rather than setting it directly, so I think we can drop this

imports/client/components/JoinHunt.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+  useEffect(() => {
+    acceptHuntInvitationCode.call({ invitationCode }, (error, huntId) => {
+      if (error) {
+        setStatus(error.reason ?? "Unknown error");
+      } else {
+        navigate(`/hunts/${huntId}`);
+      }
+    });
+  }, [invitationCode, navigate]);
```
ebroder, line 19: @zarvox Since you did the `StrictMode` migration work, is this going to result in calling the method more than once?

imports/client/components/ProfileList.tsx
```diff
@@ -367,6 +371,67 @@ const ProfileList = ({
...
+    if (!hunt || !canInvite || !hunt.invitationCode) {
+      return null;
+    }
+
+    return (
+      <p>
+        Invitation link:{" "}
+        <a href={`/join/${hunt.invitationCode}`}>
+          {`${window.location.origin}/join/${hunt.invitationCode}`}
+        </a>
```
ebroder, line 384: Should this be a copy-to-clipboard button instead of a link?

ebroder (commented):
> * I believe users can access invitation IDs from the raw hunt data, even if not made visible, since they are part
>   of the hunt schema and are not filtered from the client. Is this resolvable without a broader change?

Yes, that's accurate. We're overdue to reconsider the Hunt publishing model. In the shorter term, I think it might be easier to just create a new model for invite codes? It means that we can create a new record when the invite code is regenerated and see old codes easily

imports/client/components/JoinHunt.tsx
jpd236: AIUI - yes, in strict mode, this method will be invoked twice ([ref](https://react.dev/reference/react/useEffect#caveats)). But it is more or less idempotent - if the user is already a member of the given hunt, it will log a message (at info) but otherwise return successfully. LMK if I should handle this in some other way though.

imports/client/components/ProfileList.tsx
jpd236: Changed to plain text with a copy-to-clipboard button (cribbed from the guess queue page).

imports/lib/models/Hunts.ts
jpd236: Ack, removed as part of move to a dedicated model for invitation codes.

jpd236 (commented):
Thanks for the review!

Moved invitation codes to a separate model. The publication for that model only returns the code if the user viewing the hunt has permission to add users, so I think this resolves the permission issues, if I understand this all correctly.

imports/client/components/ProfileList.tsx
```diff
@@ -367,6 +384,79 @@ const ProfileList = ({
...
   }, [hunt, canSyncDiscord, syncDiscord]);
 
+  const invitationLink = useMemo(() => {
+    if (!hunt || !canInvite || !invitationCode) {
+      return null;
+    }
+
+    const copyTooltip = <Tooltip>Copy to clipboard</Tooltip>;
+
+    const invitationUrl = `${window.location.origin}/join/${invitationCode}`;
```
ebroder, line 394: Small nit: I think this is slightly better expressed as
```ts
Meteor.absoluteUrl(`/join/${invitationCode}`)
```

imports/client/components/ProfileList.tsx
```diff
@@ -367,6 +384,79 @@ const ProfileList = ({
...
+        </FormText>
+      </FormGroup>
+    );
+  }, [
+    hunt,
+    canUpdateHuntInvitationCode,
+    clearInvitationLink,
+    generateInvitationLink,
+    invitationCode,
+  ]);
```
ebroder, line 458: I'm going to channel @zarvox for a sec and comment that there's no user feedback when you click on these buttons. At minimum, they should probably be disabled while the API call is in flight. Drew would probably ask that errors get presented if there is one, although I'm generally a little less picky about that. (I think in this case we probably don't need an explicit indicator of success, since that's reflected in the button/link content visually changing)

ebroder (commented):
Sorry - again - that this has taken a while to get back to. I'm going to make an effort to get caught up this weekend, at least.

Overall, the implementation looks good to me. Thanks for reworking the model situation. I think this is a lot better. I noticed two things looking through this revision, but otherwise the guts look good.

In terms of UI tweaks, I'd definitely support a confirmation dialog for regenerating and clearing an invitation link (basically, anything that will make currently in-flight links invalid). It feels a little less important to me from a destructiveness perspective for the first time you generate a link, although maybe people would find it surprising.

(You also mentioned invitation code UI; I'm not sure what I would do there, but open if you have suggestions. I looked at Discord for inspiration and it basically just spits out a full URL with a copy link, so I'm not sure this is significantly different/worse)

imports/client/components/ProfileList.tsx
jpd236: Done, thanks!

imports/client/components/ProfileList.tsx
jpd236: Makes sense. I went ahead and added confirmation dialogs for all of the operations, repurposing the existing dialogs for promoting/demoting operators. Those dialogs already handle in-flight button disabling and showing errors, so this should now be covered.

ebroder (approved):
LGTM. Thanks again for tolerating our sluggishness on review!

PR #2098, opened 2024-04-12
Pin @testing-library/react to below v15

ebroder:
v15 dropped support for Node 14 (and introduced usage of the `&&=` operator, which is explicitly unsupported).


zarvox (approved):
LGTM

PR #2101, opened 2024-04-19
Support unauthenticated users in hunt invitation flow.

### LINKED ISSUE #2047: Support durable invitation links to hunts

jpd236:
Unauthenticated users who open a /join URL will see a welcome screen
asking whether they are a new or existing user. New users are switched
to the enrollment screen; existing users see the login screen as before.

The enrollment screen is updated to support brand new account creation,
rather than just setting a password for an already-created account. If
an invitation code is present, the user can enter an email, or elect to
sign in with a Google account which can be used for future sign ins. As
before, they can set initial profile fields as well. They will be added
to the hunt once they are signed in (by virtue of being redirected to
the original /join link from location state, but now as an authenticated
user).

If a new user goes through the Google sign-in flow, the email and
password fields are subsequently hidden and the credentials stashed
until the user enters the remaining profile fields and completes
enrollment.

Fixes #2047


jpd236:
This depends on https://github.com/deathandmayhem/jolly-roger/pull/2056 (which is included here). No need to review this until you're happy with that, but I figured I'd show where this is all eventually heading. Also happy to combine if you'd prefer.

imports/server/accounts.ts
```diff
@@ -52,26 +70,79 @@ const summaryFromLoginInfo = function (info: LoginInfo) {
   }
 };
 
-Accounts.registerLoginHandler((options: LoginOptions) => {
+Accounts.registerLoginHandler(async (options: LoginOptions) => {
```
ebroder, line 73: Hmm. I think that the current implementation is technically safe, although mostly by accident, but we should do some `check`-style type verification here before passing client-provided input as a query parameter. Otherwise someone could do `huntInvitationCode: {"$ne": null}` or something.

(I think the current implementation is safe because `Google.retrieveCredential` validates that `key` is a string before passing it to the database, and `key` looks to mostly be a random 40-byte-ish string that I think is unguessable by an attacker)

imports/client/components/LoginForm.tsx
```diff
@@ -1,14 +1,48 @@
...
+        const matchResult = invitationPattern.exec(location.state.pathname);
+        if (matchResult?.[1]) {
+          setHuntInvitationCode(matchResult[1]);
+          if (format == null) {
+            setFormat(AccountFormFormat.INVITATION_WELCOME);
+            return;
+          }
+        }
+      }
+      setFormat(AccountFormFormat.LOGIN);
```
ebroder, line 34: When we navigate to the login form in `AuthenticatedPage`, the value that we set for state is unconstrained (react-router just types it as `any`). We can make this cleaner by just putting more structure into state, right?

imports/client/components/AccountForm.tsx
```diff
@@ -17,13 +17,15 @@ import Form from "react-bootstrap/Form";
 import styled from "styled-components";
 import Flags from "../../Flags";
 import updateProfile from "../../methods/updateProfile";
+import { type LoginOptions } from "../../server/accounts";
```
ebroder, line 20: We should pull this into `imports/lib/` if it's going to be used from both client- and server-side code.

imports/client/components/AccountForm.tsx
```diff
@@ -427,51 +498,122 @@ const AccountForm = (props: AccountFormProps) => {
...
+                    onSucceeded={(token: string, secret: string) => {
+                      setSubmitState(AccountFormSubmitState.IDLE);
+                      setGoogleSignInCredentials({
+                        token,
+                        secret,
+                      });
+                    }}
+                    onCanceled={() => {
+                      setSubmitState(AccountFormSubmitState.IDLE);
+                    }}
```
ebroder, line 557: FTR, our strategy has generally been to not put anonymous functions directly in JSX code, and instead wrap everything with `useCallback`. My understanding is that otherwise, anonymous functions will get regenerated on every render of this component, which could trigger spurious re-renders of the components that are receiving those functions as props.

It's also the case that we may be overly concerned about that, but I think my preference is to stick with it as a style preference across the codebase.

ebroder:
Left some comments. `AccountForm.tsx` is definitely...getting mildly out of hand in terms of complexity, although I don't think that's strictly your fault. I don't have concrete suggestions for improvement though. But overall, this seems like an OK direction to me.

imports/client/components/AccountForm.tsx
jpd236: Sure, done.

imports/client/components/LoginForm.tsx
jpd236: If I understand correctly - you're suggesting that in `AuthenticatedPage`, rather than just setting `state={{ pathname, search }}`, we could do something more structured such that here, we could access that more structured data instead of having to extract the invitation code out from the pathname using a regex like this.

But I don't see what AuthenticatedPage has access to that would make that any easier - all it knows is the raw location, from which it pulls the pathname / search. So it seems to me that it would have to do exactly the same sort of path extraction in order to stash it in a more structured form, and if you're going to have to do that somewhere, it seems better to do it where it is needed rather than in a central place.

Ideally we could find a way to apply the path matching from Route.tsx directly to the path so we could actually extract something like a page identifier and the invitation code parameter, but I wasn't seeing an obvious way to do that outside of the standard routing mechanism.

Let me know if I misunderstood.

imports/server/accounts.ts
jpd236: Looks like you added an initial `check` here - I extended it with the new parameters added in this PR. Thanks!

imports/client/components/LoginForm.tsx
ebroder: Oh yeah, that's right.

Hmm. What about not using the authentication helpers on the join route, and having the join route itself redirect to the login/enroll page?

jpd236: Done. I feel a little uneasy from a security perspective about allowing a page to manage authentication itself, but I suppose its justified in this case.

PR #2128, opened 2024-05-05
Verify type of key/secret for Google logins

ebroder:
I don't believe this constitutes a security issue, because `Google.retrieveCredential` validates that `key` is a string before issuing any database queries, and the key is long and random. It may technically be possible (if you get your hand on a key) to swipe someone else's pending login by passing `secret: {"$ne": null}` or similar.

I think this is largely a case of "if you have root, it's easy to get root", but it seems like good practice to clean it up anyway.

### DIFF
--- a/imports/server/accounts.ts
+++ b/imports/server/accounts.ts
@@ -1,4 +1,5 @@
 import { Accounts } from "meteor/accounts-base";
+import { Match, check } from "meteor/check";
 import { Google } from "meteor/google-oauth";
 import { Meteor } from "meteor/meteor";
 import Mustache from "mustache";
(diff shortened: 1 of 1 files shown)


imports/server/accounts.ts
```diff
@@ -58,6 +59,11 @@ Accounts.registerLoginHandler((options: LoginOptions) => {
     return undefined;
   }
 
+  check(options, {
```
zarvox, line 62: If I'm reading this correctly, won't the `options` object always have an `isGoogleJrLogin` field, and therefore have a key which is not specified in this object, and therefore always fail this `check()`?

ebroder: Hmm. Yeah, I think that's right. Fixed

zarvox (approved):

Issue #2117, opened 2024-05-03
Optimize AWS costs, focused on lower-traffic operation


jpd236:
The current CloudFormation scheme scales nicely for higher traffic events, but is a bit overkill for running with low traffic throughout the rest of the year. I'd like to optimize the costs a bit to make keeping it running a bit more palatable, even if it comes at the cost of some overall reliability.

I think the main win here would be to support a single-instance of EC2 so the NLB could be dropped, as that is the most expensive component. If we assume coturn is necessary for full voice chat reliability, I think we have no choice but to support terminating TLS within the EC2 instance and managing our own certificate. (If it were just HTTP(S) traffic, I _think_ we could use Cloudfront, which has a pretty large always-free tier). Amazon's certificate manager can't be used directly within EC2.

So for single instance, I believe we need to:

- Provide a way to pass in a certificate, e.g. from Let's Encrypt. We could potentially try to get fancy and request+renew the certificate within JR, with automated configuration for the linked domain, and to support cycling the cert in the running server, but as a starting point I think I'm fine with letting it be managed externally and bouncing the server every 2-3 months on renewal. Our secure storage (credstash or otherwise) can provide the cert.
- Update haproxy to use the cert when serving traffic on port 443.
- Request an elastic IP and use it for the EC2 instance.
- Update the DNS to point directly to the elastic IP; remove NLB and the scaling group.

As I understand it, this would eliminate a ~$16.20/month cost for the NLB. Data transfer costs look higher from EC2 but only after the first 100 GB which is free.

From there, two other ideas I have for optimization:

- Replace credstash with AWS Systems Manager Parameter Store. The KMS key that credstash depends on is $1/month. While the parameter store uses KMS under the covers, I _believe_ that the only charges for it are the per-request ones which should be negligible here.

- Reduce the number of metrics tracked in Cloudwatch. I haven't actually enabled Cloudwatch yet, so I'm less sure how this works, but I believe that every metric beyond the 10th is $0.30/month. IIUC, with single instance, we'd just have 13 metrics, so $0.90/month, but we could probably drop 3 less essential ones to make it free.

Overall, I think that puts baseline costs at $3.60 for the IP + $0.50 for the Route 52 config + the EC2 costs ($8.35 for t2.micro) = $12.45/month. Higher than a simple Heroku instance, but much more reasonable than the ~$30 I think it would run to right now.

[...]

PR #2132, opened 2024-05-06
Add the ability to disable load balancing to CloudFormation config.

jpd236:
This mode differs as follows:

- The EC2 job is now responsible for terminating TLS. It uses certbot to request Let's Encrypt certificates on startup and renews them in a daily cron job. We use haproxy's stats socket to update the certificate in the running instance without needing to restart it.
- We request an elastic IP and point DNS to it. Whenever the ASG launches a new instance, we run a lambda to associate the elastic IP with the new instance (and disassociate it from any prior instance).

Note: Let's Encrypt limits you to 5 new certificate requests per week - see https://letsencrypt.org/docs/duplicate-certificate-limit/. Since we aren't saving/restoring the configuration data between instances, it's possible to hit this limit when testing deployment changes or making frequent updates that result in restarting EC2 instances. `--test-cert` can be added to the certbot command to point to their staging environment and avoid these limits while testing deployment changes.

This saves at least ~$16.20/month + usage charges on the NLB. The remaining costs for a minimal instance are ~$3.60 for the IP, ~$6.77 for EC2 (t3a.micro) and ~$0.50 for DNS.

See #2117

### DIFF
--- a/cloudformation/jolly-roger.yaml
+++ b/cloudformation/jolly-roger.yaml
@@ -131,16 +131,29 @@ Parameters:
     Description: Comma-separated list of SSH users, each of the form <username>=<ssh_import_id>
     Type: String
     Default: "evan=gh:ebroder,zarvox=gh:zarvox"
-  EnableServing:
-    Description: Whether to bring up the app; set to false to shut it down and save on costs
+  ServingMode:
+    Description: |
+        In "NLB" mode, bring up a load balancer and point the domain to it, allowing scaling to
+        multiple EC2 jobs. TLS termination is handled by the NLB using the certificate referenced
+        in CertificateArn. In "SingleInstance" mode, the domain is pointed directly at a single
+        EC2 instance. The TLS certificate will be automatically managed via Let's Encrypt - be sure
+        to accept their terms of service. In "None" mode, all serving is disabled. Use this to shut
+        down the app and save on costs while keeping persistent resources like the S3 bucket alive
+        and making it simpler to enable serving in the future.
     Type: String
-    Default: "true"
-    AllowedValues: ["true", "false"]
+    Default: "NLB"
+    AllowedValues: ["NLB", "SingleInstance", "None"]
+  CertNotificationEmail:
+    Description: Email to use for certificate abuse and renewal notifications when using Let's Encrypt via the SingleInstance serving mode. Leave blank to disable notifications.
+    Type: String
+    Default: ""
 
 Conditions:
   HavePapertrail: !Not [!Equals [!Ref PapertrailHost, ""]]
   HaveCloudWatch: !Equals [!Ref EnableCloudWatch, "true"]
-  HaveServing: !Equals [!Ref EnableServing, "true"]
+  HaveServing: !Not [!Equals [!Ref ServingMode, "None"]]
+  HaveLoadBalancing: !Equals [!Ref ServingMode, "NLB"]
+  HaveSingleInstance: !Equals [!Ref ServingMode, "SingleInstance"]
 
 Resources:
   LambdaExecutionRole:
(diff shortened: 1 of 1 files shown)


cloudformation/jolly-roger.yaml
```diff
@@ -522,6 +555,27 @@ Resources:
...
+                    #!/bin/bash
+                    set -e
+
+                    # Renew the certificate and update haproxy's offline copy.
+                    if [ -n "${CertNotificationEmail}" ]; then
+                      email_flag="-m ${CertNotificationEmail}"
+                    else
+                      email_flag="--register-unsafely-without-email"
+                    fi
+                    certbot -v certonly --dns-route53 -d ${AppUrl} --non-interactive --agree-tos $email_flag
```
ebroder, line 569: Does this only renew certs that are close to expiry, or will it forcibly renew the certificate daily? (It looks like `certbot renew` is intelligent, but I'm having a hard time telling from the docs if that's true of `certonly` also)

cloudformation/jolly-roger.yaml
```diff
@@ -522,6 +555,27 @@ Resources:
...
+                    mkdir -p /etc/haproxy/certs/
+                    cat /etc/letsencrypt/live/${AppUrl}/fullchain.pem /etc/letsencrypt/live/${AppUrl}/privkey.pem | tee /etc/haproxy/certs/cert.pem > /dev/null
+
+                    # If haproxy is running, perform a live certificate update.
+                    if [ -f /var/run/haproxy/stats.sock ]; then
+                      echo -e "set ssl cert /usr/local/etc/haproxy/certs/cert.pem <<\n$(cat /etc/haproxy/certs/cert.pem)\n" | socat stdio /var/run/haproxy/stats.sock
+                      echo -e "commit ssl cert /usr/local/etc/haproxy/certs/cert.pem" | socat stdio /var/run/haproxy/stats.sock
+                    fi
```
ebroder, line 577: Mildly curious if haproxy will reload the certs if you just send it a SIGHUP.

cloudformation/jolly-roger.yaml
jpd236: Same behavior as renew - it will keep the certificate until the cert is within 30 days of expiration. (Just ran the script manually on my instance to double check).

cloudformation/jolly-roger.yaml
jpd236: I tried it and it did seem to reload them.

Here's more background on the mechanism: https://www.haproxy.com/blog/dynamic-ssl-certificate-storage-in-haproxy. Maybe slightly more complicated than it needs to be; I'm fine with doing something simpler if you'd rather, though this seemed like the most targeted approach and wasn't too bad to get working.

ebroder: Yeah I guess this is probably cleaner - especially in the presence of long-lived connections (which I think a SIGHUP might gracefully shutdown). It's also certainly useful to have the haproxy socket exposed for debugging/inspection purposes.

cloudformation/jolly-roger.yaml
```diff
@@ -136,11 +136,23 @@ Parameters:
     Type: String
     Default: "true"
     AllowedValues: ["true", "false"]
+  EnableSingleInstanceMode:
+    Description: If true, disable the NLB and point the domain directly at a single instance. The TLS certificate will be automatically managed via Let's Encrypt - be sure to accept their terms of service.
+    Type: String
+    Default: "false"
+    AllowedValues: ["true", "false"]
```
ebroder, line 143: Should this just be a tristate? `ServingMode` with `AllowedValues: ["NLB", "Single", "None"]` or something like that? It feels slightly weird to be able to toggle the two independently.

jpd236: Yeah, that makes perfect sense. Done.

I was also thinking about whether to do anything different with AppMinSize/AppMaxSize/AppDesiredCapacity, since some values there don't make sense in single instance mode. But short of some sort of validation rule (is that even possible? Maybe with a lambda?) I didn't think of anything that seemed worthwhile.

ebroder: Yeah I think that's true, but I don't think there's any reasonable way to do that validation. In theory, you could do it at application time with a Lambda, but then you'd get into rolling the stack forwards and backwards and that's always been a little messy with CFN. I don't think there's any way to do it as a pre-application check. I think it's fine to say that that's pushing beyond the limits of what CFN can do.

ebroder (approved):

PR #2140, opened 2024-05-11
Unpin haproxy version.

jpd236:
There have been some bugfixes since 2.9.1 which was apparently leaking connections when 2.9.0 was not. It's unclear to me if this has been fixed, but my instance has run for a few days and has no open connections:

```
$ echo "show info" | sudo socat stdio /var/run/haproxy/stats.sock  | grep Conns 
CurrConns: 0
CumConns: 893
MaxSslConns: 0
CurrSslConns: 0
CumSslConns: 876
```

And I have seen sporadic load failures and at least one segfault in the logs with the pinned version, so it seems like we should try unpinning the version.


zarvox (approved):
LGTM given the context and history but @ebroder is the canonical owner of the cloudformation config so I'll let him opine :)

ebroder (approved):

PR #2166, opened 2024-05-31
Stop taking chai updates

ebroder:


ebroder:
Copying from #2154:

> Bah. Looks like another dependency dropping support for Node 16 although this one seems to be unintentional. In particular: chai depends on assertion-error, which uses [deno bundle](https://docs.deno.com/runtime/manual/tools/bundler) to compile TS to JS, which generates code using static class initialization blocks, which node 16 doesn't support. However, both chai and assertion-error claim to support node >= 12. Based on the overall tenor of the JS/node community, I kind of assume that if I brought this to their attention they would wontfix it and change the stated compatibility versions.

zarvox (approved):

Issue #2161, opened 2024-05-25
Occasional 429 errors (Too Many Requests) in Drive activity polls


jpd236:
My server logs show occasional instances of 429s in the Drive activity queries:

```
code: 429,
  errors: [
    {
      message: 'Resource has been exhausted (e.g. check quota).',
      domain: 'global',
      reason: 'rateLimitExceeded'
    }
  ]
}
```

The overall error rate looks quite low in the console (0.02%), though it is a bit noisy. From what I can tell we're doing between 10-15 queries per minute, and the limit per minute per user is 100.

It seems like there might be some retries here already; I wonder if we just need to back off a bit more in this case.

zarvox:
@jpd236 Do you consider this done as a result of #2196 being merged, or do you think there's more to do here?

Issue #2160, opened 2024-05-25
Internal server error on join with Google Account


jpd236:
Someone joining our test hunt via an invitation code (and signing up as a new user with a Google Account) hit an error in `ensureHuntFolderPermission` [here](https://github.com/deathandmayhem/jolly-roger/blob/d6deccbc76746b705ff2138948145d6e71805a97/imports/server/addUserToHunt.ts#L163). Full stack:

```
Exception while invoking method 'Hunts.methods.acceptHuntInvitationCode' Error: Internal Error
    at Gaxios._request (/built_app/bundle/programs/server/npm/node_modules/googleapis-common/node_modules/gaxios/build/src/gaxios.js:141:23)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
 => awaited here:
    at Function.Promise.await (/built_app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/promise_server.js:56:12)
    at imports/server/gdrive.ts:134:3
    at /built_app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40
 => awaited here:
    at Function.Promise.await (/built_app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/promise_server.js:56:12)
    at imports/server/gdrive.ts:249:3
    at /built_app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40
 => awaited here:
    at Function.Promise.await (/built_app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/promise_server.js:56:12)
    at imports/server/addUserToHunt.ts:163:7
```

Seems like it might have just been a transient error. The folder was shared with them when I checked shortly afterwards.

Ideally we'd have retries and/or something a bit more graceful here.

PR #2196, opened 2024-06-28
Retry more Google API failures.

jpd236:
The default retry policy (documented
[here](https://github.com/googleapis/gaxios/#request-options)) does not retry POST requests. But most Google API calls use POST requests under the covers, including read-only APIs like [Drive
Activity](https://developers.google.com/drive/activity/v2/reference/rest/v2/activity/query). So we can also ask to retry POST requests.

We retain other aspects of the default policy, such as the number of attempts (3) and the HTTP status codes to retry (100-109, 408, 429, 5xx). If a request succeeds on a retry, the failure is kept silent and does not pollute the logs.

The main risk is that a POST method might mutate some state despite returning a failure, and that inline retries might handle that incorrectly or otherwise make things worse than retrying the full flow from the ground up. This doesn't seem likely to be an issue with the current set of API calls, but if desired, we could narrow this change to specific known-safe requests.

See #2160
See #2161

### DIFF
--- a/imports/server/googleClientRefresher.ts
+++ b/imports/server/googleClientRefresher.ts
@@ -10,6 +10,7 @@ import type { people_v1 } from "@googleapis/people";
 import { people } from "@googleapis/people";
 import type { script_v1 } from "@googleapis/script";
 import { script } from "@googleapis/script";
+import type { RetryConfig } from "gaxios";
 import type { SettingType } from "../lib/models/Settings";
 import Settings from "../lib/models/Settings";
 
@@ -32,11 +33,18 @@ class GoogleClientRefresher {
 
   private oauthCredentialCursor: Mongo.Cursor<SettingType>;
 
+  private retryConfig: RetryConfig;
+
   constructor() {
     this.drive = undefined;
     this.oauthClient = undefined;
     this.oauthConfig = undefined;
     this.oauthRefreshToken = undefined;
+
+    this.retryConfig = {
+      // Retry POST failures as well, since most APIs use POST requests under the covers.
+      httpMethodsToRetry: ["GET", "PUT", "HEAD", "OPTIONS", "DELETE", "POST"],
+    };
 
     // Watch for config changes, and refresh the gdrive instance if anything changes
     this.oauthConfigCursor = ServiceConfiguration.configurations.find({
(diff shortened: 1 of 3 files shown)


ebroder (approved):
LGTM

Noting just for the record that we may end up needing to add some error handling logic around stuff like Drive permissions, where the APIs may not be idempotent but a "resource already exists" error shouldn't cause us to blow up. But this seems strictly better than that case

ebroder:
Not sure why builds aren't running on this commit. Maybe amend and re-push and see if it kicks off?

jpd236:
Thanks!

> Noting just for the record that we may end up needing to add some error handling logic around stuff like Drive permissions, where the APIs may not be idempotent but a "resource already exists" error shouldn't cause us to blow up. But this seems strictly better than that case

Yep, definitely possible there'll be more to do here. FWIW, I did try API Explorer with the Drive `permissions.create` API, and it seems to be idempotent in that repeated calls return 200 with the same Permission object. So this does feel like it should be enough to help with #2160 (and #2161, which is readonly).

Not sure why builds aren't triggering here, but on my fork I'm seeing an error:

```
Error: imports/server/googleClientRefresher.ts(111,7): error TS2769: No overload matches this call.
  The last overload gave the following error.
    Type 'import("/home/runner/work/jolly-roger/jolly-roger/node_modules/gaxios/build/src/common").RetryConfig' is not assignable to type 'import("/home/runner/work/jolly-roger/jolly-roger/node_modules/googleapis-common/node_modules/gaxios/build/src/common").RetryConfig'.
      Types of property 'onRetryAttempt' are incompatible.
        Type '((err: import("/home/runner/work/jolly-roger/jolly-roger/node_modules/gaxios/build/src/common").GaxiosError<any>) => void | Promise<void>) | undefined' is not assignable to type '((err: import("/home/runner/work/jolly-roger/jolly-roger/node_modules/googleapis-common/node_modules/gaxios/build/src/common").GaxiosError<any>) => void | Promise<void>) | undefined'.
          Type '(err: import("/home/runner/work/jolly-roger/jolly-roger/node_modules/gaxios/build/src/common").GaxiosError<any>) => void | Promise<void>' is not assignable to type '(err: import("/home/runner/work/jolly-roger/jolly-roger/node_modules/googleapis-common/node_modules/gaxios/build/src/common").GaxiosError<any>) => void | Promise<void>'.
            Types of parameters 'err' and 'err' are incompatible.
              Property '[GAXIOS_ERROR_SYMBOL]' is missing in type 'import("/home/runner/work/jolly-roger/jolly-roger/node_modules/googleapis-common/node_modules/gaxios/build/src/common").GaxiosError<any>' but required in type 'import("/home/runner/work/jolly-roger/jolly-roger/node_modules/gaxios/build/src/common").GaxiosError<any>'.
```

googleapis-common has a conflicting dependency on an older version of gaxios but I'm not immediately sure how to resolve it. Locally it seems like reinstalling googleapis-common fixed it, but the server build is failing even after I blew away the cache... Need to dig further, but let me know if you have any ideas.

jpd236:
Fixed the dependency conflict by running `npm dedupe`. That resulted in many unrelated changes to package-lock.json. Hopefully that's okay - let me know if there's some other way to handle it.

PR #2197, opened 2024-06-29
Enable a "minimal" mode for CloudWatch monitoring.

jpd236:
In this mode, only memory and swap usage is captured. At two metrics per instance, this usage should remain in the free tier (10 custom metrics total). Total CPU usage is available as a built-in metric in EC2, so this should give a reasonable picture of CPU/memory usage for basic monitoring.

The additional metrics around detailed CPU, disk, and network are available in the "detailed" mode (the default).

This also fixes an issue with the indentation in the YAML when CloudWatch is enabled. We also have to be careful when creating the CloudWatch JSON config via string substitution, which unfortunately makes for a somewhat cludgier YAML file.

See #2117


ebroder (approved):
LGTM!

PR #2259, opened 2024-09-01
Fix build

ebroder:


zarvox:
I think this is probably obsoleted by https://github.com/deathandmayhem/jolly-roger/pull/2272

PR #2272, opened 2024-09-15
Fix the build.

jpd236:
The attempt to pin a specific version of the meteor tool never actually worked, which means the build broke once Meteor 3.0 was officially released.


zarvox (approved):
Ah!  Thanks so much!

Issue #2274, opened 2024-10-01
"Add puzzle" dialog contents should be cleared after add/close


jpd236:
When you fill out fields in the "Add puzzle" dialog and then either close it or add the puzzle, the next time you click "Add puzzle", the previous contents will still be present in the dialog.

It seems to me that the contents should be cleared. It would be relatively rare for any of the fields to carry over from puzzle to puzzle - maybe tags, but even then it's rare, and autocomplete helps a lot. Also, refreshing the page clears the old data, which makes it feel more like a bug than intended behavior.

zarvox:
Yeah, I think the current behavior here is coincidental more so than intended.  I've felt similarly about wanting successful submission at least to clear title and URL, though I never really bothered getting around to it because when keying in e.g. the first 3-5 puzzles for a new round, it's often been convenient to keep the `group:` tag(s).  But I agree that autocomplete probably mostly covers that, and it'd feel neater/more consistent.

PR #2287, opened 2024-10-03
Reset puzzle modal form upon successful submission.

jpd236:
See #2274

### DIFF
--- a/imports/client/components/PuzzleListPage.tsx
+++ b/imports/client/components/PuzzleListPage.tsx
@@ -244,9 +244,16 @@ const PuzzleListView = ({
         return;
       }
 
-      createPuzzle.call({ docType, ...rest }, callback);
+      function onAddComplete(error?: Error) {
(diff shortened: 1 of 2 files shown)


jpd236:
This causes a regression that can lead to errors when saving existing puzzles. Simple repro steps:

- From the puzzle list page, click the edit button on a puzzle.
- Save without making any changes
- Repeat above steps again

I think the problem is that upon successful save, the form's title state is set to "" (and titleDirty set to false) in the onSubmit callback. So if we reload the dialog for the same puzzle and try to save again, the title doesn't get reset.

Need to find a better way to clear state here.

jpd236:
I think this is now fixed. PuzzleModalForm now exposes a reset() function to reset the form state. PuzzleListPage calls reset upon successful puzzle creation. This leaves the behavior unchanged for editing existing puzzles, but keeps the clearing for new puzzles.

zarvox (approved):

Issue #2275, opened 2024-10-01
Better deduplication of new puzzles


jpd236:
When multiple people on a team are permitted to add puzzles, it's all to common to accidentally add two entries for the same puzzle, which creates confusion.

I don't know that it'd be safe to require uniqueness of URL. I could imagine wanting to create multiple entries for a puzzle that's all present under one URL (It's definitely not safe to require uniqueness of title, as we've learned from GPH over the weekend...). But it feels to me like we could do more for the more common case that two entries represent the same puzzle rather than use the same URL but are intentionally distinct.

My initial thought is to detect duplicates by URL (or title?) and show a warning if there is one with a link to the other entry. From there, the user could opt to bypass the warning and add the new entry if the duplicate is expected. I think that would help a lot in preventing unintended duplicates without interfering heavily with regular workflows.

zarvox:
Wanting to prevent this failure mode (more often) is reasonable; I wouldn't be opposed to warning when we detect a dupe URL or title.  Did you find that these were usually people racing to add the same puzzle at around the same time (right as it unlocks), or more failure to notice the puzzle was already entered at some later time?

FWIW, we've usually (mostly) solved the former problem in the past by having a single person responsible for puzzle entry at any particular time during the hunt.  I could also imagine an orthogonal approach to this problem shaped like having basically a stealable lock that can be held by at most one person at any particular time, and which you have to steal before you can fill out the add new puzzle form (and if it gets stolen out from under you then you'd see that happening).  Then at least people can notice that they're all trying to do the same thing.

The latter approach might also help deal with the fact that since the Google interactions can sometimes add some seconds of latency to puzzle creation, it's entirely possible that even with perfect dupe detection/warning in the form, we might still wind up with multiple requests in flight to create the same new puzzle from different puzzle-adders at the same time.  So the "teamwide lock" idea appeals somewhat because it helps prevent that situation, but then it also means only one person can be keying in new puzzles at a time, which might add unwanted friction.  Thoughts?

jpd236:
This was typically due to multiple people racing when new unlocks happened.

For MITMH, with our large team size, we do typically have a rotation of on-duty puzzle admins, so it's not as much of a problem there. Enshrining that with a formal "lock" would be a nice enhancement though we've managed okay without it in the past.

[...]

PR #2313, opened 2024-10-24
Deduplicate new puzzles by URL.

jpd236:
When adding a puzzle, we start a transaction to atomically check if a puzzle already exists with that URL and only add it if either no such puzzle exists or if the client indicates that duplicate URLs are allowed. In the event of a transaction failure, we delete the Google document that was created optimistically before saving the puzzle.

On the client side, we add a new checkbox to allow duplicate URLs, unchecked by default. If we encounter a duplicate URL error, we show an error indicating that the puzzle may be a dupe and telling the user what to check if the duplicate URLs are intentional.

See #2275


jpd236:
This should be functional, but there are some less-polished (IMO) elements as I wanted to get feedback before moving forward. At least the following:

- The UX. Do we always want to show this checkbox as is done here? Or should it be a more interactive flow, e.g. a (nested) dialog indicating that the puzzle is potential duplicate and offering the user a simple Continue/Abort option? Should we provide more details about the duplicate? I kept it simple for now.
- The error passing. I just throw a Meteor.Error with an unusual (but reasonable) HTTP error code and detect it client side. This should probably be something a bit more semantically meaningful. What's the canonical way of doing this?
- The transaction. This seems like the first use of a transaction in the codebase. It seems reasonable at the face of it, but I'm pretty unfamiliar with MongoDB and am not sure about the semantics. In other places, we just use locks - that could also be reasonable here, e.g. with a hunt-specific puzzle creation (and updating?) lock, if that's preferable.

zarvox:
> The UX. Do we always want to show this checkbox as is done here? Or should it be a more interactive flow, e.g. a (nested) dialog indicating that the puzzle is potential duplicate and offering the user a simple Continue/Abort option? Should we provide more details about the duplicate? I kept it simple for now.

I'm in favor of keeping things simple, but given how relatively rare we expect this case (actually wanting to create a duplicate Puzzle object and sheet for an existing URL) to be, I'd even say "don't clutter the form with the checkbox unless you've already tried and failed the method once." kinda like how we handle duplicate guess/answer submission.

> The error passing. I just throw a Meteor.Error with an unusual (but reasonable) HTTP error code and detect it client side. This should probably be something a bit more semantically meaningful. What's the canonical way of doing this?

Huh, while the Meteor docs suggest that `error`, `reason`, and `details` should all in fact be strings ( https://docs.meteor.com/api/meteor.html#Meteor-Error ), the rest of our existing codebase does not conform to that guidance, at all.  I'm fine with sticking to a reasonable HTTP code; it's what we do elsewhere and if we want to revise that we can refactor separately.

> The transaction. This seems like the first use of a transaction in the codebase. It seems reasonable at the face of it, but I'm pretty unfamiliar with MongoDB and am not sure about the semantics. In other places, we just use locks - that could also be reasonable here, e.g. with a hunt-specific puzzle creation (and updating?) lock, if that's preferable.

I am unfortunately a little wary of the transaction.  It would indeed be the first such use in the codebase.  Some stuff I'm reading from a Meteor community package for mongo transactions suggests that [on Mongo Atlas' free tier (M0), writes from transactions fail to appear in the oplog as seen by Meteor](https://docs.meteor.com/community-packages/mongo-transactions#using-mongo-atlas-as-your-db), which gives me some pause as IIRC D&M uses Atlas' free tier in the off-season.

It's not entirely clear to me *why* that is the case and if that's something on Meteor or Mongo to replace, but it seems like enough of a yakshave to dig into that I might nudge you towards the lock-based approach instead.

For lock naming, you could use something like `hunts:${huntId}:puzzle-url:{url}`.  I considered for a moment that you might update the `updatePuzzle` method to hold that lock, but actually I think that if someone is changing the URL that should just be allowed to just work.  I just can't imagine a flow where updating the url should fail because there's another puzzle with that URL, and handling a new failure mode would mean we'd have to change the "update puzzle" UI as well.

You might also decline to bother with locking if url is unset, or empty.  I can't imagine wanting to allow empty URLs but restrict whether or not you can have more than one of them.

Lastly, in either approach, I'd recommend doing an additional opportunistic check for expected conflict at the top of createPuzzle, before ensuring the document exists or doing the transaction-guarded action, because if there's already a Puzzle with the same url and we'd reject the write, we might as well save ourselves the nontrivial latency of creating and then deleting a sheet via the Google Drive API.  This would give a better user experience for the (likely common) case of "the operator didn't notice the other puzzle was already entered and the writes were not issued within $OPLOG_TAILING_LATENCY of each other".

imports/server/methods/createPuzzle.ts
```diff
@@ -81,7 +91,39 @@ defineMethod(createPuzzle, {
...
 
-    await Puzzles.insertAsync(fullPuzzle);
+    // In a transaction, look for a puzzle with the same URL. If present, we
+    // reject the insertion unless the client overrides it.
+    const client = MongoInternals.defaultRemoteCollectionDriver().mongo.client;
+    const session = client.startSession();
+    try {
+      await session.withTransaction(async () => {
+        if (url) {
+          const existingPuzzleWithUrl = await Puzzles.collection
```
zarvox, line 101: I think it would be good to **also** do this sort of check **before** the `ensureDocument` call above -- in the case where there is already another puzzle with the same URL (i.e. the operator missed that the puzzle had already been entered), we can reject the request quickly and without causing side-effects or unnecessary Google Drive API calls (which are often slow relative to our internal method invocations).

imports/client/components/PuzzleModalForm.tsx
```diff
@@ -288,6 +330,19 @@ const PuzzleModalForm = React.forwardRef(
         </FormGroup>
       ) : null;
 
+    const allowDuplicateUrlsCheckbox =
+      !puzzle &&
+      typeof allowDuplicateUrls === "boolean" &&
+      confirmingDuplicateUrl ? (
+        <FormCheck
```
zarvox, line 337: It's good to specify a unique `id` prop to the `FormCheck` here, so that the `<label>` it ultimately produces will reference the `<input>`'s id, which makes it possible to click anywhere on the label text to toggle the checkbox instead of having to hit just the square of the checkbox.  This can be achieved in other contexts by specifying `controlId` to the containing `FormGroup`, but here we've put multiple controls into the same form group for reasonable reasons (the checkbox's meaning is bound to the url) so we should do it on the `FormCheck` itself instead.

If we're being consistent with other inputs, something like `id="jr-new-puzzle-allow-duplicate-urls"` would be perfect.

```suggestion
        <FormCheck
          id="jr-new-puzzle-allow-duplicate-urls"
```

zarvox (commented):
Broadly looks good; two small nits and one question for @ebroder about indexes; if he wants all queries indexed I'll ask you to append `, url: 1` to the `.addIndex` call in `imports/lib/models/Puzzles.ts`.

Issue #2378, opened 2024-12-21
Double-encoding of HTML entities in certain chat messages


jpd236:
The chat message renderer uses `marked.lexer` to tokenize Markdown chat messages; in some cases, we reference the resulting `token.text` field to obtain just the text and render it inside an HTML tag. We then have our own rendering logic. However, as noted at https://github.com/markedjs/marked/discussions/1737, the lexer will encode HTML entities, rather than leaving that to the renderer. React applies its own encoding for dynamic content, which results in double-encoding.

For example, sending a code block (inside backticks) with an apostrophe inside the code will result in the user-visible message containing the HTML entity for apostrophe.

We can fix this by decoding the output of `marked.lexer` whenever it is used.

PR #2379, opened 2024-12-21
Fix double-encoding of HTML entities in certain chat messages.

jpd236:
See #2378


imports/client/components/ChatMessage.tsx
```diff
@@ -95,9 +100,11 @@ const MarkdownToken = ({ token }: { token: marked.Token }) => {
     ));
     return <del>{children}</del>;
   } else if (token.type === "codespan") {
-    return <code>{token.text}</code>;
+    const decodedText = he.decode(token.text);
+    return <code>{decodedText}</code>;
   } else if (token.type === "code") {
-    return <StyledCodeBlock>{token.text}</StyledCodeBlock>;
+    const decodedText = he.decode(token.text);
+    return <StyledCodeBlock>{decodedText}</StyledCodeBlock>;
```
zarvox, line 107: I believe code blocks (multi-line blocks encased in triple backticks, which render in grey in the preview window) *don't* want to decode text, as in my testing they appear to behave correctly already.  (I agree `codespan`s do need the added decoding here).

A testcase: 

Enter in the input area (with the two three-backtick lines as rendered on the GitHub website):

````
```
it's okay
Q & A
A manually-escaped ampersand: &amp;
```
````

The preview should show a grey background rather than text in red.  Send the message.  I'd expect it to read back exactly as written by the user, so you should still see `&amp;` when rendered by the markdown renderer, but with this patch the final line just shows `&` instead of what the user wrote (`&amp;`).

Is that the behavior you think would make the most sense here?  Am I missing something else that the existing implementation gets wrong with code blocks (vs. codespans)?

PR #2423, opened 2025-02-22
Local voice mute

jimsug:
My first non-accidental PR!

This introduces the concept of local voice mute. The main use case I can think of is when multiple users are in the same location and can hear each other without the call, but want to use it to collaborate with users in other locations.

A user can mute any other user, and this is tracked separately to the other user's RTC mute status. We also still display the spectrogram when the user speaks even when they are locally muted, as a signal that they are communicating something (and as a gentle cue for the user to unmute them if needed).

For simplicity (and I think alignment with encouraging collaboration), I've implemented this as a client-side mute only. This setting does not persist at all; each time you connect to a call, you will need to re-mute anyone you've muted before. 

I'm considering retaining this setting for specific users across calls (at least in the same hunt), since (I _think_ most of the time) when you've muted a user because you're in the same space, you will probably remain in the same space, and having to re-mute someone repeatedly could be frustrating.

On one hand, having to unmute someone might be frustrating; on the other, it's only a couple of clicks away. Persisting mute might also be a smidge more complex, but I think it would be very doable if desired.

### DIFF
--- a/imports/client/components/CallSection.tsx
+++ b/imports/client/components/CallSection.tsx
@@ -216,7 +216,17 @@ type RemoteMuteConfirmModalHandle = {
 
 const RemoteMuteConfirmModal = React.forwardRef(
   (
-    { peerId, name }: { peerId: string; name: string },
+    {
+      peerId,
+      name,
+      isLocallyMuted,
+      onLocalMuteToggle,
+    }: {
+      peerId: string;
+      name: string;
+      isLocallyMuted: boolean;
+      onLocalMuteToggle: () => void;
+    },
     forwardedRef: React.Ref<RemoteMuteConfirmModalHandle>,
   ) => {
     const [visible, setVisible] = useState(true);
@@ -227,6 +237,7 @@ const RemoteMuteConfirmModal = React.forwardRef(
     const [disabled, setDisabled] = useState(false);
     const [error, setError] = useState<Error>();
     const clearError = useCallback(() => setError(undefined), []);
+    const [localMuteEnabled, setLocalMuteEnabled] = useState(isLocallyMuted);
 
     const mute = useCallback(() => {
       mediasoupRemoteMutePeer.call({ peerId }, (err) => {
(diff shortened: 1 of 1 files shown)


imports/client/components/CallSection.tsx
```diff
@@ -216,7 +216,17 @@ type RemoteMuteConfirmModalHandle = {
 
 const RemoteMuteConfirmModal = React.forwardRef(
```
ebroder: We should probably rename this given the expanded functionality? Maybe `PeerMuteConfirmModal` or something?

imports/client/components/CallSection.tsx
```diff
@@ -227,6 +237,7 @@ const RemoteMuteConfirmModal = React.forwardRef(
     const [disabled, setDisabled] = useState(false);
     const [error, setError] = useState<Error>();
     const clearError = useCallback(() => setError(undefined), []);
+    const [localMuteEnabled, setLocalMuteEnabled] = useState(isLocallyMuted);
```
ebroder: Do you actually need separate in-component state here? Can you just use `isLocallyMuted` directly?

imports/client/components/CallSection.tsx
```diff
@@ -240,32 +251,61 @@ const RemoteMuteConfirmModal = React.forwardRef(
...
+
+    useEffect(() => {
+      setLocalMuteEnabled(isLocallyMuted); // Sync modal's local state with prop updates
+    }, [isLocallyMuted]);
+
     const modal = (
       <Modal show={visible} onHide={hide}>
         <Modal.Header closeButton>
-          <Modal.Title>Mute {name}</Modal.Title>
+          <Modal.Title>Manage audio for {name}</Modal.Title>{" "}
```
ebroder: That space at the end of the title looks spurious

ebroder (commented):
A few nits, but overall this seems fine. Sorry again for not reviewing this when I originally promised :)

PR #2427, opened 2025-08-27
Replace react-split-pane/SplitPanePlus with SplitPaneMinus

zarvox:
`react-split-pane` is so old that it declares a react dependency that is version-incompatible with what we actually use, and does not appear to be likely to see an updated release.

In the interest of removing the `--legacy-peer-deps` requirement to install anything (or update deps, or do other npm actions), this patch reimplements the currently-used surface area of our react-split-pane wrapper, `SplitPanePlus`, as `SplitPaneMinus`.

This does remove some latent unused functionality, like collapsible panes and relative (percentage-based) resizing, but we can always reimplement those features later if we decide we want to use them again (and perhaps more easily than how they were done before, since we can modify the underlying code instead of just wrapping it).

Unlike react-split-pane, this reimplementation is TypeScript-native. Also unlike react-split-pane, we're willing to rely on `PointerEvent`s existing, so thanks to `setPointerCapture`, we can avoid writing all the event handlers twice and still have them work on touch surfaces.

This also gives us an opportunity to remove more of the standalone `scss` in favor of `styled-components`, which is more in line with our current approach to managing layout and styles.

Tested on:
* Chrome
* Firefox
* Safari (desktop)
* Safari (iPad)


ebroder (approved):
I agree this code seems basically cromulent and I wasn't enough of an expert on react-split-pane to have strong opinions about any divergences. I'm satisfied if you've tested it.

PR #2392, opened 2025-01-02
Initial implementation of a browser extension to add puzzles.

jpd236:
(More details in individual commit messages - this PR is a stack of a few open requests).

See #237

PR #2433, opened 2025-09-06
Improve API key management

zarvox:
This is a patch series that aims to make API key management possible through the user settings page.  It's structured as a series of mostly-orthogonal commits, and I recommend reading it commit-by-commit.

Some background: the `fetchAPIKey` and `rollAPIKey` methods were first written in probably 2017, and were intended to be called from the browser console, and besides poking around in the database, there was no other way to interact with APIKeys.  The API, at present, includes a single method which was used to restrict access controls on a puzzle-of-the-day site to team members (identified by their email address).  That design has remained in place to this day, even though it was very much a minimum-viable-product.

Now, we have more users and some teams which are excited about interacting with jolly-roger programmatically, and the API will need to grow to support them.  A notable desired use case highlighted in #2392 which aims to make Puzzle creation more expedient.  Other teams may aim to implement slash-commands in their teamwide chat solution to create Puzzles or submit Guesses.

As the current API key management UI is effectively nonexistent, and in keeping with our broad goal of trying to make most server management something we can perform through the UI, we should improve API key management.

The existing max-of-one-key-per-user limit enforced by the existing methods is not particularly good for operations -- it encourages reuse of keys in the presence of multiple integrations, and also makes key rotation without integration downtime effectively impossible.  This patchset switches to a model where we allow each user to create arbitrarily many API keys, and surfaces time of creation and use to help developers better understand what keys are likely in active use.

I extracted the first two commits from @jpd236's diff in #2391 and preserved his authorship on those patches.

In the process of building the UI, I also discovered a latent bug in our `RelativeTime` component where we would fail to promptly update the displayed time when the `date` props (or other props) changed, only changing the rendered text once the timer for updating the text (based on the delta) would fire.

I'm open to adding limits to the number of keys we allow a user to generate, tweaking the throttling around `lastUsedAt` updates, or potentially limiting API key creation (and the corresponding UI) to server admins if we think that's a meaningful restriction.

I also considered adding user-specified labels to API keys, but decided I'd start with surfacing creation/last used timestamps and see if that gets us far enough for our purposes now.

Some screenshots of the UI:

Listing keys:

Screenshot 2025-09-06 at 00 32 10

Last used times help distinguish keys:

Screenshot 2025-09-06 at 00 32 48

We confirm deletion with a modal:

Screenshot 2025-09-06 at 00 39 13


imports/server/methods/createAPIKey.ts
```diff
@@ -0,0 +1,23 @@
...
+    return arg;
+  },
+
+  async run() {
+    check(this.userId, String);
+
+    const key = await APIKeys.insertAsync({
+      user: this.userId,
+      key: Random.id(32),
+    });
```
ebroder, line 19: Hmm this is maybe not a novel issue, but it looks like we don't currently have a unique index on key, which seems like an oversight: https://github.com/deathandmayhem/jolly-roger/blob/94427dcfa1610eb15098e6456a2c987fdfaf2fdf/imports/server/models/APIKeys.ts#L15

ebroder (approved):

imports/server/methods/createAPIKey.ts
zarvox: Hmmm, yeah, it would be better if that were unique.  The trivial addition of `, { unique: true }` to that `addIndex` call results in an index spec conflict for existing servers though (because the existing non-unique index is also implicitly named `key_1`):

```
W20250906-15:46:30.469(-7)? (STDERR)     errmsg: 'An existing index has the same name as the requested index. When index names are not specified, they are auto generated and can cause conflicts. Please refer to our documentation. Requested index: { v: 2, unique: true, key: { key: 1 }, name: "key_1" }, existing index: { v: 2, key: { key: 1 }, name: "key_1" }',
```

We may need to adjust our implementation to either handle replacing indexes in-place, or provide a way to provide a not-autogenerated index name to allow avoiding name conflicts.  I filed #2434 for future follow-up.

Issue #1948, opened 2024-01-08
Feature request: Simple REST(?) API for adding puzzles


egnor:
We have users (https://github.com/dougz) who like to write little scrapers to auto-populate our solving aids from the Hunt site. In recent years, Hunt site runners have been offering puzzle lists in JSON or otherwise machine-readable format which is amazing!

To make this easier to do it would be neat to have some very simple interface which scripts could use to add puzzles. (And maybe enumerate existing puzzles, so they could check for dupes.)

ebroder:
Huh, I don't think I had realized there was a JSON dump. We've had an open ticket (#237) to build this kind of scraping into Jolly Roger for ages, but I'd also be pretty open to exposing an API for external scripting.

We do have some basic REST API structure. Right now it exists to allow a third-party team website to use Jolly Roger as an authz source, but it shouldn't be terribly hard to extend. Some basic notes on how it works (since we should - but don't currently - have any docs):

- You can grab your API key by running `await Meteor.callAsync('fetchAPIKey', {})` in the console. There's also a `rollAPIKey` that does what it sounds like.
- The API lives under `/api` and takes HTTP Bearer auth.
- The code for the API starts [in `/imports/server/api-init.ts`](https://github.com/deathandmayhem/jolly-roger/blob/218c0be267be6dc927e2fd1d82eed9bc41b55a19/imports/server/api-init.ts) and threads through from there.

egnor:
@ebroder (get your barf bag out) as a stopgap, would just adding an entry to the database do the right thing as far as adding a puzzle? (obviously very brittle and to be replaced with a proper API in future) or is there enough denormalized content around puzzle creation (not to mention sheet creation, etc) that it would be way too hard?

ebroder:
Hmm. It might work? I definitely wouldn't recommend it. Main thing I'll note is to make sure you get the types right (although the database schema should enforce it). I think the main functionality you'll lose is the ability to create puzzles with Google Docs instead of Sheets - may not be important to you but we tend to use it for meta-hunt things.

If you really want to hack something together, I might try to find a standalone DDP client (a la https://github.com/Gregivy/simpleddp) and invoke the Meteor methods that way.

jpd236:
[...]

Issue #237, opened 2019-01-29
Make new puzzle entry easier


ebroder:
Instead of requiring operators to enter details about puzzles by hand, we should try to do something smarter. There are a few options:
- We could assume that there will be a puzzle list page (a la http://web.mit.edu/puzzle/www/2019/puzzle.html) and try to scrape it somehow. (It might be hard to make assumptions about what the structure of that markup looks like. Maybe we need some hunt-time inputs as to what portions of the page to crawl, like a CSS selector?)
- We could provide a bookmarklet or Chrome extension or some other client-side mechanism for operators to use when they have a puzzle page opened.
- ???

zarvox:
Another light automation idea here: we could make the puzzle entry form start out with the URL input at the top, and when that input is changed, attempt a `fetch()` on the URL, and attempt to autofill the puzzle title field with a suggestion by extracting the `<title>` or a plausible `<h1>` from the loaded puzzle page, which could potentially reduce copy-pasting/typing.

I imagine operators would still probably need to do some amount of trimming of excess content from autofilled titles, since I suspect they may wind up being like "Puzzle title -- Maybe round title -- Mystery Hunt 2018".

ebroder:
I could see that working, although I will note that it could break flow for the operator a little bit, as basically while I'm entering a puzzle, I have to sit while I wait for JR to fetch and process the URL, during which time I can't do anything. I can see that coming up especially when there are multiple new puzzles released - today there aren't any delays from JR, just delays from me going back and forth between tabs and copy/pasting things.

zarvox:
After poking at it briefly, I think CORS means my idea (doing the `fetch()` from the user's browser) isn't viable in general, and doing it server-side would require the server to keep credentials/cookies, which I think exceeds the complexity budget I'd given myself.

A bookmarklet might still be worth investigating.  We could make it so that if your URL had a particular pattern in the fragment we auto-show the add puzzle modal and fill it with other values from the fragment, and then we just make the bookmarklet stuff the title & location in there?

jimsug:
fwiw I did a bit of a naive implementation of this in [jimsug/feature-guess-puzzle-name](https://github.com/jimsug/jolly-roger/tree/feature-guess-puzzle-name), where it tries to guess the puzzle title from the URL, but it assumes that the puzzle name will be sluggified in the last part of the URL path, which might not always be the case.

jpd236:
[...]

PR #2437, opened 2025-09-11
Add new APIs for creating puzzles.

jpd236:
- Propagate user ID to API methods in authenticator
- Add API to list hunts
- Add API to list tags for a hunt (for autocompletion)
- Add API to create a puzzle

Most of the complexity is in the latter. The bulk of the implementation is shared with the existing flow for creating puzzles, but there is some jank:

- Wherever we were looking at this.userId, which is populated for Meteor method execution but not this API path, we need to pass an explicit userId instead. One pain point is the auto-inferred userId for the createdBy column of many tables.
- We check the types of the input in both the Meteor method (to take advantage of the validate scaffolding) as well as the shared logic (so it's there for API calls).

See #237
See #1948

### DIFF
--- a/imports/server/api/resources/hunts.ts
+++ b/imports/server/api/resources/hunts.ts
@@ -29,13 +29,10 @@ hunts.get(
       return;
     }
 
-    const allHunts = Hunts.find({}, { sort: { createdAt: -1 } });
-    const userHunts = [];
-    for await (const hunt of allHunts) {
-      if (user.hunts?.includes(hunt._id)) {
-        userHunts.push(renderHunt(hunt));
-      }
-    }
+    const userHunts = await Hunts.find(
+      { _id: { $in: user.hunts ?? [] } },
+      { sort: { createdAt: -1 } },
+    ).mapAsync((hunt) => renderHunt(hunt));
     res.json({
       hunts: userHunts,
     });


imports/server/api/authenticator.ts
```diff
@@ -39,7 +40,13 @@ const authenticator: express.Handler = expressAsyncWrapper(
       void APIKeys.updateAsync({ _id: key._id }, { $set: { lastUsedAt: now } });
     }
 
-    next();
+    // eslint-disable-next-line no-underscore-dangle
```
ebroder: Oh and you should be able to drop this if you rebase.

imports/server/api.ts
```diff
@@ -1,9 +1,15 @@
...
+import hunts from "./api/resources/hunts";
+import tags from "./api/resources/tags";
 import users from "./api/resources/users";
 
 const api = express();
 api.use(authenticator);
 api.use("/users", users);
+api.use("/hunts", hunts);
+api.use("/tags", tags);
+api.use("/createPuzzle", createPuzzle);
```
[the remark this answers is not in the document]
jpd236: Sure, makes sense - done. To keep with the current convention, I moved the list tags and create puzzle implementations into hunts.ts. Let me know if you'd prefer I do something else.

jpd236:
Changes made, did a quick test w/ curl to make sure nothing broke with the reorganization. PTAL!

imports/server/api/resources/hunts.ts
```diff
@@ -0,0 +1,113 @@
...
+    check(userId, String);
+
+    const user = await MeteorUsers.findOneAsync({ _id: userId });
+    if (!user) {
+      // Should never happen if the API key passed authentication.
+      res.sendStatus(500);
+      return;
+    }
+
+    const allHunts = Hunts.find({}, { sort: { createdAt: -1 } });
```
ebroder: This is on me for not noticing it before, but any reason this can't just be `Hunts.find({ _id: { $in: user.hunts ?? [] }})`?

ebroder (commented):
Small nit, but otherwise LGTM

PR #2457, opened 2025-09-13
Bump slate, slate-history, slate-react

zarvox:
This revealed that our placeholder renderer was kinda going out of its way to do the wrong thing.  Upstream passes style props that in fact included opacity controls and `pointer-events: none`; we should really just tweak the `top:` value if we don't like the placement in the editor.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,6 +37,11 @@ updates:
       fontawesome:
         patterns:
           - "@fortawesome/*"
+      slate:
+        patterns:
+          - "slate"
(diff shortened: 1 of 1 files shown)


ebroder (changes requested):
This seems fine but if these need to be upgraded together, let's future-proof and add a new rule to: https://github.com/deathandmayhem/jolly-roger/blob/3e3bb093ecc5f549808cc92b347bcb8951ba82ac/.github/dependabot.yml#L30-L39

ebroder (approved):

Issue #2169, opened 2024-06-07
Infinite loop of MediaSoup errors 


jpd236:
My JR instance has been up for ~3 weeks.

Starting on 6/4 (two days after https://github.com/deathandmayhem/jolly-roger/issues/2167 happened, with nothing else in between in the server logs), the logs started showing a repeated sequence like this:

```
2024-06-04T02:40:07.127Z mediasoup:ERROR:Channel [pid:54 ChannelMessageRegistrator::RegisterHandler() | throwing MediaSoupError: Channel request handler with ID c93bdf42-c7e6-4f63-b288-f8d9546165d2 already exists
2024-06-04T02:40:07.170Z mediasoup:ERROR:Channel [pid:54 Worker::HandleRequest() | throwing MediaSoupError: Channel request handler with ID c93bdf42-c7e6-4f63-b288-f8d9546165d2 already exists [method:transport.produceData]
2024-06-04T02:40:07.178Z mediasoup:WARN:Channel request failed [method:transport.produceData, id:53]: Channel request handler with ID c93bdf42-c7e6-4f63-b288-f8d9546165d2 already exists [method:transport.produceData]
2024-06-04T02:40:51.254Z mediasoup:ERROR:Channel [pid:54 ChannelMessageRegistrator::RegisterHandler() | throwing MediaSoupError: Channel request handler with ID c93bdf42-c7e6-4f63-b288-f8d9546165d2 already exists
2024-06-04T02:40:51.254Z mediasoup:ERROR:Channel [pid:54 Worker::HandleRequest() | throwing MediaSoupError: Channel request handler with ID c93bdf42-c7e6-4f63-b288-f8d9546165d2 already exists [method:transport.produceData]
2024-06-04T02:40:51.255Z mediasoup:WARN:Channel request failed [method:transport.produceData, id:71]: Channel request handler with ID c93bdf42-c7e6-4f63-b288-f8d9546165d2 already exists [method:transport.produceData]
2024-06-04T02:41:20.346Z mediasoup:ERROR:Channel [pid:54 ChannelMessageRegistrator::RegisterHandler() | throwing MediaSoupError: Channel request handler with ID c93bdf42-c7e6-4f63-b288-f8d9546165d2 already exists
[...]

Issue #2167, opened 2024-06-03
Peer / channel request handler not found errors in audio call


jpd236:
A user was testing out audio calls on our instance of Jolly Roger; I'm not sure if they saw any visible symptoms of issues, but BugSnag did pick up a pair of errors at one point, both at the same time.

They were:

```
Error Peer not found [404] 
    imports/server/methods/mediasoupSetPeerState.ts:31:13 
    /built_app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40 
    /built_app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/promise_server.js:56:12 Function.Promise.await
    imports/server/methods/defineMethod.ts:65:9 
    /built_app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40 
```

and:

```
Error Channel request handler with ID 6e81e675-c152-480b-a100-e8a2ebb0aad0 not found [method:producer.resume] 
    /built_app/bundle/programs/server/npm/node_modules/mediasoup/node/lib/Channel.js:199:37 Channel.processMessage
    /built_app/bundle/programs/server/npm/node_modules/mediasoup/node/lib/Channel.js:69:34 Socket.<anonymous>
    events.js:400:28 Socket.emit
    domain.js:475:12 Socket.emit
    internal/streams/readable.js:293:12 addChunk
    internal/streams/readable.js:267:9 readableAddChunk
    internal/streams/readable.js:206:10 Socket.Readable.push
    internal/stream_base_commons.js:188:23 Pipe.onStreamRead
    /built_app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/promise_server.js:56:12 Function.Promise.await
    imports/server/mediasoup.ts:1175:7 
    /built_app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40 
```

Filing here for tracking since I didn't see anything with these error messages in a quick search.

PR #2470, opened 2025-09-13
Update mediasoup to 3.19.2

zarvox:
Split into a package update commit and a "the things I did to make the lints pass and the code run".

Tested only the `isDevelopment` local addressing codepath at present.  I don't have a particularly convenient way to test the "using a public IP" codepath short of "deploy it fully and try it out" (which I'm happy to do on merge), and I definitely have not tried the backend-to-backend forwarding, but ~none of the meaningful code there changed on our end as far as I know.

My intuition on protocol support is that we want to support both UDP and TCP for client-to-server connections (as UDP is generally better for low latency, but TCP is sometimes needed for TURN or other fallback paths), and that we would likely prefer UDP for server-to-server connections (since we control the networking on the backend), but I can adjust those if you prefer something else for any reason.

### DIFF
--- a/imports/server/mediasoup.ts
+++ b/imports/server/mediasoup.ts
@@ -5,8 +5,8 @@ import { networkInterfaces } from "os";
 import { Meteor } from "meteor/meteor";
 import { Random } from "meteor/random";
 import { Address6 } from "ip-address";
-import { createWorker, types } from "mediasoup";
-import { AudioLevelObserver } from "mediasoup/node/lib/AudioLevelObserver";
+import { createWorker, type types } from "mediasoup";
+import { AudioLevelObserverImpl } from "mediasoup/node/lib/AudioLevelObserver";
 import Flags from "../Flags";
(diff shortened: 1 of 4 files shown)


zarvox:
cc @jpd236 I have no particular sense for whether this will make a difference for your observed mediasoup stability issues, but figured I'd flag that we're taking the updates and it might be worth seeing if #2169 or #2167 recur still or not

imports/server/mediasoup.ts
```diff
@@ -705,7 +713,9 @@ class SFU {
       return;
     }
 
-    transport.observer.on(
+    const wtransport = transport as types.WebRtcTransport;
```
ebroder, line 716: Ugh. I'm annoyed that they made their code all Java so there's no way to do this with type narrowing

zarvox: Yeah I looked to see if they had type maps or anything hiding around but didn't find anything

PR #2482, opened 2025-09-18
Replace `react-copy-to-clipboard` with a modern implementation

### LINKED ISSUE #2429: Copy to clipboard should give visual indication that copy has completed
Some of the buttons don't even have a distinct state when pressed, so it's hard to tell if the copy actually happened.  It'd be nice if we:

1. consistently showed a tooltip atop the copy button indicating what it will do
2. upon successfully triggering a copy, updated the tooltip text to say "Copied!" until the tooltip is dismissed.

We currently have copy-to-clipboard in:

- [ ] ProfileList (for invite codes)
- [ ] NotificationCenter (for operators processing guess submissions)
- [ ] PuzzlePage (in the guess history/submission modal)
- [ ] GuessQueuePage (for copying guesses)
- [ ] Tag (for copying the entire table of puzzles in the group and their answers, if known)
- [ ] RTCDebugPage (for copying various ids)

zarvox:
`react-copy-to-clipboard` has a couple issues that make me want to replace it:

* it doesn't declare compatibility with React 19, so if we want to take the React 19 update we otherwise have to re-allow legacy peer deps
* the code to do so is merged upstream to master, but there's been no release in 3 years, and a related 9-month-old issue
* it doesn't have first-party typescript types
* also, it's ultimately just a thin wrapper around the `copy-to-clipboard` NPM package, which has its own cruft

When I look at the underlying copy-to-clipboard NPM package, I'm also not super enthusiastic about it: it does a whole bunch of complicated logic that affects the selection range, and then calls `document.execCommand("copy")` (which is deprecated per MDN, which recommends the use of the Clipboard API instead).  jolly-roger already only supports the evergreen browsers, so there's no great reason for us to use complicated logic for IE11 compatibility.

This PR implements a replacement copy-to-clipboard button component which uses `navigator.clipboard.writeText()` to interact with the clipboard, which is in the browser Baseline profile since 2020.

While I was here, I decided to also be consistent about including a tooltip and feedback on successful copy, which fixes #2429.


imports/client/components/GuessQueuePage.tsx
```diff
@@ -137,9 +137,9 @@ const StyledGuessConfidence = styled(GuessConfidence)`
   )}
 `;
 
-const StyledLinkButton: FC<ComponentPropsWithRef<typeof Button>> = styled(
-  Button,
-)`
+const StyledCopyToClipboardLinkButton: FC<
+  ComponentPropsWithRef<typeof CopyToClipboardButton>
+> = styled(CopyToClipboardButton)`
```
ebroder, line 142: OOC, is this type declaration actually necessary for anything?

ebroder (approved):

PR #2498, opened 2025-09-28
Refactor AccountForm pages and allow provisioning Google accounts from `/join/:invitationCode` pages

zarvox:
AccountForm was getting very complicated and it was getting hard to
imagine what any single page was going to look like, especially once we
wanted to make enrolling by invitation code able to provision new users.

This pulls apart the several pages that deal with login, password reset,
and onboarding so that each renders its own forms and tracks its own 
state.  A few shared components and hooks for form fields used on
multiple pages have been extracted into `AccountFormHelpers.tsx`.

To wit:

* `/login` now only shows the login form.
* `/forgot-password` is a new route used for requesting a password reset;
to enable smooth transitions between `/login` and `/forgot-password`,
both accept an initial email field value from `location.state`.
* `/reset-password/:token` is the route used to complete a password reset
* `/enroll/:token` is the route used to complete email/password account
registration.  Possession of the token is proof of ownership of the 
email address.
* `/join/:invitationCode` is discussed further below.

One invariant we preserve through all this is that we continue to not 
allow user accounts to be signed-in without having proven possession of
the email address in question (either by receiving an email or by having
Google tell us that they own that email).

We make /join able to support account creation, since it is a likely
first entrypoint for teams using shared hunt invite links rather than
individual email invites.  Authenticated users will be prompted to
explicitly accept the invitation by clicking a button, whereas
unauthenticated users will be able to either log in or provision a new 
account.

This gets into some complicated case logic:

* The user can either log in with Google or enter an email address.
* Logging in with Google is now able to provision a new account if
  logging in from the /join/:invitationCode page.  If the Google account is already associated with an existing account, we just sign that account in instead.  So Google login users don't have to know if they had an account already or not.
* The form also now intelligently handles whether an account matching a
  specified email exists or not, and what sign-in methods are available
  for the named user:
  * if a user enters an email address that is not associated with an
    account, we'll send them an account enrollment email to that
    address.
  * if they enter an address that is associated with an existing
    account, we'll prompt them to login with one of the currently-valid
    login methods (password or Google login) for that account.
  * if the email address has no valid login method (perhaps because it
    is a stub account that received an enroll email that expired), we'll
    detect this and nudge them to the password reset flow.

Google login is updated to populate the user's `displayName` from
Google's user profile `name` field on Google-driven account creation,
and all Google logins are updated to capture the `picture` field from
the login in case we later want to use those for avatars.

Co-authored-by: Jeff Davidson <jpd@google.com>


zarvox:
@jpd236 here's what I wound up with -- it sprawls a bit more, but hopefully each of the forms has fewer possible states.  Well, except `JoinHunt`, which is now very complicated as a result of handling the set of possible login methods and account provisioning methods, but at least the rest seem much more straightforward and should be trivial to modify without breaking other pages should we wish to tweak what we show on any of the pages.

zarvox:
Some screenshots to highlight the flows, since this is a big change.

## Login page

The login page is about the same, though I disabled the log in button until you have included at least one character in both email and password:

Screenshot 2025-09-28 at 10 04 24

Screenshot 2025-09-28 at 10 04 46

Screenshot 2025-09-28 at 10 05 32

## Forgot password page

Clicking the "Forgot your password?" link now takes you to a whole new route, but we preserve the contents of the `email` field across the navigation by stuffing it in `location`'s `state` field: 

Screenshot 2025-09-28 at 10 06 38

Screenshot 2025-09-28 at 10 07 03

## Reset password page

The `/reset-password/:token` page is broadly unmodified, but again the submit button is disabled until you provide at least one character in the password field:

Screenshot 2025-09-28 at 10 08 30

So that's login and password reset.  The Google sign-in button on the log-in page only allows signing in as users that already exist.

## Enroll page

The enroll page is also basically the same:

Screenshot 2025-09-28 at 10 11 06

Screenshot 2025-09-28 at 10 12 34

## JoinHunt page

`/join` is where things get fun.

### Logged in

If you're already signed-in, you'll basically get a "confirm your consent" page:

Screenshot 2025-09-28 at 10 15 25

Clicking the button joins you to the hunt, and redirects you to the hunt page.  If you're already in the hunt, then we just redirect you straight away to the hunt page; there's no action to be taken.

### Logged out

Now if you're logged out, instead you'll see this:

Screenshot 2025-09-28 at 10 17 03

Both sign-in mechanisms now handle both user creation or login.  An existing Google user clicking on the "Sign in with Google" button will yield:

 Screenshot 2025-09-28 at 10 18 54

The appearance to the end user is the same whether or not the User object existed already at the time of Google authentication -- they wind up logged in as a user with that `googleAccountId` (and related fields) on the user object.  We populate their `displayName` from Google's profile information included with the credential.  And then they get the logged-in view of this route, and pick up from there.

If they entered an email instead, we have a few possibilities:

#### no account

the email they provided is not associated with a user account yet.  We'll redeem the invitationCode for a /enroll/ link sent to the specified email address:

Screenshot 2025-09-28 at 10 30 00

Screenshot 2025-09-28 at 10 33 00

#### account with password set

You get a login form.

Screenshot 2025-09-28 at 10 37 20

Log in, and you get the "confirm your action" page:

Screenshot 2025-09-28 at 10 37 58

#### account with only Google login

That's right!  Now that we autoprovision accounts with Google, there's no guarantee that a user account will have a password set.  If such a user enters their email address, we should direct them to log in the way they will have the most success: with Google

Screenshot 2025-09-28 at 10 38 35

An edge case I don't care to make much better at the moment: if the `disable.google` feature flag is active, we handle this kinda badly:

Screenshot 2025-09-28 at 10 40 08

Theoretically that user could initiate a password reset to add a password to their account, but I think it's unlikely anyone runs with that feature flag flipped for a long period, so I find this less pressing.

#### account with password set and Google user associated

We encourage using the Google account in this case, since it's single-click.  But we can totally tweak the strings on the message here if we wanted.

Screenshot 2025-09-28 at 15 42 35

#### account, but with neither google nor password set

This can happen if the account previously received an enroll link, but the user never used it to finish account creation and login.  The recovery path requires doing a password reset.

Screenshot 2025-09-28 at 15 43 59

If you complete the password reset, you will have no `displayName` set, but the "no displayName detected, please update your profile" popup will appear and prompt the user to do that.  And then you might have to open the `/join` link again.  Maybe we should also attempt to redeem the invitation code to get the user account added to the hunt in this codepath too?

Anyway, thanks for looking through this novel.

imports/server/accounts.ts
```diff
@@ -55,32 +51,115 @@ const summaryFromLoginInfo = function (info: LoginInfo) {
...
+    }
+
+    // Google credentials were provided - obtain the email from the associated account,
+    // create a passwordless account with that email, and link it to the Google account.
+    // Also set the initial displayName for the user to their Google account's name.
+    const credential = await Google.retrieveCredential(
+      options.googleCredentials.key,
+      options.googleCredentials.secret,
+    );
+    const { email, id, name, picture } = credential.serviceData;
```
jpd236, line 101: Do we actually use this picture anywhere, or is it just being stored for potential future usage? I'm not seeing any actual usage of it, at a glance.

I'm a little uncertain how this interacts with the profile photo that can be inferred from the Discord connection (which was always a bit funny to me in that it can't be updated in JR, only Discord).

imports/server/accounts.ts
```diff
@@ -55,32 +51,115 @@ const summaryFromLoginInfo = function (info: LoginInfo) {
...
+        // Autoprovision a new user
+        const userId = await Accounts.createUserAsync({
+          email,
+        });
+        await MeteorUsers.updateAsync(userId, {
+          $set: {
+            googleAccount: email,
+            googleAccountId: id,
+            googleProfilePicture: picture,
+            displayName: name,
```
jpd236, line 117: Is this guaranteed to be populated?

I'm guessing not, and that if it isn't, we have the fallback of the site notifying you to set one while still not having things fall apart in the meantime, but just want to make sure it'd be handled reasonably gracefully.

imports/lib/models/User.ts
```diff
@@ -45,6 +45,7 @@ export const User = z.object({
   displayName: nonEmptyString.optional(),
   googleAccount: nonEmptyString.optional(),
   googleAccountId: nonEmptyString.optional(),
+  googleProfilePicture: nonEmptyString.optional(),
```
jpd236, line 48: should we document anywhere that this won't be populated for existing linked accounts? I assume we don't particularly care about backfilling (e.g. on the next sign-in for that user?)

imports/server/accounts.ts
zarvox: Not actually used anywhere yet.

I'm interested in improving profile picture options for users, though one reason we haven't really done much on that front is that I'm not super excited about trying to add image manipulation code (complicated UI, potentially large file manipulation, resizing) & dependencies, nor dealing with storage for profile images -- right now the only mutable storage we reliably have is mongodb, and I'm not sure I want to be regularly serving large blobs out of the database.  Although I guess we already do this for whitelabeling assets?  Probably not the end of the world, but worth thinking about how we'd want to handle it, and how we'd want to support caching so we're not blowing all our database bandwidth on profile pictures all the time.

Anyway, if we could piggyback on Google profiles, or at least offer the option of using that image (vs unilaterally saying "it'll be your Discord profile pic", that might be an easier incremental win.  It's possible we won't ever use this, but I figured it was worth capturing while I was here thinking about it.

imports/client/components/ForgotPasswordForm.tsx
```diff
@@ -0,0 +1,90 @@
...
+          type="submit"
+          variant={submitDisabled ? "outline-primary" : "primary"}
+          disabled={submitDisabled}
+        >
+          Email Reset Link
+        </Button>
+      </div>
+      <StyledModeSwitchLink>
+        <p>
+          If you already have an account,{" "}
```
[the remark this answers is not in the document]
zarvox: Fixed in https://github.com/deathandmayhem/jolly-roger/pull/2498/commits/9768fb50613026f98c324f6834b03c8f2c1ce7ca

imports/client/components/AccountFormHelpers.tsx
```diff
@@ -0,0 +1,260 @@
...
+        token,
+        secret,
+      });
+      onCredentialsSet?.({ token, secret });
+    },
+    [onCredentialsSet],
+  );
+
+  const onClick = useCallback(() => {
+    setSubmitState(AccountFormSubmitState.SUBMITTING);
```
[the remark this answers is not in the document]
zarvox: Fixed in https://github.com/deathandmayhem/jolly-roger/pull/2498/commits/5a8f42644650515c2f69d00e9959c40d51eea584

imports/client/components/JoinHunt.tsx
```diff
@@ -1,24 +1,450 @@
...
+// If the user is logged in and not yet part of the hunt, we'll present them
+// with a big clickable button to join the hunt.
+// If they're already in the hunt, we'll just redirect them to the hunt page.
 const JoinHunt = () => {
+  const [authLoading, loggedIn] = useAuthenticated();
+  const teamNameLoadingFunc = useSubscribe("teamName");
+  const teamNameLoading = teamNameLoadingFunc();
+  const teamName = useTracker(() => {
+    if (teamNameLoading) return "(loading team name)";
+    return TeamName.findOne("teamName")?.name ?? "Default Team Name";
```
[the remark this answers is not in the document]
zarvox: Okay, I talked myself into fixing this after all 😅  https://github.com/deathandmayhem/jolly-roger/pull/2498/commits/aab7da81351b4bcff6163b0112e57b0dedaf75ed

PR #2556, opened 2025-12-05
Bump gaxios and Google APIs together

ebroder:
Supersedes #2530 and #2458


.github/dependabot.yml
```diff
@@ -32,6 +32,7 @@ updates:
       google:
         patterns:
           - "@googleapis/*"
+          - gaxios
```
zarvox: Thanks for adding this so it'll get rolled in together in the future

zarvox (approved):

PR #2565, opened 2025-12-05
Support uploading or pasting images in chat

welkin25:
Allows user to upload image or direct paste in (e.g. screenshot) into the chat box. The image gets uploaded to s3 and then rendered in chat. If image is larger than the chat bar it'll be a link which opens the image in a new tab.


imports/methods/createChatImageUpload.ts
```diff
@@ -0,0 +1,7 @@
+import TypedMethod from "./TypedMethod";
+
+export default new TypedMethod<
+  { puzzleId: string; mimeType: string },
+  | { publicUrl: string; uploadUrl: string; fields: Record<string, string> }
+  | undefined
```
ebroder: This disjunction doesn't actually match the type that the method is checking for, right? Should just be the first option?

imports/server/hooks/DiscordHooks.ts
```diff
@@ -37,6 +38,9 @@ async function renderChatMessageContent(
 ): Promise<string> {
   const chunks = await Promise.all(
     content.children.map(async (child) => {
+      if (nodeIsImage(child)) {
+        return "[image]";
```
ebroder: Any reason not to include the URL here? Discord will probably do a reasonable job of unfurling it

ebroder (commented):
First off, thanks for sending this in :) 

I've done a first pass through this and left some comments. The overall approach seems plausible to me, although I still need to wire up the uploads and poke at the FancyEditor changes a bit. (Apparently my local dev environment does not have S3 configured)

Error handling is definitely IMO the most important thing to refine here; a lot of the rest I still want to see fixed but is more nitpicky and local than sweeping and invasive.

imports/client/components/ChatMessage.tsx
```diff
@@ -38,6 +40,70 @@ const StyledCodeBlock = styled.code`
...
+    }
+
+    return () => {
+      if (observer) observer.disconnect();
+    };
+  }, [handleLoad]);
+
+  return (
+    <div ref={containerRef} style={{ width: "100%" }}>
+      {isLarge ? (
```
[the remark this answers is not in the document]
welkin25: I added the observer so that the image will be rendered differently, i.e. it will add a link and use thumbnail style around the image when the width is greater than the container width, to make it easier to know whether you need to click on it.

imports/server/hooks/DiscordHooks.ts
welkin25: ah you're right, will fix.

imports/client/components/PuzzlePage.tsx
```diff
@@ -686,15 +711,77 @@ const ChatInput = React.memo(
...
+              })
+                .then(() => {
+                  fancyEditorRef.current?.replaceImage(publicUrl, tempId);
+                })
+                .catch((uploadErr) => {
+                  throw new Error("S3 upload failed", uploadErr);
+                });
+            }
+          },
+        );
```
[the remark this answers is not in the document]
welkin25: thanks! I've pushed a couple new commits with the fixes; currently it displays a broken image icon along with an error modal. I debated about deleting the broken image node automatically but thought maybe it's better to leave it in place as a reminder that the image upload failed.

imports/client/components/FancyEditor.tsx
```diff
@@ -130,17 +144,65 @@ const EditableMentionRenderer = ({
...
+        return markableVoid(element);
+    }
+  };
+  editor.insertData = (data: DataTransfer) => {
+    if (!data) return;
+    for (const file of Array.from(data.files)) {
+      if (file.type.startsWith("image/")) {
+        uploadImageFile(file);
+        return;
+      }
```
ebroder: I think this has some slightly weird interactions with the fact that `createChatImageUpload` doesn't do any validation of the submitted mime type.

Setting `accept="image/*"` on the file input means that my browser defaults to image-only uploads, but I can override that to upload any file type. If I do that and upload, say, a `text/plain` file, it gets uploaded to S3 and then disappears - I think because this check short-circuits.

My instinct is that we should check and refuse to upload files that aren't mime type `image/*` - probably by checking both in `createChatImageUpload` and `uploadImageFile`

welkin25:
Thanks, added type validation and changed the images to fontAwesome icons!

ebroder (approved):
Great, thanks for bearing with me. This looks good to me.

PR #2571, opened 2025-12-06
Bump eslint dependency chain

ebroder:
Primarily a bump to the various typescript-eslint packages. As part of that, typescript-eslint dropped support for declaring `tsconfigRootDir` as "." in the eslint config, which means that we need an eslintrc.js instead of a json file. Prettier ensures that it's formatted slightly differently, but the content is mostly unchanged.

eslint-plugin-deprecation got rolled into typescript-eslint, so that's dropped and replaced with the appropriate new overrides.

Most of the code changes are related to typescript-eslint being pickier about enum type compatibility — it's now an error to compare an enum value with a string or number directly, so those comparisons have been tweaked.


ebroder:
(This doesn't bump eslint-plugin-react-hooks because eslint-config-airbnb has a peer dependency on a specific version, which is pretty annoying. We'll likely end up having to dump eslint-config-airbnb anyway, either because we upgrade to eslint v9 and flat config, which it doesn't support, or because we switch to biome. Not a yak I wanted to shave right this instant)

PR #2576, opened 2025-12-09
Prep for React 19 upgrade

ebroder:
React 19 changed how refs behave. These changes make our code compatible with both the behavior in v18 and v19.

(I think we should theoretically be able to auto-upgrade to React 19 once this + #2575 + https://github.com/ianstormtaylor/slate/pull/5986 land, although it's separately worth taking advantage of some of the new features in v19, like ref-as-prop, the new `useActionState` hook, and potentially the new `use` API for suspending while promises are resolving)


zarvox (approved):
Looks like just typing changes, that allow being more precise/narrow with the actual types of the `RefObject`s?  Sounds great.

PR #2587, opened 2025-12-10
Rework Docker and CI builds

ebroder:
This originally started with the goal of creating multi-architecture Docker images that included support for arm64, but in the process of doing that, this basically ended up rototilling our build pipeline.

The biggest end-user visible change is that tests now run under playwright instead of puppeteer. Playwright seems to generally have broader usage and better tooling for managing browsers and their dependencies. (Our old approach of installing Chrome's dependencies was messy and non-portable). The Meteor testing framework has built-in support for both, and other than swapping the packages and an environment variable, no other changes were required.

With that change, our builds run under arm64. But in order to generate a single Docker image that runs under both amd64 and arm64 (without relying on QEMU), we need to rework the GitHub Actions workflow to build each architecture in parallel and then combine them in a separate step.

Along the way, it also made sense to split out linting and testing into a separate parallel step to the production Docker image build, rather than running them sequentially.

And finally, while testing and validating the new build changes, I found several places where our Docker images, layers, and caching were operating less efficiently than they could be; this change attempts to fix those.

A few specific tweaks:

* Meteor seems very eager to regenerate its local sqlite package database cache on every run, even of `meteor npm` and `meteor npx`. Setting `METEOR_OFFLINE_CATALOG=1` doesn't completely eliminate it, but seems to tame it somewhat.
* It also likes to re-install NPM dependencies of Meteor packages very
  frequently, which results in dependencies like SWC showing up in
  multiple different image layers. Since we almost never change Meteor
  package versions outside of Meteor upgrades, merge the Meteor install
  and the Meteor package fetch into a single step.
* Meteor also really likes to install separate copies of NPM dependencies in each Meteor package with that dependency, even if they depend on the same version. This resulted in (e.g.) several copies of SWC. By using hardlink deduplication, we can significantly shrink the image size.
* Uploading the Docker image caches seems to be particularly expensive, so for the lint/test step (which already has to do more work than the build), skip pushing the cache back to GitHub (and instead rely on the cache from the build step).
* And finally, improve the cache hit rate by not including the full git repository in the production Docker build, since that results in unnecessary churn in the image layers.

All of this together means that, even though we're adding new functionality in the form of the arm64 build, the overall build time improves. In testing, a build on a completely cold cache seems to take about 8 minutes, while a perfect cache (i.e. an immediate rebuild) takes about 4.5.


ebroder:
Alright, I think this is finally ready for review.

For reference, here's a build from a completely cold cache (I manually purged it before starting the run): https://github.com/deathandmayhem/jolly-roger/actions/runs/20121760479/attempts/3, and a build with a fully populated cache: https://github.com/deathandmayhem/jolly-roger/actions/runs/20121760479/attempts/5

There's obviously some variability in build times, but overall I think this comes out to a net win. Plus now I can go experiment with Graviton instances which I've always wanted to do

ebroder:
(Oh, because the step names changed, we'll need to edit the required status check configs, but that's easy enough)

PR #2589, opened 2025-12-12
Apply lint TODOs

ebroder:
Go through the biome rules that were tagged as worth revisiting and revisit them. Many of these can be easily codemod'd (d1b847e553eb5d11a573131bad9bfd701c0a4d46). The rest are each addressed in separate commits.

As a note, 64dc7828eafd324f84f1d757a6cf31c4ccbbfc31 is reasonably invasive and has some implied philosophical choices about how we should be using id attributes in our code. I have tested the flows that were edited and confirmed that they still work.

None of this should result in any behavioral change.

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -85,7 +85,7 @@
       "suspicious": {
         // This is a good idea, but isn't type aware so complains even if the lambda returns void
         "useIterableCallbackReturn": "off",
-        // TODO: consider migrating this on, but right now our code isn't compliant
+        // This rule seems flawed - TypeScript is perfectly capable of inferring types in let declarations
         "noImplicitAnyLet": "off",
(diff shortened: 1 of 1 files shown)


zarvox:
Okay yeah the Enable noAccumulatingSpread patch added the error that the unit test is now hitting.

Patchset looks broadly good, modulo you'll either want to support multiple regexes in your updated `generateJsonSchema` implementation in c1201ad7cdae20a21b8b5e03f7828d768f299cf7 or remove/update the failing unit test to match the new behavior.

ebroder:
Yeah, I was being a little lazy. Should be fixed now (although I had to munge some `readonly` prop declarations around to make it work)

zarvox (approved):

PR #2606, opened 2025-12-13
Workers should clean themselves up if another server thinks they died

ebroder:
We've seen weird cases where a process's database records get garbage collected while they are still running (likely due to scheduling weirdness or temporary hangs). This causes problems as there can still be in-memory resources (e.g. Mediasoup transports) that don't get cleaned up and recreated when this happens. So if a server sees that it has been garbage collected by another process, it should exit to try and restore normalcy.

As a guard against aggressive cascading, only do this if the process is part of a worker pool, and therefore will be replaced by the pool manager.

I'm hopeful this will improve #2169, if not resolve it.

### DIFF
--- a/imports/server/garbage-collection.ts
+++ b/imports/server/garbage-collection.ts
@@ -7,38 +7,38 @@ import { Meteor } from "meteor/meteor";
 import { Random } from "meteor/random";
 import Logger from "../Logger";
 import Servers from "../lib/models/Servers";
+import ignoringDuplicateKeyErrors from "./ignoringDuplicateKeyErrors";
+import onExit from "./onExit";
 
 const serverId = Random.id();
 
 // Global registry of callbacks to run when we determine that a backend is dead.
-const globalGCHooks: ((deadServers: string[]) => void | Promise<void>)[] = [];
+const globalGCHooks: ((deadServers: string) => void | Promise<void>)[] = [];
 
 function registerPeriodicCleanupHook(
-  f: (deadServers: string[]) => void | Promise<void>,
+  f: (deadServer: string) => void | Promise<void>,
 ): void {
   globalGCHooks.push(f);
 }
 
-let firstUpsert = true;
 async function cleanup() {
-  const result = await Servers.upsertAsync(
-    { _id: serverId },
+  const updated = await Servers.updateAsync(
+    {
+      _id: serverId,
+      cleanupInProgressBy: { $exists: false },
+    },
     {
       $set: {
-        pid: process.pid,
-        hostname: os.hostname(),
+        updatedAt: new Date(),
       },
     },
   );
-  if (!firstUpsert && result.insertedId) {
-    Logger.warn("Server record unexpectedly deleted", {
-      serverId,
-      pid: process.pid,
-      hostname: os.hostname(),
-      error: new Error("Server record unexpectedly deleted"),
-    });
+  if (updated === 0) {
+    // Someone else is cleaning us up. The observer below should notice this and
+    // exit, but we should go ahead and short-circuit to avoid doing any more
+    // work
+    return;
   }
-  firstUpsert = false;
 
   // Servers disappearing should be a fairly rare occurrence, so it's
   // OK for the timeouts here to be generous. Servers get 120 seconds
@@ -48,6 +48,7 @@ async function cleanup() {
   const deadServers = await Servers.find({
     updatedAt: { $lt: timeout },
   }).mapAsync((server) => server._id);
+
   if (deadServers.length === 0) {
     return;
   }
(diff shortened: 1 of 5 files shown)


zarvox (approved):
This probably helps, and I'm happy to merge it, but it still feels a little looser with regards to the possible implied state machine than I'd like.

I think I'd feel a little more comfortable if we tweaked the model a bit more, perhaps like:

1. A worker process always does an explicit `insert()` for the first insertion, and notes the created `_id`.
2. A worker process always does an explicit `update()` for subsequent refreshes, with a selector including their `_id`, allowing them to detect (through 1 vs 0 documents modified) the deletion of their own records **without** re-creating the Servers record.
3. (optional) When checking other Server records for staleness, if a server record is observed to be stale, before cleaning up records bound to the stale Server, a garbage-collector process tries to set fields `cleanupInProgressBy` and `cleanupInProgressAt` on that Server record (with a selector including both `_id` and `{ cleanupInProgressBy: { $exists: false } }`) before proceeding to delete functional records.  If we do this, then the check in 2. can also include `cleanupInProgressBy: { $exists: false }` in its update() selector, which will let that process know to abort before we remove other records out from under it, since [we delete those records first and only at the end do we delete the Server record](https://github.com/deathandmayhem/jolly-roger/blob/2e7697d3db1413973e94447daea5a44f72bc719e/imports/server/garbage-collection.ts#L59-L65). 

So then the flow becomes something like:

1. worker A creates Server record for self (init -> live)
2. worker A updates Server record `updatedAt` periodically (live -> live)
3. if worker A fails to timely update `updatedAt`, worker A is considered stale (live -> stale)
4. if worker B sees stale Server record for A, worker B sets server A's cleanupInProgressBy to B (signalling start of GC action) (stale -> cleanup)
5. after B marks A for cleanup, B can steal A's work (or just delete A's records, as appropriate), then once done, B deletes A's record (cleanup -> dead)
6. if worker A sees its own Server record has been deleted **or** has had `cleanupInProgressBy` set, worker A immediately aborts

My hope with such an approach is to allow a somewhat orderly cleanup, to flag to the server being collected that this is happening at the start rather than the end of the action, and to limit the number of competing servers trying to clean up a stale Server (to try to reduce risk/frequency of such races).

One additional thing we might consider is using `observeChanges()` to detect the stale -> cleanup transition more immediately.

Thoughts?

ebroder:
Yeah I'm OK with that approach. I attempted to do something that matched that rough sketch (although the details look a little different). PTAL?

imports/server/garbage-collection.ts
```diff
@@ -48,43 +48,102 @@ async function cleanup() {
...
+      });
+    });
+
     periodic();
-  }),
-);
+
+    onExit(async () => {
+      handle?.stop();
+      await Servers.removeAsync(serverId);
```
zarvox: You probably **don't** want to proactively remove the record from `Servers` unless you're also in the same breath going to remove all the other e.g. mediasoup records that have that `serverId` -- GC cleaning up the other references to that `serverId` is how we avoid leaking those records forever.

imports/server/garbage-collection.ts
```diff
@@ -7,38 +7,38 @@ import { Meteor } from "meteor/meteor";
...
 }
 
-let firstUpsert = true;
 async function cleanup() {
-  const result = await Servers.upsertAsync(
-    { _id: serverId },
+  const updated = await Servers.updateAsync(
+    {
+      _id: serverId,
+      cleanupInProgressBy: undefined,
```
zarvox: I always forget: does this do the right thing, or does `undefined` get dropped from the selector entirely we need this selector to be `$exists: false`?

zarvox (commented):
Seems reasonable on the whole; two things worth double-checking

imports/server/garbage-collection.ts
ebroder: Good point. Historically I think we've just always leaked these records, but in actuality I think we can just trigger garbage collection on ourselves at this point? Pushed a mod to do that.

imports/server/garbage-collection.ts
```diff
@@ -48,43 +48,103 @@ async function cleanup() {
...
+    });
+
     periodic();
-  }),
-);
+
+    onExit(async () => {
+      handle?.stop();
+      // Attempt to clean our own server record up synchronously on exit
+      await cleanupDeadServer(serverId);
```
zarvox: I think eager cleanup is still messy due to https://github.com/deathandmayhem/jolly-roger/blob/b95d4d36e3bd9a67bd598d65b9738f63d6afc3c4/imports/server/mediasoup-api.ts#L28-L55 trying to take over for the dead server.

imports/server/garbage-collection.ts
zarvox: Unfortunately I think this doesn't do what you think it does, at least in JS:

```
> await Models.Servers.find({ _id: "KH3KXvw4A98hQ74Ea", cleanupInProgressBy: undefined }).fetch()
[
  {
    _id: 'KH3KXvw4A98hQ74Ea',
    hostname: 'Mac.attlocal.net',
    pid: 51912,
    updatedAt: 2025-12-23T20:38:45.227Z,
    cleanupInProgressBy: 'KH3KXvw4A98hQ74Eb'
  }
]
```

I think the `undefined` gets dropped from the selector on the JS side before it gets to Mongo; you may be able to achieve what you were expecting with `null` but `undefined` here definitely is not excluding servers that have a `cleanupInProgressBy` set already.

zarvox (changes requested):
Correct me if I'm wrong, but there appear to be multiple correctness bugs still here

imports/server/garbage-collection.ts
```diff
@@ -48,43 +48,98 @@ async function cleanup() {
...
+        },
+        removed() {
+          aborted();
+        },
+      });
+
+      onExit(() => handle.stop());
+    }
+
+    await ignoringDuplicateKeyErrors(async () => {
```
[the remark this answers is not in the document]
zarvox: Ah, that makes sense -- might be worth a comment to that effect.

PR #2612, opened 2025-12-16
Support @-mentioning operators in chat messages

ebroder:
Although I still quite like the idea, I wanted to try a slightly different approach to @welkin25's #2608 PR. This has two major changes by comparison:

- Instead of treating "special users" as a new type of standard mention (with a hard-coded but non-conflicting fake user ID), create a new type of "role" mention. This feels (IMO) analogous to, e.g., Discord, which distinguishes between mentioning users with `<@snowflake>` and roles with `<&@snowflake>`, even though the UI shows both of them together.
- Avoid sending notifications to all operators, instead only sending the notification to operators that we believe are currently actively operating (based on whether they are online and have the operator interface enabled). Of course, since there's some possibility that *no* operators are online/operating, include a fallback to send the notification to all operators.

From the commit message (although it basically says the same thing in different words):

> This introduces the concept of a "role mention" as a separate construct from a user "mention", although for the moment the only role that can be mentioned is "operator".
> 
> To limit the blast radius of such mentions, @operator mentions only notify operators that are currently online and with operating mode enabled. This is detected by introducing a new usage of the subscribers counters - operators with operating mode enabled for any hunt will increment a "operators" subscriber counter with the hunts for which they're actively operating in the context, allowing us to track them down from the server.
> 
> If no operators are online with operating mode enabled, the mention goes out to all operators for the hunt, active or not.

### DIFF
--- a/imports/client/components/FancyEditor.tsx
+++ b/imports/client/components/FancyEditor.tsx
@@ -49,11 +49,18 @@ export const DEBUG_EDITOR = false;
 export type CustomText = { text: string };
 export type MessageElement = {
   type: "message";
-  children: (MentionElement | ImageElement | CustomText)[];
+  children: (MentionElement | RoleMentionElement | ImageElement | CustomText)[];
 };
 export type MentionElement = {
   type: "mention";
   userId: string; // user._id of the mentioned user
+  children: CustomText[];
+};
+const AllRoles = ["operator"] as const;
+type AllRolesType = (typeof AllRoles)[number];
+export type RoleMentionElement = {
+  type: "role-mention";
+  roleId: AllRolesType; // name of the mentioned role
   children: CustomText[];
 };
 export type ImageStatus = "success" | "error" | "loading";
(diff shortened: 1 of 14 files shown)


imports/client/components/FancyEditor.tsx
```diff
@@ -147,12 +159,23 @@ const EditableMentionRenderer = ({
...
+  switch (element.type) {
+    case "mention":
+      name = users.get(element.userId)?.displayName ?? element.userId;
+      break;
+    case "role-mention":
+      name = element.roleId;
+      break;
+    default:
+      // biome-ignore lint/nursery/noUnusedExpressions: exhaustive check
+      element satisfies never;
```
zarvox: ooooh nice exhaustiveness checking

ebroder:
@welkin25 For the record, I can't actually tag you as a reviewer, but I'm curious about your take as well.

ebroder:
Thanks for the review, should all be fixed. PTAL?

zarvox (approved):

welkin25:
> @welkin25 For the record, I can't actually tag you as a reviewer, but I'm curious about your take as well.

This looks great! Using roles seems a cleaner fix, and I like the added active operators check. Thanks for adding this!

PR #2616, opened 2025-12-19
Catch all voided promises on server

zarvox:
Since node 15, unhandled promise rejections have changed behavior: whereas they used to log an error to the console, now they cause the process to exit.

This is good for helping track down subtle incorrectness caused by rejecting promises, but we have a bunch of cases where we were `void`-ing promises when we couldn't do anything particularly useful if they were to fail.  While we don't see these promises reject often, we definitely don't want a situation where e.g. a mongo write for one workflow fails which proceeds to take down the entire backend.

To that end, we ensure that all promises that we had simply been ignoring that could reject are now caught.

### DIFF
--- a/imports/server/googleClientRefresher.ts
+++ b/imports/server/googleClientRefresher.ts
@@ -1,3 +1,4 @@
+import { Meteor } from "meteor/meteor";
 import type { Mongo } from "meteor/mongo";
 import { OAuth } from "meteor/oauth";
 import type { Configuration } from "meteor/service-configuration";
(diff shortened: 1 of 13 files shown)


imports/server/onExit.ts
```diff
@@ -2,12 +2,15 @@ const exitHandlers: (() => void | Promise<void>)[] = [];
...
-    void (async () => {
+    (async () => {
       for (const handler of exitHandlers.splice(0)) {
         await handler();
       }
       process.kill(process.pid, signal);
-    })();
+    })().catch((error) => {
+      // biome-ignore lint/suspicious/noConsole: would prefer Logger but want to minimize dependencies in signal handler
+      console.log(`error in signal ${signal} handler`, error);
```
ebroder: If we catch an error here we probably want to still re-send the signal

imports/server/googleClientRefresher.ts
```diff
@@ -51,16 +52,24 @@ class GoogleClientRefresher {
...
+      });
+    this.oauthCredentialCursor
+      .observeAsync({
+        added: (doc) => this.updateOauthCredentials(doc),
+        changed: (doc) => this.updateOauthCredentials(doc),
+        removed: () => this.clearOauthCredentials(),
+      })
+      .catch((error) => {
+        Logger.error("Failed observing oauth credential cursor", { error });
+      });
```
ebroder: Given that `GoogleClientRefresher` is a singleton and this constructor only runs at startup, what do you think about instead moving this to an `async init()` similar to the Discord client refresher (or a `static async create` a la the mediasoup SFU)?

imports/server/setup.ts
```diff
@@ -120,23 +121,27 @@ Meteor.publish("googleScriptInfo", async function () {
...
         tracked = true;
         this.added(
           "googleScriptInfo",
           "googleScriptInfo",
           await formatDoc(doc),
         );
-      })();
+      })().catch((error) => {
+        Logger.error("googleScriptInfo added() failed", { error });
+      });
```
ebroder: (Similar question here about whether we should `this.error` and unwind the subscription on error)

imports/server/users.ts
```diff
@@ -77,7 +78,12 @@ const republishOnUserChange = async (
           merger.removeSub(currentSub);
         }
         currentSub = newSub;
-      })();
+      })().catch((error) => {
+        Logger.error("Error republishing on user change", {
+          error,
+          user: sub.userId,
+        });
+      });
```
ebroder: (Also here re: `this.error`)

ebroder (commented):
I'm definitely not wildly enamored with this, TBH, mostly because I don't love the impact on legibility (especially e.g. in the Mediasoup code). But fine, it seems like an improvement.

There are several cases where I think we can do better than logging the rejection and marching onwards; where it's sensible, I think I'd prefer that over the alternative.

PR #2638, opened 2025-12-22
Miscellaneous cleanups

ebroder:
I've been intending to go around and clean most of these up for a while.

### DIFF
--- a/imports/server/addRuntimeConfig.ts
+++ b/imports/server/addRuntimeConfig.ts
@@ -1,5 +1,8 @@
-// Work around a bug in Meteor's runtime config hooks (which will only run a
-// single hook) by accumulating our own list
+// Our usage of Meteor's runtime config hooks is only to augment additional
+// fields, so add a wrapper which makes doing that a little more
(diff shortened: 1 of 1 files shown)


imports/server/addPuzzle.ts
```diff
@@ -149,11 +149,9 @@ export default async function addPuzzle({
 
   // Run any puzzle-creation hooks, like creating a default document
   // attachment or announcing the puzzle to Slack.
-  Meteor.defer(
-    Meteor.bindEnvironment(() => {
-      void GlobalHooks.runPuzzleCreatedHooks(puzzleId);
-    }),
-  );
+  Meteor.defer(() => {
+    void GlobalHooks.runPuzzleCreatedHooks(puzzleId);
```
zarvox: Huh, did I miss this `void` without a catch or a comment?

ebroder: Yeah I guess so, although the hook runners catch and log exceptions so I think it's fairly safe

PR #2639, opened 2025-12-23
Add some rules for import consistency

ebroder:
Meteor currently is not capable of tree-shaking (and even though it may get support in 3.4 with rspack, it seems like react-bootstrap may not support it anyway). By forcing some consistency in how we import from otherwise large packages, we can make sure we don't accidentally bloat our bundle size.


biome.jsonc
```diff
@@ -80,7 +80,23 @@
...
+        "noRestrictedImports": {
+          "level": "warn",
+          "options": {
+            "paths": {
+              "react-bootstrap": "Please import from 'react-bootstrap/Component' instead to reduce bundle size.",
+              "@fortawesome/free-solid-svg-icons": "Please import from '@fortawesome/free-solid-svg-icons/IconName' instead to reduce bundle size."
+            },
+            "patterns": [
+              {
+                "group": "react-bootstrap/esm/*",
```
ebroder: (FTR, vscode likes to auto-import from these esm/ paths, which is why we ended up with some of them)

Issue #549, opened 2022-01-10
Feedback on Invite Page


flipdog:
Currently a lot of our on-boarding workflow involves copy-pasting email addresses to the invite or "bulk invite" box, and then hoping the rest of the process has moved smoothly. In particular, there's no indicator of state change on submission, so if the client-side code just fails to fire (or my internet connection is broken) there's no way for me to know. It would be helpful to add some feedback for the person doing that inviting. None of this is critical, and the number of users affected by it is low.

Some useful components would be:
- success/failure of invite process
- success/bounce of invite email
- info about user being invited

For example, for a single invite I could see this being implemented as a tooltip/popover/etc. that says one of the following:
- New user $EmailAddress invited to Jolly Roger and $HuntName Hunt
- Existing user $EmailAddress ($FirstName $LastName) added to $HuntName Hunt
- Failed to add user $EmailAddress. $ErrorReason

For the bulk add functionality, perhaps a popover that just says "$X new hunters invited and added to $HuntName. $Y existing hunters added to $HuntName. $Z failures: $ListofFailedAddressesAndReasons"

I'm also aware that email bounces are a pain in the butt and don't necessarily process right away, so it's possible that this feedback loop is best delayed and done via email?

PR #2647, opened 2025-12-26
Update bulk add hunt for better error reporting

ebroder:
Two changes:

- One, we previously attempted to add all users in parallel, which sounds like a great way to get rate limited by everything everywhere. So instead let's do it in serial.
- Two, we attempted to sanitize errors that we sent back to the client, but in practice the error "reason" was always unset, so just send the whole stringified error.

(This is a "let's make things strictly less worse" patch, not a "let's make things good" patch)

Related to #549 but goodness gracious does not come anywhere close to fixing it.


zarvox (approved):
I agree this is very likely no worse. 🙂

PR #2649, opened 2025-12-30
Implements dark mode

### LINKED ISSUE #1821: Dark mode
Bootstrap now [supports color modes](https://getbootstrap.com/docs/5.3/customize/color-modes/), which can be used to toggle between a light and dark mode. We can either have a UI toggle element or go based on the system mode setting (by setting `$color-mode-type: media-query`).

Sadly, it's not quite as easy as just turning this on, as the default Bootstrap dark mode theme is not especially compatible with our color choices, so we likely need to do an audit for everywhere that we're specifying colors and make sure they work across modes. [This search](https://github.com/search?q=repo%3Adeathandmayhem%2Fjolly-roger+%28%2F%23%5B0-9a-f%5D%7B3%7D%2F+OR+%2Frgba%3F%5C%28%29&type=code) is probably a good place to start

jimsug:
This introduces a theming system to jolly-roger, adding support for **dark mode** in addition to the current (default) light mode:

- The themes are parameterised in `imports/client/theme.ts`, with a (large!) number of definitions for various elements; we may wish to rationalise them at a future time, but I don't think that's necessary at the moment
- We use localStorage (via `imports/client/hooks/persisted-state.ts`) to persist user theme selections (`useAppThemeState`)
- The selector is added to the user menu/profile dropdown in `imports/client/components/App.tsx`, and we add styling to the app's container element to cover many cases by default
- Some components have custom styling, which have been updated in the remaining changed files.
- We additionally make an `auto` mode available, which will select either light or dark depending on the OS preference as detected by media query.

This would close #1821.

### DIFF
--- a/imports/client/components/App.tsx
+++ b/imports/client/components/App.tsx
@@ -2,10 +2,10 @@ import { Meteor } from "meteor/meteor";
 import { useTracker } from "meteor/react-meteor-data";
 import Bugsnag from "@bugsnag/js";
 import { faExclamationTriangle } from "@fortawesome/free-solid-svg-icons/faExclamationTriangle";
-import { faMagic } from "@fortawesome/free-solid-svg-icons/faMagic";
 import { faMoon } from "@fortawesome/free-solid-svg-icons/faMoon";
 import { faSun } from "@fortawesome/free-solid-svg-icons/faSun";
 import { faUser } from "@fortawesome/free-solid-svg-icons/faUser";
+import { faWandMagicSparkles } from "@fortawesome/free-solid-svg-icons/faWandMagicSparkles";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import React, { useCallback, useEffect, useMemo, useState } from "react";
 import Alert from "react-bootstrap/Alert";
@@ -305,7 +305,7 @@ const AppNavbar = ({
                 onClick={() => setAppTheme("auto")}
                 active={appTheme === "auto"}
               >
-                <FontAwesomeIcon icon={faMagic} fixedWidth /> Auto
+                <FontAwesomeIcon icon={faWandMagicSparkles} fixedWidth /> Auto
               </DropdownItem>
             </DropdownMenu>
             <DropdownItem
--- a/imports/client/components/CallSection.tsx
+++ b/imports/client/components/CallSection.tsx
@@ -27,7 +27,6 @@ import MeteorUsers from "../../lib/models/MeteorUsers";
 import type { PeerType } from "../../lib/models/mediasoup/Peers";
 import mediasoupRemoteMutePeer from "../../methods/mediasoupRemoteMutePeer";
 import type { Action, CallState } from "../hooks/useCallState";
-import type { Theme } from "../theme";
 import Avatar from "./Avatar";
 import Loading from "./Loading";
 import Spectrum from "./Spectrum";
@@ -40,7 +39,7 @@ import {
   PeopleListDiv,
 } from "./styling/PeopleComponents";
 
-const CallStateIcon = styled.span<{ theme: Theme }>`
+const CallStateIcon = styled.span`
   font-size: 10px;
   width: 14px;
   height: 15px;
@@ -53,7 +52,7 @@ const CallStateIcon = styled.span<{ theme: Theme }>`
   background: white;
 `;
 
-const MutedIcon = styled(CallStateIcon)<{ theme: Theme }>`
+const MutedIcon = styled(CallStateIcon)`
   top: 0;
   border-bottom-left-radius: 6px;
   border: 0.5px solid ${({ theme }) => theme.colors.mutedIconBorder};
@@ -62,7 +61,7 @@ const MutedIcon = styled(CallStateIcon)<{ theme: Theme }>`
   color: ${({ theme }) => theme.colors.mutedIconText};
 `;
 
-const DeafenedIcon = styled(CallStateIcon)<{ theme: Theme }>`
+const DeafenedIcon = styled(CallStateIcon)`
   bottom: 0;
   border-top-left-radius: 6px;
   border: 0.5px solid ${({ theme }) => theme.colors.deafenedIconBorder};
@@ -73,7 +72,7 @@ const DeafenedIcon = styled(CallStateIcon)<{ theme: Theme }>`
   font-size: 9px;
 `;
 
-const RemoteMuteButton = styled.div<{ theme: Theme }>`
+const RemoteMuteButton = styled.div`
   position: absolute;
   inset: 0;
   font-size: 24px;
--- a/imports/client/components/GuessQueuePage.tsx
+++ b/imports/client/components/GuessQueuePage.tsx
@@ -31,7 +31,6 @@ import { useBreadcrumb } from "../hooks/breadcrumb";
 import useFocusRefOnFindHotkey from "../hooks/useFocusRefOnFindHotkey";
 import useTypedSubscribe from "../hooks/useTypedSubscribe";
 import indexedDisplayNames from "../indexedDisplayNames";
-import type { Theme } from "../theme";
 import CopyToClipboardButton from "./CopyToClipboardButton";
 import GuessState from "./GuessState";
 import {
@@ -71,7 +70,7 @@ const StyledHeaderRow = styled.div`
   display: contents;
 `;
 
-const StyledHeader = styled.div<{ theme: Theme }>`
+const StyledHeader = styled.div`
   position: sticky;
   top: 0;
   background-color: ${({ theme }) => theme.colors.background};
@@ -84,7 +83,7 @@ const StyledHeader = styled.div<{ theme: Theme }>`
   )}
 `;
 
-const StyledRow = styled.div<{ $state: GuessType["state"]; theme: Theme }>`
+const StyledRow = styled.div<{ $state: GuessType["state"] }>`
   display: contents;
   margin-bottom: 8px;
   background-color: ${({ theme, $state }) => theme.colors.guess[$state].background};
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -90,7 +90,6 @@ import useDocumentTitle from "../hooks/useDocumentTitle";
 import useSubscribeDisplayNames from "../hooks/useSubscribeDisplayNames";
 import useTypedSubscribe from "../hooks/useTypedSubscribe";
 import indexedDisplayNames from "../indexedDisplayNames";
-import type { Theme } from "../theme";
 import { trace } from "../tracing";
 import BookmarkButton from "./BookmarkButton";
 import ChatMessage from "./ChatMessage";
@@ -199,7 +198,6 @@ const PUZZLE_PAGE_PADDING = 8;
 const ChatMessageDiv = styled.div<{
   $isSystemMessage: boolean;
   $isHighlighted: boolean;
-  theme: Theme;
 }>`
   padding: 0 ${PUZZLE_PAGE_PADDING}px 2px;
   overflow-wrap: break-word;
--- a/imports/client/components/SetupPage.tsx
+++ b/imports/client/components/SetupPage.tsx
@@ -46,7 +46,6 @@ import type { DiscordGuildType } from "../discord";
 import { useBreadcrumb } from "../hooks/breadcrumb";
 import useTypedSubscribe from "../hooks/useTypedSubscribe";
 import lookupUrl from "../lookupUrl";
-import type { Theme } from "../theme";
 import ActionButtonRow from "./ActionButtonRow";
 
 const PageContainer = styled.div`
@@ -57,7 +56,7 @@ const Section = styled.section`
   margin-bottom: 24px;
 `;
 
-const SectionHeader = styled.h1<{ theme: Theme }>`
+const SectionHeader = styled.h1`
   background-color: ${({ theme }) => theme.colors.setupPageHeaderBackground};
   font-size: 18px;
   border-bottom: 1px solid black;
@@ -2312,7 +2311,7 @@ const CircuitBreaker = styled.div`
   margin-bottom: 16px;
 `;
 
-const CircuitBreakerRow = styled.div<{ theme: Theme }>`
+const CircuitBreakerRow = styled.div`
   display: flex;
   flex-direction: row;
   align-items: baseline;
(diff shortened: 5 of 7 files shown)


imports/client/components/App.tsx
```diff
@@ -23,16 +25,19 @@ import * as RRBS from "react-router-bootstrap";
...
 import { useBreadcrumbItems } from "../hooks/breadcrumb";
+import { AppThemeState, useAppThemeState } from "../hooks/persisted-state";
 import lookupUrl from "../lookupUrl";
 import ConnectionStatus from "./ConnectionStatus";
 import HuntNav from "./HuntNav";
 import Loading from "./Loading";
 import NotificationCenter from "./NotificationCenter";
 import { NavBarHeight } from "./styling/constants";
 import { mediaBreakpointDown } from "./styling/responsive";
+import { darkTheme, lightTheme, Theme } from "../theme";
```
zarvox: I think `Theme` should be a type import?

imports/client/components/AnnouncementToast.tsx
```diff
@@ -13,16 +16,32 @@ const AnnouncementToast = ({
...
+      className={className}
+      onClose={onClose}
+      style={{
+        backgroundColor: theme.colors.announcementToastBackground,
+        color: theme.colors.announcementToastText,
+        border: `1px solid ${theme.colors.announcementToastBorder}`,
+      }}
+    >
       <Toast.Header closeButton={!!onClose}>
+        <FontAwesomeIcon
```
zarvox: These icons appear unrelated to this PR?

imports/client/components/App.tsx
```diff
@@ -65,21 +70,21 @@ const BreadcrumbList = styled.ol`
 `;
 /* stylelint-enable value-no-vendor-prefix */
 
-const BreadcrumbItem = styled.li`
+const BreadcrumbItem = styled.li<{ theme: Theme }>`
```
zarvox: If this parameterization is required for type checking, we should probably have it everywhere `theme` is used.  If not, we should probably have it nowhere?

imports/client/components/PuzzlePage.tsx
```diff
@@ -260,10 +260,12 @@ const PuzzleContent = styled.div`
   flex-direction: column;
 `;
 
-const PuzzleMetadata = styled.div`
+const PuzzleMetadata = styled.div<{ theme: Theme }>`
   flex: none;
   padding: ${PUZZLE_PAGE_PADDING - 2}px 8px;
   border-bottom: 1px solid #dadce0;
+  z-index: 10;
```
zarvox: This `z-index` doesn't seem related to theming

imports/client/components/PuzzlePage.tsx
```diff
@@ -40,7 +40,7 @@ import Tooltip from "react-bootstrap/Tooltip";
 import { createPortal } from "react-dom";
 import { Link, useParams } from "react-router-dom";
 import type { Descendant } from "slate";
-import styled, { css } from "styled-components";
+import styled, { css, keyframes, useTheme } from "styled-components";
```
zarvox: I don't see any usage of `keyframes` in this file

imports/client/components/TagEditor.tsx
```diff
@@ -41,8 +43,89 @@ const TagEditor = ({
     .filter(Boolean)
     .map((t) => {
       return { value: t, label: t };
+    })
+    .sort((a, b) => {
```
zarvox: This doesn't seem to be theming-related

imports/client/components/GuessQueuePage.tsx
```diff
@@ -84,11 +84,25 @@ const StyledHeader = styled.div`
...
+      case "intermediate":
+        return theme.colors.guessColorIntermediateBackground;
+      case "incorrect":
+        return theme.colors.guessColorIncorrectBackground;
+      case "rejected":
+        return theme.colors.guessColorRejectedBackground;
+      case "pending":
+        return theme.colors.guessColorPendingBackground;
+      default:
+        return "transparent";
```
zarvox: If we think this is impossible and the switch is exhaustive, you can ensure that remains the case via a:

```typescript
  default:
    // biome-ignore lint/nursery/noUnusedExpressions: exhaustive check
    $state satisfies never
```

imports/client/components/FancyEditor.tsx
```diff
@@ -126,19 +128,36 @@ export const MentionSpan = styled.span<{
...
-      color: #fff;
+      background-color: ${({ theme }) => theme.colors.mentionSpanBackground};
+      color: ${({ theme }) => theme.colors.mentionSpanText};
     `}
   font-size: 0.8rem;
 `;
 
-const SelectedMentionSpan = styled(MentionSpan)`
-  box-shadow: 0 0 0 2px #b4d5ff;
+export const PuzzleSpan = styled.span<{
```
zarvox: This doesn't appear to be used in this patchset?  It's imported but not used in `ChatMessage.tsx`.  Maybe connected to some other feature?

imports/client/components/CallSection.tsx
```diff
@@ -27,6 +27,7 @@ import MeteorUsers from "../../lib/models/MeteorUsers";
 import type { PeerType } from "../../lib/models/mediasoup/Peers";
 import mediasoupRemoteMutePeer from "../../methods/mediasoupRemoteMutePeer";
 import type { Action, CallState } from "../hooks/useCallState";
+import { Theme } from "../theme";
```
zarvox: this (and others) will want to be a type import.  Should be something `biome check --write` can fix.

imports/client/theme.ts
```diff
@@ -0,0 +1,372 @@
...
+export type SolvednessState = keyof Theme["colors"]["solvedness"];
+
+export interface Theme {
+  basicMode: string;
+  colors: {
+    background: string;
+    text: string;
+    primary: string;
+    secondary: string;
+    muted: string;
```
zarvox: This field appears unused in this patchset

imports/client/theme.ts
```diff
@@ -0,0 +1,372 @@
...
+    danger: string;
+    warning: string;
+    info: string;
+    success: string;
```
zarvox: so are `warning`, `info`, and `success`.

imports/client/theme.ts
```diff
@@ -0,0 +1,372 @@
...
+    callStateIcon: string;
+    mutedIconBorder: string;
+    deafenedIconBorder: string;
+    mutedIconText: string;
+    deafenedIconText: string;
```
zarvox: This doesn't appear to be used in this patchset.

imports/client/components/FancyEditor.tsx
```diff
@@ -126,19 +128,36 @@ export const MentionSpan = styled.span<{
...
-  color: #4649ef;
-  background-color: #ced0ed;
+  color: ${({ theme }) => theme.colors.mentionSpanText};
+  background-color: ${({ theme }) => theme.colors.mentionSpanBackground};
   ${({ $isSelf }) =>
     $isSelf &&
     css`
-      background-color: #4649ef;
-      color: #fff;
+      background-color: ${({ theme }) => theme.colors.mentionSpanBackground};
```
zarvox: Consider preserving an emphasized coloration for mentions that refer to the logged-in user (someone was highlighting you, in particular).  This styling (and the line that follows) basically undoes the work done to identify `$isSelf` by making it styled exactly the same.

imports/client/components/FancyEditor.tsx
```diff
@@ -126,19 +128,36 @@ export const MentionSpan = styled.span<{
...
+  border-radius: 4px;
+  color: ${({ theme }) => theme.colors.mentionSpanText};
+  background-color: ${({ $solvedness, theme }) => {
+    return theme.colors.solvedness[$solvedness];
+  }};
+  font-size: 0.8rem;
+`;
+
+const SelectedMentionSpan = styled(MentionSpan)<{ theme: Theme }>`
+  box-shadow: 0 0 0 2px ${({ theme }) => theme.colors.selectedMentionSpanShadow};
```
zarvox: This makes use of `selectedMentionSpanShadow`, but not `selectedMentionSpanBackground` or `selectedMentionSpanText`.  If they're not used, remove them from the theme; if they are, let's get that here.

imports/client/theme.ts
```diff
@@ -0,0 +1,372 @@
...
+    chatterSectionBackground: string;
+    firehosePreWrapBackground: string;
+    firehosePreWrapText: string;
+    codeBlockBackground: string;
+    codeBlockText: string;
+    mentionSpanBackground: string;
+    mentionSpanText: string;
+    selectedMentionSpanShadow: string;
+    selectedMentionSpanBackground: string;
+    selectedMentionSpanText: string;
```
zarvox: commented in FancyEditor, but `selectedMentionSpanBackground` and `selectedMentionSpanText` appear unused

imports/client/theme.ts
```diff
@@ -0,0 +1,372 @@
...
+    avatarInitialText: string;
+    avatarInitialBackground: string;
+    avatarSelfBorder: string;
+    avatarSelfShadow: string;
+    breadcrumbText: string;
+    breadcrumbBeforeText: string;
+    announcementToastBackground: string;
+    announcementToastText: string;
+    announcementToastBorder: string;
+    replyChatMessageBackground: string;
```
zarvox: This appears unused in this patchset -- I'm guessing it's part of a separate chat feature?

imports/client/theme.ts
```diff
@@ -0,0 +1,372 @@
...
+    hoverChatMessageBackground: string;
```
zarvox: ditto

imports/client/theme.ts
```diff
@@ -0,0 +1,372 @@
...
+    systemChatMessageBackground: string;
+    pinnedChatMessageBackground: string;
```
zarvox: this **is** used, but for highlighted messages (we don't yet have a pinned messages feature upstream).

imports/client/theme.ts
```diff
@@ -0,0 +1,372 @@
...
+    replyButtonText: string;
+    hoverReplyButtonText: string;
+    replyingToBackground: string;
```
zarvox: this and the two lines above it appear to be related to an unmerged chat feature

imports/client/components/App.tsx
```diff
@@ -239,12 +250,13 @@ const AppNavbar = () => {
   // Note: the .brand class on the <img> ensures that the logo takes up the
   // correct amount of space in the top bar even if we haven't actually picked
   // a nonempty source for it yet.
+
+  const theme = useTheme();
   return (
     <NavbarInset
       variant="light"
       style={{
-        backgroundColor: "#f0f0f0",
```
zarvox: any reason you dropped the `background-color` on the navbar here rather than theming it?

zarvox (changes requested):
Thanks for putting this together!  I wrote a boatload of inline comments, but a brief summary:

* I noticed a few chunks that look like they slipped in that seemed unrelated to theming that I'd prefer to consider orthogonally.  I expect at least some of them would be flagged by the linters; did you have a chance to run those locally?
* I'd prefer if the default app scheme (for those who don't have anything set in localStorage) remained light mode, with dark mode being an opt-in feature -- it's more consistent with the logged-out UI, and it's what our users are used to seeing.
* If it's not too much to ask, I'd also prefer if we could keep the contents of the `Theme` data structure self-consistent, and only include keys that are actually used somewhere.  I exhaustively flagged the keys that I think are not yet in use by this patchset; feel free to add them back alongside the functionality that makes use of them in future PRs.

imports/client/components/PuzzlePage.tsx
jimsug: Also noticed that we don't actually use `useTheme`

imports/client/components/GuessQueuePage.tsx
jimsug: Yes, I think so - I think that'll also allow us to remove the flat theme properties and simplify the theme definitions a little more.

imports/client/components/App.tsx
```diff
@@ -319,15 +347,22 @@ const App = ({ children }: { children: React.ReactNode }) => {
...
   return (
-    <div>
-      <NotificationCenter />
-      <AppNavbar />
-      <ConnectionStatus />
-      <ContentContainer className="container-fluid">
-        {errorBoundary}
-      </ContentContainer>
-    </div>
+    <ThemeProvider theme={theme}>
```
zarvox: Actually, let's hoist this all the way up to `Routes.tsx`.  That way we can be sure that the theme object is available to all routes, not just logged-in routes, in case any of these components get used elsewhere.

imports/client/components/PuzzleModalForm.tsx
```diff
@@ -405,7 +408,7 @@ const PuzzleModalForm = React.forwardRef(
             </FormLabel>
             <Col xs={9}>
               <Creatable
-                id={`${idPrefix}-new-puzzle-tags`}
```
zarvox: I think this `id` should remain.

imports/client/components/PuzzlePage.tsx
```diff
@@ -587,9 +585,9 @@ const ChatHistory = React.forwardRef(
 const ChatHistoryMemo = React.memo(ChatHistory);
 
 const StyledFancyEditor = styled(FancyEditor)`
-  flex: 1;
+  flex: 1 1 auto;
```
zarvox: Curious why this line is here; is there a bug this is fixing?  Would prefer it in a separate patchset if so

imports/client/components/SetupPage.tsx
```diff
@@ -2308,13 +2312,13 @@ const CircuitBreaker = styled.div`
...
 
-const CircuitBreakerRow = styled.div`
+const CircuitBreakerRow = styled.div<{ theme: Theme }>`
   display: flex;
   flex-direction: row;
   align-items: baseline;
   justify-content: space-between;
   margin-bottom: 8px;
-  background: #eef;
+  background: ${({ theme }) => theme.colors.background};
```
zarvox: Hmmm, I liked the slightly blueish background giving clearer delineation between these rows and their detailed descriptions, which this change loses -- could these get their original colors back?  

Happy to discuss any design change proposals (if that's what was intended) in subsequent PRs or on Discord, but I'd prefer to keep this PR contained to the plumbing.

imports/client/hooks/persisted-state.ts
```diff
@@ -2,6 +2,15 @@ import type { SetStateAction } from "react";
 import { useCallback } from "react";
 import { useLocalStorage } from "usehooks-ts";
 
+export type AppThemeState = "dark" | "light";
+export const useAppThemeState = () => {
+  return useLocalStorage<AppThemeState>("appTheme", "light");
+};
+
+export const usePersistedSidebarWidth = () => {
```
zarvox: This seems like it probably goes with a different feature 🙂

zarvox (approved):
This is looking great!

I noticed that the logged-out page looks a little blinding under the dark theme, but that's the only thing that I think this is still missing, so I'm just going to merge this now and then tack on a tiny commit for that immediately afterwards.

Thanks so much for extracting this and accommodating all my nitpicking 🙂

PR #2654, opened 2026-01-02
Merge collapsible panes to upstream

jimsug:
These changes introduce the ability to collapse both the metadata pane and the chat pane on `PuzzlePage.tsx`.

To do this, I've made some changes to make `SplitPaneMinus.tsx` reactive, and also created a `MinimizedChatInfo.tsx` component that shows basic information about who is active on the puzzle (although I'm now thinking that I should've added a minimal "mode" to the callersection or something to keep the logic in one place?).

Some design choices that I made for this feature:
- Currently, I don't persist the collapsed state of either the metadata pane or the chat pane. I think I want to err on the side of **not** making it easy for remote solvers to be ignored, accidentally or otherwise; however...
- I also restore the chat pane whenever a new message is sent to the puzzle chat, so I _think_ I could be okay with persisting collapse state of the chat pane; and
- I don't think the puzzle metadata collapse state should be persisted, because you probably want people to be able to see it when they open a puzzle, even if to confirm that the puzzle remains unsolved before they start looking at it

I was also able to use this as an opportunity to put the chat restore button into the MinimizedChatInfo cluster, which is a nice bonus! (It was previously separate and this caused weird positioning issues.)

### DIFF
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -17,7 +17,6 @@ import { faTimes } from "@fortawesome/free-solid-svg-icons/faTimes";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import type { ComponentPropsWithRef, FC } from "react";
 import React, {
-  type ReactElement,
   useCallback,
   useEffect,
   useId,


imports/client/components/PuzzlePage.tsx
```diff
@@ -2102,14 +2238,60 @@ const PuzzlePage = React.memo(() => {
...
+  useEffect(() => {
+    const currentLength = chatMessages.length;
+    if (
+      currentLength > prevMessagesLength.current &&
+      prevMessagesLength.current > 0
+    ) {
+      if (isChatMinimized) {
+        restoreChat();
+        setTimeout(() => {
+          chatSectionRef.current?.snapToBottom();
```
zarvox: This seems to be a third mechanism attempting to snap chat to the bottom.

imports/client/components/PuzzlePage.tsx
```diff
@@ -2196,23 +2383,63 @@ const PuzzlePage = React.memo(() => {
...
+            </OverlayTrigger>
+          )}
           <SplitPaneMinus
             split="vertical"
-            minSize={MinimumSidebarWidth}
+            minSize={isChatMinimized ? 1 : MinimumSidebarWidth}
             maxSize={-MinimumDocumentWidth}
             primary="first"
-            size={sidebarWidth}
+            size={effectiveSidebarWidth}
```
zarvox: `size=1` seems like it would be messy -- are we sure we don't want to bring back some sort of section collapse (i.e. some sort of `display:none`) instead?

zarvox:
#2680 built atop this and is now merged.  Thanks for the design/first pass!

PR #2660, opened 2026-01-04
Add tag manager

zarvox:
Adds a tag manager page based on Palindrome's implementation at https://github.com/jimsug/jolly-roger/blob/main/imports/client/components/TagBulkEditPage.tsx , updated to more closely align with our stylistic and functional practices.

There's a couple commits of pre-factoring to make this work more neatly: extracting a function that otherwise causes a circular dependency that leads to surprising import behavior, and extracting some reusable search-matching functions (one of which we make use of on the tag manager page).  Makes sense to review commit-by-commit.

### DIFF
--- a/imports/client/components/HuntNav.tsx
+++ b/imports/client/components/HuntNav.tsx
@@ -4,6 +4,7 @@ import { faBullhorn } from "@fortawesome/free-solid-svg-icons/faBullhorn";
 import { faFaucet } from "@fortawesome/free-solid-svg-icons/faFaucet";
 import { faMap } from "@fortawesome/free-solid-svg-icons/faMap";
 import { faReceipt } from "@fortawesome/free-solid-svg-icons/faReceipt";
+import { faTags } from "@fortawesome/free-solid-svg-icons/faTags";
 import { faUsers } from "@fortawesome/free-solid-svg-icons/faUsers";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import Nav from "react-bootstrap/Nav";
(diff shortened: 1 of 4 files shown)


imports/client/components/HuntNav.tsx
```diff
@@ -183,6 +184,11 @@ const HuntNav = () => {
           <StyledPuzzleListLinkLabel>Hunters</StyledPuzzleListLinkLabel>
         </StyledPuzzleListLinkAnchor>
 
+        <StyledPuzzleListLinkAnchor to={`/hunts/${huntId}/tags`} title="Tags">
+          <MenuIcon icon={faTags} />
+          <StyledPuzzleListLinkLabel>Tags</StyledPuzzleListLinkLabel>
+        </StyledPuzzleListLinkAnchor>
+
```
ebroder: This will display for everyone, right? Do we want to limit this to operators like the firehose? (Not necessarily advocating for one way or another but curious what your thoughts are)

imports/client/components/HuntTagManagerPage.tsx
```diff
@@ -0,0 +1,665 @@
...
+        (error?: Error) => {
+          if (error) {
+            setSubmitState(SubmitState.FAILED);
+            setErrorMessage(error.message);
+          } else {
+            setSubmitState(SubmitState.SUCCESS);
+            setErrorMessage("");
+          }
+        },
+      );
```
ebroder: I'll accept pushback that I'm moving the goalposts, but should the `renameTag` method be aware of the `meta-for:` / `group:` connection?

PR #2680, opened 2026-01-12
Support collapsing chat and metadata panes

zarvox:
Adapted from @jimsug's #2654.

I took the general visual design (and in particular kept the placement of various controls) but cleaned up the state management and a few gaps in the design.

The central clever insight that simplifies state management is that when chat is collapsed, we can simply not render the `ChatMessages` pane at all, so when it is no longer collapsed, we automatically get the desired behavior of "snap to the bottom" (to show the new message) on mount.  Then, the only additional change we require is to `setIsChatMinimized(false)` when `chatMessages.length` changes, which is a very simple `useEffect()`.

This also allows us to avoid tracking "last known size when not collapsed" as a separate value -- the last-known size is the only size we need to keep track of.

The metadata pane collapse is also simple, with a hook to re-expand it when the number of correct answers to the puzzle changes -- intended to prevent solvers from failing to notice that the puzzle has in fact been solved.  

I ensured that all tooltips were given ids, so that they'd get placed correctly, though I've still been observing the occasional transient frame where tooltips are getting placed off screen (and causing scrollbars to appear for a frame where I don't expect them).  I remember seeing stuff like that before, but I forget what the fix was, and if it was a bug in popper or react-bootstrap or something else.

I also made sure that the minimized UI respects the `disable.webrtc` flag.

One possible wart that I don't care deeply about: the un-collapse metadata section button can overlap the "start audio call" button at mobile widths; I consider this tradeoff acceptable: the original design would get stuck in the collapsed state at mobile widths since it didn't show the button at all, and it is the effect of a floating button.

### DIFF
--- a/imports/client/components/PuzzlePage.tsx
+++ b/imports/client/components/PuzzlePage.tsx
@@ -2200,6 +2200,16 @@ const PuzzlePage = React.memo(() => {
     };
   }, [onResize]);
 
+  useLayoutEffect(() => {
+    // Adjust sidebar width based on window size on initial mount
+    if (puzzlePageDivRef.current) {
+      const clientWidth = puzzlePageDivRef.current.clientWidth;
+      setSidebarWidth((sidebarWidth) => {
+        return Math.min(sidebarWidth, clientWidth - MinimumDocumentWidth);
+      });
+    }
+  }, []);
+
   useEffect(() => {
     if (activePuzzle && !activePuzzle.deleted) {
       ensurePuzzleDocument.call({ puzzleId: activePuzzle._id });


imports/client/components/MinimizedChatInfo.tsx
```diff
@@ -0,0 +1,247 @@
...
+    <MinimizedChatInfoContainer>
+      <OverlayTrigger
+        placement="right"
+        overlay={
+          <Tooltip id={`${idPrefix}-mini-restore`}>Restore Chat</Tooltip>
+        }
+      >
+        <VoiceButton onClick={onRestore} style={{ marginBottom: "4px" }}>
+          <FontAwesomeIcon icon={faChevronRight} />
+        </VoiceButton>
```
ebroder: It's a little weird that this is using a class called `VoiceButton`?

imports/client/components/PuzzlePage.tsx
```diff
@@ -2126,17 +2221,23 @@ const PuzzlePage = React.memo(() => {
...
       selfUser={selfUser}
+      onMinimizeMetadata={toggleMetadata}
     />
   );
-  const chat = (
+
+  const effectiveSidebarWidth = isChatMinimized ? 0 : sidebarWidth;
+
+  const chat = isChatMinimized ? (
+    <div />
```
ebroder: Do we actually need a DOM element here?

ebroder:
> I remember seeing stuff like that before, but I forget what the fix was, and if it was a bug in popper or react-bootstrap or something else.

For the record, this is a broad rant, and I'm not ready to do anything about it, but I'm increasingly salty at the state of Bootstrap (which seems to have fallen behind the rest of the ecosystem and isn't really managing to catch up) and suspicious we should *eventually* consider switching to something else.

In this specific case, Popper.js renamed itself to floating-ui almost 3 years ago, so the Bootstrap popup code is a developmental dead end at this point.

imports/client/components/PuzzlePage.tsx
```diff
@@ -2091,7 +2186,7 @@ const PuzzlePage = React.memo(() => {
     if (puzzlePageDivRef.current) {
       setSidebarWidth(
         Math.min(
-          DefaultSidebarWidth,
+          sidebarWidth,
           puzzlePageDivRef.current.clientWidth - MinimumDocumentWidth,
         ),
       );
```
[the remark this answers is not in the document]
zarvox: Adjusted to do this in a layout effect run once on mount.

ebroder (approved):
Still LGTM

PR #2681, opened 2026-01-13
Add support for i18n along with zh translation

welkin25:
my attempt using react-i18next and I think it'd be great to get some feedback before I go further.

In short wherever there's a string displayed to the user, e.g.
```
<div>hello world</div>
```
we'd need to wrap it like
```ts
const { t } = useTranslation("MyComponent");
<div>{t("greeting", "hello world")}</div>
```
I think it's probably the easiest to keep English text as the default string (so there's no need of creating translation files for English, except in a few instances where I find it more convenient), as well as using "en" as the fallback language. Then anyone who needs i18n can take the key ("greeting" in this example) and add the translation later.

For now I've organized the translation files (filename is the same as the namespace) by the component file where the string appears. So a string that appears in PuzzleListPage.tsx for example should be translated in /public/locales/{lang}/PuzzleListPage.json.

### DIFF
--- a/imports/client/components/NotificationCenter.tsx
+++ b/imports/client/components/NotificationCenter.tsx
@@ -172,8 +172,12 @@ const GuessMessage = React.memo(
 
     const idPrefix = useId();
 
+    const { t, i18n } = useTranslation();
+
     const extLinkTooltip = (
-      <Tooltip id={`${idPrefix}-ext-link-tooltip`}>Open puzzle</Tooltip>
+      <Tooltip id={`${idPrefix}-ext-link-tooltip`}>
+        {t("guessQueue.openPuzzleTooltip", "Open puzzle")}
+      </Tooltip>
     );
 
     const linkTarget = `/hunts/${puzzle.hunt}/puzzles/${puzzle._id}`;
@@ -188,8 +192,6 @@ const GuessMessage = React.memo(
       guess.state === "incorrect" ? "danger" : "outline-secondary";
     const rejectButtonVariant =
       guess.state === "rejected" ? "secondary" : "outline-secondary";
-
-    const { t, i18n } = useTranslation();
 
     const stageTwoLabels = {
       intermediate: t(
--- a/imports/client/components/HuntEditPage.tsx
+++ b/imports/client/components/HuntEditPage.tsx
@@ -18,7 +18,7 @@ import FormText from "react-bootstrap/FormText";
 import Modal from "react-bootstrap/Modal";
 import Row from "react-bootstrap/Row";
 import { createPortal } from "react-dom";
-import { useTranslation } from "react-i18next";
+import { Trans, useTranslation } from "react-i18next";
 import { useNavigate, useParams } from "react-router-dom";
 import DiscordCache from "../../lib/models/DiscordCache";
 import type {
@@ -71,6 +71,8 @@ const DiscordSelector = ({
   loading,
   options,
 }: DiscordSelectorProps) => {
+  const { t } = useTranslation();
+
   const onValueChanged: NonNullable<FormControlProps["onChange"]> = useCallback(
     (e) => {
       if (e.currentTarget.value === "empty") {
@@ -108,7 +110,9 @@ const DiscordSelector = ({
   }, [value, options]);
 
   if (loading) {
-    return <div>Loading discord resources...</div>;
+    return (
+      <div>{t("huntEdit.loadingDiscord", "Loading discord resources")}...</div>
+    );
   } else {
     return (
       <FormControl
@@ -423,7 +427,7 @@ const HuntEditPage = () => {
       </Modal.Body>
       <Modal.Footer>
         <Button variant="primary" onClick={toggleShowTermsOfUsePreview}>
-          Close
+          {t("common.close", "Close")}
         </Button>
       </Modal.Footer>
     </Modal>,
@@ -607,7 +611,7 @@ const HuntEditPage = () => {
           controlId={`${idPrefix}-hunt-form-submit-template`}
         >
           <FormLabel column xs={3}>
-            Submit URL template
+            {t("huntEdit.submitUrlTemplate.field", "Submit URL template")}
           </FormLabel>
           <Col xs={9}>
             <FormControl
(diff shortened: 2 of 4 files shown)


ebroder:
I haven't read through this in detail yet, but I'm curious about the choice to use component names for grouping translations. I feel like I'm not personally wild about it (it feels easy to get out of sync or cause extra churn if we do any sort of refactoring), although I don't have an obvious alternative (other than, vaguely, "something more semantic") and would take pushback on this if other folks disagree

public/google-signin.svg
zarvox: Don't delete this file either; it's relied upon at https://github.com/deathandmayhem/jolly-roger/blob/7fb15954923d4877b75863d16058f31f2e9ca137/imports/client/components/AccountFormHelpers.tsx#L106

welkin25:
I think I've enclosed almost all regular user facing strings in i18n functions such that the current state would meet my team's need, and would be happy if this gets reviewed again.

Should this be merged into jr, I don't have any concrete plans on how any new text ought to be i18n'd. For simplicity I'd say others can make changes in English without concern for i18n, and I (or anyone else who might need i18n) can make periodic PRs to update them in batch.

imports/client/components/NotificationCenter.tsx
```diff
@@ -589,21 +620,30 @@ const ChatNotificationMessage = ({
 
   const senderDisplayName = displayNames.get(cn.sender) ?? "???";
 
+  useTranslation("NotificationCenter");
```
ebroder: Is this dead code?

welkin25:
> I haven't read through this in detail yet, but I'm curious about the choice to use component names for grouping translations. I feel like I'm not personally wild about it (it feels easy to get out of sync or cause extra churn if we do any sort of refactoring), although I don't have an obvious alternative (other than, vaguely, "something more semantic") and would take pushback on this if other folks disagree

Actually I was just introduced to VSCode extension i18n-ally, which seems to be capable of resolving the "where to write/find translation" headache. So now I'm leaning towards semantic organization, maybe even in one monolithic {locale}.json with nested namespaces. I'll give this a try tomorrow!

welkin25:
> I'm also curious what the right flow is for maintaining the translations. I'm assuming that we're outsourcing all of our translation efforts to, like, you and maybe other folks on 喵喵喵 (although, you know, GenAI is surprisingly good at foreign languages), but I'm curious whether there's tooling that's worth building now to detect (a) new translatable strings (b) translatable strings that have drifted.

If we're committed to 100% i18n*, for (a) there's a rule `lint/nursery/noJsxLiterals`  that'd detect hardcoded string in jsx that helps enforce i18n compliance. As for (b), I haven't experimented with how well GenAI can automate this, but I think maybe what's needed is to move English text into en.json files as well (as opposed to leave them in as defaults), so that we have locale json files with identical structure and then AI can match the keys and compare the values. I'm not sure how everyone feels about that, you lose a bit of context when the code looks like

```ts
       {t("invite.instruction)}
```

instead of

```ts
       {t(
          "invite.instruction",
          `Invite someone to join this hunt. They'll get an email with
          instructions (even if they already have a Jolly Roger account)`,
        )}
```

Extensions like i18n-ally certainly mitigate this, but I see no reason everyone else should have this extra overhead when (as of now) we're the only users of this feature.

*: I'm not sure if the remaining ~10% of the stuff really needs to be translated though, a lot of the instruction for external integration at least for me is best left in English.

zarvox (approved):
I'm broadly happy with the design and implementation here!  I'm gonna attach a few spots where I saw strings not wrapped in the translation framework that seem like they probably ought to be (though I'm happy to leave whether you think they benefit from a specific zh translation up to you).

With that said, I recognize there may some longer-tail surface area here and I don't want to make you keep going through painful rebases before getting this landed, so let me know if you'd rather get this landed and follow up on the additional translatable strings async or just omit them from the zh translation.

On the HuntEditPage, there's several strings not t-ified:

* An [internal loading message](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/HuntEditPage.tsx#L111)
* The [Close button](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/HuntEditPage.tsx#L426) of the terms-of-use preview modal
* ["Submit URL template"](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/HuntEditPage.tsx#L610) and [explanation](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/HuntEditPage.tsx#L620-L638)
* ["Discord role for members"](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/HuntEditPage.tsx#L762) and the [explanation thereof](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/HuntEditPage.tsx#L772-L775)
* [This conditionally-displayed Alert](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/HuntEditPage.tsx#L783): Screenshot 2026-01-31 at 12 08 54


In NotificationCenter: 
* [The tooltip for the "Open puzzle" button](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/NotificationCenter.tsx#L176)
Screenshot 2026-01-31 at 12 17 00

* A couple of notifications that you might only encounter in slightly rarer situations:
  * The "update the google apps scripts" [header](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/NotificationCenter.tsx#L557), [notification](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/NotificationCenter.tsx#L560-L562) and [update button](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/NotificationCenter.tsx#L570)
  * The "profile missing" [notification header](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/NotificationCenter.tsx#L590) and [body](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/NotificationCenter.tsx#L593-L596)
  * The "we've done an update and we'd like to refresh the page but you're in an audio call so we're waiting to reload" [title](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/NotificationCenter.tsx#L736), [body](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/NotificationCenter.tsx#L740-L750),  and [reload button](https://github.com/deathandmayhem/jolly-roger/blob/265e9c59c04b405259e89810e7621fe798e56c11/imports/client/components/NotificationCenter.tsx#L755)
  
I'm out of time to look at things for now, but there's a few others.  Let me know what you think?

PR #2683, opened 2026-01-14
Reimplement TypedMethod and TypedPublication using zod

ebroder:
I would probably recommend looking at just the new implementation rather than the diff.

Our existing Typed* systems built on top of `check` are fine I guess, but zod is great for putting (a) type declarations (b) type validations in the same place, which is one less set of code that we need to write when adding a method. I especially like the simplification this offers for `createHunt` and `updateHunt`, where we previously had to maintain basically two versions of the same complex validation. (Although I still would need to deal with `Object.keys(HuntPattern)` and replace that with some zod probing.

I'll proceed with conversion if this seems sufficiently interesting. (After Mystery Hunt, of course)

### DIFF
--- a/imports/methods/createPuzzle.ts
+++ b/imports/methods/createPuzzle.ts
@@ -1,16 +1,22 @@
-import type { GdriveMimeTypesType } from "../lib/GdriveMimeTypes";
+import z from "zod";
+import GdriveMimeTypes from "../lib/GdriveMimeTypes";
 import TypedMethod from "./TypedMethod";
 
-export default new TypedMethod<
-  {
-    huntId: string;
-    title: string;
-    url?: string;
-    tags: string[];
-    expectedAnswerCount: number;
-    docType: GdriveMimeTypesType;
-    allowDuplicateUrls?: boolean;
-    completedWithNoAnswer?: boolean;
-  },
-  string
->("Puzzles.methods.create");
+export default new TypedMethod(
+  "Puzzles.methods.create",
+  z.tuple([
+    z.strictObject({
+      huntId: z.string(),
+      title: z.string(),
+      url: z.string().optional(),
+      tags: z.string().array(),
+      expectedAnswerCount: z.number(),
+      docType: z.enum(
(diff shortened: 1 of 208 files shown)


ebroder:
For the record, I *think* that the framework changes here will mostly survive the Zod 4 upgrade, modulo needing to change `z.AnyZodTuple` to `z.ZodTuple`. The individual schema definitions would have to change more, so it's a good thing I haven't done those conversions yet

ebroder:
(Pushing this up because I find GH's web interface to be better for code reading, but this is Claude's take that I haven't reviewed yet)

imports/methods/createPuzzle.ts
```diff
@@ -1,16 +1,22 @@
...
+      title: z.string(),
+      url: z.string().optional(),
+      tags: z.string().array(),
+      expectedAnswerCount: z.number(),
+      docType: z.enum(
+        Object.keys(GdriveMimeTypes) as (keyof typeof GdriveMimeTypes)[],
+      ),
+      allowDuplicateUrls: z.boolean().optional(),
+      completedWithNoAnswer: z.boolean().optional(),
+    }),
```
ebroder, line 19: Left as a potential exercise for some future reader: extract an `EditablePuzzle` a la `EditableHunt` and use it here.

ebroder:
OK. I think this is ready for review at this point. I broke it up into separate commits for ease of review:

* The first two commits are bugs discovered while doing the conversion (I believe they became type errors where they previously were silent)
* The third commit (9029c462f63d2752d5431966f8d0e0afaa1b48ed) introduces the new implementations, and the fourth (c180adbb5a9e6a9b4989274f633f6fd7da0f3aad) adds some quick test coverage. I'd recommend against looking at the diffs and instead just reading the new code de novo.
* The final two commits are rote conversion of all TypedMethods and TypedPublications. They are long but should be easy to review.

@zarvox, since I've already made you do an annoying rebase of #2808 on top of the Zod 4 branch, I'm fine sitting on merging this until after that lands so you don't have to do another annoying rebase.

PR #2707, opened 2026-01-20
dependabot: upgrade stylelint deps together

zarvox:
Guessing that this is why all of #2688, #2693, and #2695 are failing.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -33,6 +33,10 @@ updates:
           - "slate"
           - "slate-history"
           - "slate-react"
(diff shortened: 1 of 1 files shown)


ebroder:
Given that dependabot supports wildcards, do we want to word this as "stylelint*" or "stylelint" + "stylelint-*" or something?

(Also it looks like we might have to do some adjustments for the major version bump but we can take that separately)

zarvox:
Ahhh I suppose some indirect deps are also `stylelint*`, sure.

PR #2715, opened 2026-01-21
Allow external URL access from formulas by default in new sheets.

jpd236:
When creating a new spreadsheet, update it to allow external URL access. This avoids the (unclickable) permission request coming up every time we attempt to use a function like `=IMAGE` to load an external resource, and seems like a reasonable tradeoff of security for simplicity given the lack of sensitive information expected in a sheet. This can also streamline including helpful formulas in the default template which depend on external resources.

To minimize the added latency from the additional call, we run this in parallel to the permission grant request that is made after the file is created.

### DIFF
--- a/imports/client/components/SetupPage.tsx
+++ b/imports/client/components/SetupPage.tsx
@@ -679,7 +679,7 @@ const GoogleScriptForm = ({
           >
             Google Cloud console
           </a>{" "}
-          and click &quot;ENABLE API&quot; near the top of the page.
+          and click &quot;Enable&quot; near the top of the page.
         </li>
         <li>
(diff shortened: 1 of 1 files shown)


zarvox:
Do we also need to update https://github.com/deathandmayhem/jolly-roger/blob/0f3bf4bd7bc96e8e1835f8255829deb5e6d495dd/imports/client/components/SetupPage.tsx#L288-L303 to include an additional scope for the additional API, or do the existing scopes cover this API already?

And if so: we should make sure we communicate to all known users that they'll need to re-link the Drive user to get a token with the additional scope for this to work.

Issue #2284, opened 2024-10-01
Reduce noise from logged errors for bad user inputs


jpd236:
A few instances of bad user input cause client+server errors to be logged, which causes noise w/ BugSnag and notifications, and in some cases results in less clear user-facing errors. IMO, these would ideally be handled more gracefully to avoid the noise.

From our test run with GPH, I noticed the following:

- Invalid URL when creating puzzle (looks like there's some sort of regex powering this in the MongoDB schema)
- Puzzle title is empty when creating/updating puzzle
- Guess is empty when adding answer

PR #2749, opened 2026-01-30
Upgrade to Zod 4

ebroder:
This is still very much a work in progress, but I wanted to go ahead and push it up so it wasn't living exclusively in my head and on my computer.

Here's what I currently know to be broken on this branch, although at this point the brokenness is likely covering up additional issues:

- [x] Zod 4 + zod-to-mongo-schema don't currently support outputting schemas for `z.date()` or `z
  .instanceof(Uint8Array<ArrayBuffer>)` (which we use for BSON binary data). Like I say in the commit message below, we don't have any differentiated requirements around generating a MongoDB-compatible JSON schema from a Zod schema, so this *should* be a task that we can outsource to existing open-source software. I'm going to try and work with Zod and zod-to-mongo-schema upstreams to figure out how best to support this. It's possible that I will have to give up on my dream and re-implement schema generation in-house, but I don't want to do it just yet
  
  For the time being, I've worked around this by switching to a fork of zod-to-mongo-schema
- [x] The current approach for `Model.autoPopulatedFields()` runs into some issues invoking `Meteor.userId()` in contexts where a user ID isn't available, and calling `Meteor.userId()` outside of a method or publication throws an exception. (As an easy example, in `tests/unit/imports/server/Flags.ts` we provide a fake `createdBy` to try and side-step this, but `autoPopulatedFields` queries for a value unconditionally)
  
  I've fixed this by checking `DDP._Current{Method,Publication}Invocation` to make sure we're in a context where `Meteor.userId` should work.
- [x] We're hitting an assortment of issues with numeric types. Some of this is zod-to-mongo-schema attempting to be more specific about BSON "int" vs. "long" (instead of just using "number", which is mostly what we've done historically). Some of this is a Zod bug (it doesn't correctly conflict between minimum and exclusiveMinimum or maximum and exclusiveMaximum when encoding a draft-04 JSON schema)
- [x] Zod 4 does not emit a regex pattern in JSON Schema when encoding `z.string().url()`, which is a regression from our previous behavior. These patterns on Puzzle and Hunt are arguably load-bearing right now, since we don't do any other validation of user-provided URLs (other than ensuring that they're a string). This came up in #2284 and is maybe not ideal behavior, but we should probably tackle those orthogonally.
  
  I've fixed this by re-introducing our regex and attaching it to any `z.url()` fields.
- [x] The new approach to discriminated unions is buggy. The Zod 4 schema ends up with something like:
  
  ```
  allOf: [
    { oneOf: [... variants with {name, value} ...], additionalProperties: false },
    { properties: {createdAt, updatedAt, createdBy, updatedBy}, additionalProperties: false },
    { properties: {deleted}, additionalProperties: false },
    { properties: {_id}, additionalProperties: false }
  ]
  ```

  which is unsatisfiable, because each entry rejects every other.

Here's the commit message, which describes the approach:

> In truth, this is largely a ground-up reimplementation of our typed model layer that takes advantage of a few years of experience working with the original implementation. Upgrading to Zod 4 is *somewhat* orthogonal.
> 
> My primary goal with this upgrade was to take advantage of Zod 4's native support for JSON Schemas. There already exists a library (zod-to-mongo-schema) which can utilize Zod's native JSON Schema support to generate MongoDB-compatible JSON Schemas. Given that our needs here are fairly undifferentiated, it doesn't feel like there's significant benefit to maintaining this code ourselves. This allows us to clean up the old code to generate JSON Schemas.
> 
> As always, the most complex element of our typed model layer is handling fields which should be automatically populated. Our previous approach (documented in #1394) was powerful and flexible, but abusing Zod's input vs. output schemas and lying to the type system with transforms broke down with Zod's native JSON Schemas, which refuses to serialize transforms.
> 
> Instead of trying to modernize that approach, this commit uses two systems to track automated fields. Within the type system, we use Zod's branded types to mark fields which should be auto-populated (effectively as a boolean marker), which we can then use to filter auto-populated fields, making them optional at insertion-time. At runtime, we use schema metadata, which captures when a field should be populated (insert, update, or both) and what value should be used (which has been limited to the specific types of values we actually use).
> 
> (In bringing in branded types, we do return to our old friend of input/output schemas — auto-populated types are only branded on the input side, which is only used for computing the insert type, since we don't actually want nominal typing for our auto-populated fields.)
> 
> This has the effect of dropping support for arbitrary transformation functions. In practice, we were only using that function to apply the `answerify` transform to answer fields, but we had previously concluded that answers were already uppercased at input time. In exchange for dropping transforms, we no longer need to analyze the entire update operation to make sure transformations are applied properly, and can instead just rely on MongoDB to enforce that the result matches our schema. This in turn lets us remove a significant amount of code around updates (including the mechanisms around `relaxSchema` and `parseMongoModifierAsync`).
> 
> In addition to dropping support for transform functions, I also dropped support for the `bypassSchema` option for inserts, updates, and upserts. This wasn't a great abstraction, since it attempted to otherwise preserve the behavior of Meteor Collections' native methods, but had to do some extra work to do that. Instead, users can just reach for rawCollection. This only comes up in migration code anyway, which generally requires a fair amount of abstraction bypass regardless.
> 
> And finally, I pulled the code into imports/lib/typedModel, rather than leaving it strewn about with actual model declarations.

And in terms of review sequence, I'd recommend something like this:
1. `autoPopulate.ts`
2. `customTypes.ts`
3. `validateSchema.ts`
4. `Model.ts`
5. `SoftDeletedModel.ts`


ebroder:
Created https://github.com/colinhacks/zod/pull/5700 for the first of those issues.

ebroder:
I asked Claude to dump out the schemas and advertise the differences. Here's what it came up with:

## Schema Diff Report: Zod 3 (main) vs Zod 4

### Systematic Changes (all/most collections, superficial)

These appear across virtually every collection and are safe:

1. **`bsonType` → `type` for standard types**: `"string"`, `"object"`, `"array"`, `"bool"`→`"boolean"` all switch from MongoDB's `bsonType` to standard JSON Schema `type`. BSON-specific types (`"date"`, `"int"`, `"binData"`) are preserved. MongoDB accepts both.

2. **`deleted` field simplified**: `allOf: [{not: {bsonType: "null"}}, {bsonType: "bool"}]` → `{type: "boolean"}`. Semantically equivalent since `boolean` already excludes null. Same pattern applies to other defaulted fields (`openSignups`, `mailingLists` in `jr_hunts`).

3. **`deleted`, `updatedAt`, `tags`, `mailingLists`, `openSignups` now required**: Fields with `.default()` values are now listed as required. These will always be populated on insertion.

4. **Enum fields gain explicit `type: "string"`**: e.g. `{enum: ["audio","video"]}` → `{enum: ["audio","video"], type: "string"}`. More explicit but equivalent.

5. **`exclusiveMinimum: false` / `exclusiveMaximum: false` removed**: These were the default values and had no effect. Affects `jr_blobs.size`, `jr_guesses.confidence`/`direction`, `jr_puzzles.expectedAnswerCount`, and `port` fields.

6. **`anyOf` → `oneOf` for discriminated unions** (`jr_documents`, `jr_settings`): More correct for mutually exclusive variants. Equivalent in practice with single-value enum discriminators.

7. **Flatter `allOf` structure** (`jr_documents`, `jr_settings`): Zod 3 produced deeply nested `allOf` with `{}` placeholder properties; Zod 4 produces a cleaner flat structure. Semantically equivalent.

### Concerning Changes

#### `port` field lost `minimum: 0` (high risk)

**Affected**: `jr_mediasoup_monitor_connect_acks.port`, `jr_mediasoup_monitor_connect_requests.port`

| | main | zod-4 |
|---|---|---|
| Schema | `{minimum: 0, exclusiveMinimum: true, maximum: 65535, bsonType: "int"}` | `{exclusiveMinimum: true, maximum: 65535, bsonType: "int"}` |

In JSON Schema draft-04, `exclusiveMinimum` is a boolean modifier of `minimum`. Without `minimum`, `exclusiveMinimum: true` is meaningless — **the port > 0 constraint is lost**. This is the Zod `z.positive()` bug already reported upstream.

#### `answer`/`guess` fields gained constraints (medium risk)

**Affected**: `jr_puzzles.answers[]`, `jr_guesses.guess`, `jr_bookmark_notifications.answer`

| | main | zod-4 |
|---|---|---|
| Schema | `{bsonType: "string"}` | `{type: "string", minLength: 1, pattern: "^[^a-z]+$"}` |

The `answer` custom type likely uses a `.pipe()` that Zod 3's JSON Schema converter couldn't serialize but Zod 4's can. This is **more correct** (answers should be uppercase), but it's a real validation tightening. Existing documents with lowercase answers would fail validation on update.

#### Email regex changed (medium risk)

**Affected**: `users.emails[].address`, `jr_folder_perms.googleAccount`

| | main | zod-4 |
|---|---|---|
| Pattern | `` ^[a-zA-Z0-9.!#$%&'*+/=?^_`{⏐}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$ `` | `^(?!\.)(?!.*\.\.)([A-Za-z0-9_'+\-\.]*)[A-Za-z0-9_+-]@([A-Za-z0-9][A-Za-z0-9\-]*\.)+[A-Za-z]{2,}$` |

Zod 4 ships a different `z.string().email()` regex. The new one disallows leading/consecutive dots, has a narrower local-part character set, and requires a TLD of 2+ chars. **Could reject previously-valid email addresses.**

#### URL patterns lost (low risk)

**Affected**: `jr_puzzles.url`, `jr_hunts.homepageUrl`, `jr_chatmessages`/`jr_chatnotifications` image child `url`

Zod 3 emitted a URL-matching regex from `z.string().url()`. Zod 4 emits no pattern — just `{type: "string"}`. This is a validation loosening (won't reject valid documents, but won't catch invalid URLs at the MongoDB level). Likely `z.toJSONSchema()` doesn't serialize URL checks.

#### UUID regex updated (low risk)

**Affected**: All UUID fields across mediasoup collections (`consumerId`, `producerId`, `routerId`, `transportId`)

| | main | zod-4 |
|---|---|---|
| Version digit | `[1-5]` | `[1-8]` |
| Variant nibble | unconstrained | `[89abAB]` |
| Max UUID | not allowed | allowed |

More permissive on version range (6-8), slightly more restrictive on variant nibble. All standard RFC 4122 UUIDs have variant bits in `[89abAB]`, so existing data should be fine.

### Summary

| Change | Risk | Action needed? |
|---|---|---|
| `bsonType` → `type` | None | No |
| `deleted`/`updatedAt`/`tags` etc. now required | None (have defaults) | No |
| `allOf[not null, bool]` → `boolean` | None | No |
| Enums gain `type: "string"` | None | No |
| `anyOf` → `oneOf` | None | No |
| Flatter `allOf` nesting | None | No |
| `exclusiveMinimum/Maximum: false` removed | None | No |
| UUID regex updated | Low | Monitor, likely fine |
| URL patterns lost | Low | Upstream: `z.toJSONSchema()` / `zod-to-mongo-schema` |
| Email regex changed | Medium | Verify existing emails pass new regex |
| `answer`/`guess` gained constraints | Medium | Correct behavior, but verify no lowercase data exists |
| `port` lost `minimum: 0` | **High** | Needs fix (Zod bug already reported) |

ebroder:
Couple of notes on that:

- The URL regex patterns on Puzzle and Hunt are arguably load-bearing right now, since we don't do any other validation of user-provided URLs (other than ensuring that they're a string). This came up in #2284 and is maybe not ideal behavior, but we should probably tackle those orthogonally.
- The new approach to discriminated unions is buggy. The Zod 4 schema ends up with something like:
  
  ```
  allOf: [
    { oneOf: [... variants with {name, value} ...], additionalProperties: false },
    { properties: {createdAt, updatedAt, createdBy, updatedBy}, additionalProperties: false },
    { properties: {deleted}, additionalProperties: false },
    { properties: {_id}, additionalProperties: false }
  ]
  ```

  which is unsatisfiable, because each entry rejects every other.

ebroder:
Opened https://github.com/colinhacks/zod/pull/5702 which I believe will fix the discriminated unions.

ebroder:
I've picked this back up again in an attempt to get it to the point that we can push it over the finish line. Here are changes since the last time the code was pushed up:

* Since Zod seems to have gone back into hibernation, instead of waiting for colinhacks/zod#5702 to land, instead add a `fixAllOfAdditionalProperties` wrapper function that post-processes Zod's JSON Schema. (Basically does the same thing that I did in the PR). Also add a test to validate the specific behavior we care about.
* Drop auto-generating an `_id` value in `autoPopulatedFields()`. This broke `upsert({_id: 'foo'}, [...])` because the `_id` value in `$setOnInsert` would conflict with the selector.
* Treat `update(..., {})` as a no-op modifier rather than a full document replacement, matching MongoDB's behavior
* If `autoPopulatedFields` leaves `$set` or `$setOnInsert` empty, clean it up (otherwise it's an error)
* Add the range restrictions on `portNumber` back, now that colinhacks/zod#5700 is merged and released
* Wrap default values in a lambda so that we don't capture them once and reuse the exact objects (i.e. don't recreate the Python problem)
* Make `validateSchema` more strict, since we changed `zodToMongoSchema` to be less strict
* Update `InsertModelType` to do more :sparkles:TypeScript magic:sparkles: and correctly distribute over unions, rather than collapsing unions down to a common type.

Pushing this up now, but I'm going to do another attempt at getting a comparable schema dump, this time by trying to do some harmonization transforms on both the old and new schemas so that they hopefully can be compared more directly; will post that once I have it.

ebroder:
OK. I've looked through the comparator that Claude wrote and I think I'm mostly satisfied that it's comparing the right things. Here's the set of canonicalizations that it's running both the old and new schemas through:

* Type unification: Collapse `bsonType` vs. `type` into a single field, and unify type names (e.g. `boolean` vs. `bool`). Also remove the `type: "string"` that the new generator emits for enums (it's inferred if absent)
* Normalize discriminated unions: Zod 4 emits discriminated unions using `oneOf`; the old code used `anyOf`. These only differ in behavior if an object can match more than one, which isn't possible with a discriminated union. Detect discriminated unions and rewrite their `oneOf` to `anyOf` for compatibility
* Unwrap default not-null handling: The old generator wrapped fields with defaults in an`allOf: [{not: {null}}, inner]` structure, I suspect in a failed attempt to mark them as required, but which was actually meaningless and absent from the Zod 4 schema, so unwrap that
* Flatten regex patterns: The old generator would use `allOf: [{pattern}]` to apply constraints; pull those up to the top of the field definition
* Flatten intersections: The old code would generate complex trees to represent various intersection types; Zod 4 flattens them where possible, so we need to flatten them too.
* Cleanup defaults: The old generator always emits `exclusiveMinimum` or `exclusiveMaximum`, but those both default to `false` and are excluded in the new code
* Sort arrays of required keys, object keys

And after doing all of that, here were the residual differences:

* Fields with a default value are now required; previously they were optional. (Discussed in the commit message)
* `answer` fields gained `pattern: "^[^a-z]+$"` — `jr_guesses.guess`, `jr_puzzles.answers[]`, `jr_bookmark_notifications.answer`. Intentional change, discussed in the commit message as a replacement for the old transformation approach. I've confirmed that our prod database is clean, and done a historical code audit to conclude that it shouldn't have ever been possible to insert records that didn't match this.
* Email regex got stricter. I stole the old regex from Stack Overflow or something; Zod 4 embeds a new one. It's somewhat stricter: it rejects leading/doubled dots and some other symbols in the local part, single-label domains (a@b), and domains whose TLD label isn't purely alphabetic, which includes punycode TLDs like xn--p1ai.
* UUID regex changed. Same deal - old regex I copy-pasted from somewhere; new one is Zod-embedded. It is more generous on UUID versions (supports v1-8 and a few other misc values) but less generous on variants (old accepted any variant, new requires a valid variant). Mediasoup generates valid v4 UUIDs so this should be harmless.
* `homepageUrl` on `Hunts` used to be `minLength: 1` and no longer is. This is fine because the URL regex won't accept an empty string.
* `jr_merge_operations.participants`: Zod 4 doesn't set `additionalItems: false` on tuples, so the schema would have accepted an array longer than the spec. This is arguably an upstream Zod bug, but it's also pretty harmless here since we don't read back the `participants` field, we just use it for the unique index. Still, as a safety net, I added `minItems: 2, maxItems: 2` as an override.

I'm mostly comfortable with all of these. I could be convinced that the email regex is concerning and we should widen that. (Maybe that's a question for @welkin25 to see if rejecting Punycode TLDs is likely to be a problem for i18n/l10n reasons?)

If you want to poke at the raw data and compare the schemas yourself:
* [schemas-main.json](https://github.com/user-attachments/files/30208562/schemas-main.json)
* [schemas-zod4.json](https://github.com/user-attachments/files/30208945/schemas-zod4.json)

zarvox (approved):
I read through the resulting JSON schemas (and I'd read through diff this the first time), and I poked around at a couple things and they all seemed to work as expected, so if you're happy with this now, I am too

welkin25:
Can't speak for others but I've never seen non-alpha email TLDs on our team (or anyone else for that matter) so this won't be a problem!

ebroder:
Three quick changes before merging:
- The URL regex was too strict. It didn't matter before because it wasn't anchored on `^` and `$` and it matched enough of a URL to be fine. But we need the anchors to replace validation we were previously getting from Zod's native URL validation, and that exposed that it didn't accept a bunch of valid URLs
- `z.ZodType<Solvedness>` got inferred as an input type of `unknown` and effectively loosened the type, so use `satisfies` instead of a cast
- `fixAllOfAdditionalProperties` (and the upstream fix I proposed) had an issue with nested `allOf` -> `anyOf` schemas, where it would effectively widen what was supposed to be a disjoint union.

But I think otherwise this should be ready to go.

PR #2752, opened 2026-02-01
Update top-level documentation files

ebroder:
Easy stuff first: update the README.md to reflect the last (*checks notes*) 6 years of changes. It hasn't seen meaningful updates since 2020. This includes discussion of new features, like audio/video calls, dingwords, dark mode, and internationalization.

Also update DEVELOPMENT.md to not throw as much shade at us (we're not on Meteor 2 or Node 16 anymore!). I also opted to split DEVELOPMENT.md into a DEVELOPMENT.md and a DEPLOYMENT.md, since much of the content felt disconnected from *development* of the app.

But IMO the best part of this is the screenshots in the README file, which were about as stale as the README file itself. I've added a new script that can generate all of the screenshots in the README file using a temporary clean setup of Jolly Roger and Playwright. In order to show off our dark-mode support, it takes these screenshots in both light and dark mode, and in the README file, we use `<picture>` tags to automatically swap out the screenshots to match your system's current display mode. (This uses a technique [originally documented by GitHub](https://github.blog/developer-skills/github/how-to-make-your-images-in-markdown-on-github-adjust-for-dark-mode-and-light-mode/)).

I mostly assume that this script will not hold up against the second law of thermodynamics and will degrade over time. We could protect against that drift by integrating it into our CI process, but that seems heavyweight and not especially useful (and would probably add a good 30-60 seconds to CI time). I think that's OK - when we want to update the screenshots again in the future, I'm hoping it won't be a ton of work to update the script to start working again.


imports/ScreenshotFixture.ts
```diff
@@ -0,0 +1,44 @@
+// Users
+
+export const PRIMARY_USER = {
+  email: "drew@deathandmayhem.com",
+  password: "screenshots-password",
+  displayName: "Drew",
```
zarvox: Honored to be the primary user 🙂

zarvox:
Did...did we regress font loading, or is it just a matter of not waiting long enough for the browser to finish loading the font before taking the screenshot?  I'm looking at the onion skins of https://github.com/deathandmayhem/jolly-roger/compare/07f6e9ff19c8b0c8cfa896193aa67b1652268edc..38bf27c1656e553da83fc20f3e4bb491d6daa420 and seeing a lot of places where the type is now different and (because it's less horizontally compact) we now overflow into a second row in a bunch of places.  The guess queue is particularly noticeable, but dingwords also show a difference in type, and the puzzle list's horizontal packing is nontrivially affected by the font metrics.

Oh, and the puzzle page seems to have not loaded the Sheet.

The diff also includes that running the script at a two-digit hour changes what we see for timestamps, but I think that's reasonable.

ebroder:
Unfortunately, GitHub is pretty restrictive in how it's willing to render mp4s in Markdown files. They specifically have to be served off of githubusercontent (i.e. uploaded through drag-and-drop on the web interface); you can't reference them from within a repository like you can with images.

We could theoretically use webp, which seems like it is supposed to be more efficient than animated GIFs, but in practice all of the files I've been able to generate have been at least 3x the size of the GIFs.

README.md
```diff
@@ -27,23 +57,66 @@ show their name, rather than "Anonymous Animal".
...
-losing relevant discussion to 10000-message retention limits in the middle
-of the Hunt. Supports basic formatting -- bulleted lists, \_italics\_,
-\*bold\*, \`monospace\`, and \`\`\`code blocks\`\`\`.
+losing relevant discussion from last year to a paid service's retention limits.
+
+Chat supports rich text editing, including bulleted lists, \_italics\_,
+\*bold\*, \`monospace\`, \`\`\`code blocks\`\`\`, and @-mentions with formatting
+previews. You can also upload images directly into chat.
+
+We work hard to make sure that chat activity is visible whenever there is
```
jimsug: > We work hard to make sure chat activity is visible whenever there is activity

looks like a typo, even though I suspect it may not be? Could be worth rewording to "the puzzle chat" or "chat section" or similar, but might be nitpicking.

ebroder:
Grump. Yeah I guess they're just using a different sanitizer for mobile rendering. I'll add a separate link so that something shows up

PR #2754, opened 2026-02-01
Work around issue with device selection

ebroder:
Strangely, getUserMedia doesn't seem to always respect the provided deviceId unless it's constrained with the "exact" keyword. In particular, I regularly observed that even with a valid deviceId, Chrome would sometimes prefer the *first* available audio input device (which, annoyingly, happened to be my iPhone in Continuity Camera mode).

So instead, require the exact deviceId, and since it's possible that device is no longer available, fall back to the "default" device (which should match the OS default) before aborting.


ebroder:
(Also for what it's worth, I bet that in the year 2026 we could drop the WebKit vendor prefix fallback too, although I didn't do it here)

PR #2774, opened 2026-02-10
Add job queue with Discord role sync as first job

ebroder:
> Add job queue with Discord role sync as first job
> 
> Introduce a general-purpose job queue backed by MongoDB:
> - Jobs model with TTL index for automatic cleanup
> - TypedJob for type-safe job definitions with Zod validation
>   - Args schema constrained to records at the type level
>   - Optional result schema with setResult callback for
>     incremental progress updates
> - Job worker with atomic claiming, retry with backoff, and
>   graceful shutdown
> - Dead server cleanup via existing garbage-collection hooks
> 
> Framework code lives in imports/server/jobs/framework/,
> with concrete job implementations alongside in
> imports/server/jobs/.
> 
> Migrate Discord role sync from inline execution in
> addUsersToDiscordRole to a queued job, preserving the same
> logic and rate limiting.
> 
> Add admin-only publication for job monitoring.

This is a sort of proof-of-concept for something I've been thinking about for a while, because I think it's well-suited to a class of problems we have. Specifically: stuff that is potentially long-running that wants to support retries and some vague notion of feedback.

This PR starts by converting the Discord role sync operation, mostly because it's relatively straightforward, but I have some specific use cases where I think it will actually be potentially higher leverage:

* Bulk invitations for users. (The `setResult` mechanism lets us deliver feedback, including incrementally, which gives us an angle on #549)
* Cleaning up the createdBy/updatedBy fields on existing records when we are merging users in #296

Plus potentially some other cases where we are currently blocking UI on slow external APIs (e.g. copying Google Sheets tabs when we delete a puzzle) and some miscellaneous functions that are currently slow and blocking (e.g. the `configureOrganizeGoogleDrive` method)

(For the record, I am aware that I am basically reinventing Monster. But Monster was good!)


imports/server/jobs/framework/jobWorker.ts
```diff
@@ -0,0 +1,185 @@
...
+      },
+      $unset: { claimedBy: "", claimedAt: "" },
+    });
+    return;
+  }
+
+  const runHandler = () => entry.handler.run(job.args, signal);
+  try {
+    await (job.createdBy
+      ? DDP._CurrentInvocation.withValue({ userId: job.createdBy }, runHandler)
```
zarvox: Ah, propagating the user context forward from whatever triggered the job.  I like it.

PR #2776, opened 2026-02-11
Add multi-email support for user accounts

ebroder:
## Motivation

This is groundwork toward #296 (merging accounts). The issue describes a common problem: users sign up multiple times with different email addresses, creating duplicate accounts. To merge those accounts, we need the ability to associate multiple emails with a single user — which Meteor supports natively but we've never used.

This PR adds the infrastructure for managing multiple emails on a single account: adding, removing, verifying, and reordering (make primary). It doesn't yet implement the actual account merge flow, but once this lands, merging two accounts becomes: move emails from account B to account A, then reassign (or tombstone) account B's historical data.

## What's here

1. **`primaryEmail()` utility** — replaces all `emails[0]` accesses with a shared helper, so the rest of the codebase doesn't need to know about the emails array structure.

2. **Server methods** — `addUserAccountEmail`, `removeUserAccountEmail`, `makeUserEmailPrimary`, `sendUserVerificationEmail`. The remove method uses an atomic MongoDB update to enforce preconditions (not primary, not last email) without races. The make-primary method uses optimistic concurrency control to detect concurrent modifications.

3. **Email verification infrastructure** — Configures `Accounts.urls.verifyEmail` to generate proper verification URLs, adds a `/verify-email/:token` route and `VerifyEmailPage` component.

4. **OwnProfilePage UI** — Replaces the single disabled email field with a full email management section (add, remove, make primary, resend verification).

5. **Publication privacy** — Other users only see your primary email via `allProfiles`, `huntProfiles`, and `profile` publications. Your full email list is only visible to yourself.

6. **Acceptance tests** — Tests for add, remove, make-primary, error paths (unverified primary, remove primary), and publication privacy.

## Review notes

- The `removeUserAccountEmail` method does a single atomic `updateAsync` with all preconditions in the query filter, then re-reads to produce a specific error message only on failure. This avoids TOCTOU races between concurrent callers.
- `makeUserEmailPrimary` takes a different approach (read-then-write with optimistic concurrency) because reordering the array can't easily be expressed as a single atomic MongoDB operation.
- `addUserAccountEmail` automatically sends a verification email. The UI also has a "resend" button.
- The `filterToPrimaryEmail` transform in publications slices `emails` to just index 0 for other users. The logged-in user's own document is not filtered (Meteor always publishes the full user doc to the owning client via `autopublish` / the login system).

### DIFF
--- a/imports/server/checkVerificationEmailCooldown.ts
+++ b/imports/server/checkVerificationEmailCooldown.ts
@@ -1,0 +1,32 @@
+import { Meteor } from "meteor/meteor";
+import MeteorUsers from "../lib/models/MeteorUsers";
+
+const COOLDOWN_MS = 60_000;
+
+export default async function checkVerificationEmailCooldown(
+  userId: string,
+): Promise<void> {
+  const user = await MeteorUsers.findOneAsync(userId, {
+    fields: { "services.email.verificationTokens": 1 },
+  });
+
+  const tokens = user?.services?.email?.verificationTokens;
+  if (!tokens || tokens.length === 0) {
+    return;
+  }
+
+  let mostRecent = 0;
+  for (const token of tokens) {
+    const when = token.when.getTime();
+    if (when > mostRecent) {
+      mostRecent = when;
+    }
+  }
+
+  if (Date.now() - mostRecent < COOLDOWN_MS) {
+    throw new Meteor.Error(
+      429,
+      "Please wait before requesting another verification email",
+    );
+  }
+}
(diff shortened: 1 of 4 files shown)


imports/client/components/VerifyEmailPage.tsx
```diff
@@ -0,0 +1,42 @@
...
+      } else {
+        navigate("/", { replace: true });
+      }
+    });
+  }, [token, navigate]);
+
+  if (error) {
+    return (
+      <Container>
+        <Alert variant="danger">Email verification failed: {error}</Alert>
```
zarvox: Translation for the message?

zarvox (approved):

PR #2779, opened 2026-02-12
Prevent Google Sheets from stealing focus on load

### LINKED ISSUE #2283: Focus issues when trying to add solution to unopened puzzle
Occasionally, we may need to open a puzzle and immediately mark it as solved (imagine we just solved it without first adding it to the sheet, or weren't actively editing it but now have an answer to submit, e.g. because we had to wait for HQ to evaluate a creative submission).

The natural way to do this is to open the puzzle and immediately click Answer. But what commonly happens is that you do this, and then Sheets steals focus from the dialog, so you start typing the answer into the sheet itself, which is unexpected given that the dialog should be a modal over the whole page. (Presumably this might also affect other dialogs like the "Edit" one, but it was less critical there in our experience).

One thought would be to see if there's a way we can force holding focus or reclaim it if taken by Docs/Sheets. I wonder if an onBlur listener on the dialog/text field that checks if focus has moved to somewhere within the Sheet iframe and, if so, refocuses the dialog/text field might work without causing too much disruption?

Another thought could be to provide a way to submit answers from the puzzle list view.

ebroder:
When a spreadsheet iframe is embedded, Google Sheets aggressively grabs focus during initialization, interrupting chat input or other interactions. Add a focus guard overlay that polls activeElement and restores focus while the guard is up. The user clicks the overlay to dismiss it and allow normal sheet interaction.

We use requestAnimationFrame polling rather than window blur/focusout events because browsers do not reliably fire blur on the parent window when focus moves into a cross-origin iframe via script.

Key the DocumentDisplay on document._id so the guard resets on puzzle navigation.

Fixes #2283

### DIFF
--- a/imports/client/components/DocumentDisplay.tsx
+++ b/imports/client/components/DocumentDisplay.tsx
@@ -3,8 +3,19 @@ import type { IconDefinition } from "@fortawesome/fontawesome-svg-core";
 import { faFileAlt } from "@fortawesome/free-solid-svg-icons/faFileAlt";
 import { faTable } from "@fortawesome/free-solid-svg-icons/faTable";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { useCallback, useEffect, useRef, useState } from "react";
 import styled from "styled-components";
 import type { DocumentType } from "../../lib/models/Documents";
+
+// Google Sheets steals focus during initialization. The number of times
+// depends on the browser engine: Chromium does it twice, Firefox once, and
+// Safari not at all (it blocks cross-origin iframe focus stealing).
+const EXPECTED_FOCUS_STEALS = (() => {
+  const ua = navigator.userAgent;
+  if (/Safari\//.test(ua) && !/Chrome\//.test(ua)) return 0;
+  if ("userAgentData" in navigator) return 2; // Chromium-based
+  return 1; // Firefox and others
+})();
 
 interface DocumentDisplayProps {
   document: DocumentType;
@@ -34,6 +45,27 @@ const StyledIframe = styled.iframe`
   background-color: #f1f3f4;
 `;
 
+const FocusGuard = styled.div`
+  position: absolute;
+  inset: 0;
+  z-index: 1;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background: rgb(0 0 0 / 15%);
+  cursor: pointer;
+  font-size: 1.1rem;
+  font-weight: 500;
+  user-select: none;
+`;
+
+const FocusGuardLabel = styled.span`
+  background: rgb(255 255 255 / 85%);
+  padding: 0.4em 0.8em;
+  border-radius: 0.3em;
+  color: #333;
+`;
+
 export const DocumentMessage = styled.span`
   display: block;
   width: 100%;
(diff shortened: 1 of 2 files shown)


zarvox:
I feel pretty hesitant about this approach.

This is a neat technique, but I'm a bit concerned about widespread use of `requestAnimationFrame` worsening power consumption by never allowing the processor to enter lower power states for very long (especially with ProMotion 120Hz displays), which can have a deleterious effect on battery life.

I feel more okay spending the core time/power when we have something that is more meaningful to show the user, like audio levels or an animation that is user-visible.

I also don't love adding an additional click to the user flow here, though I would probably feel slightly better if it were contributing some more meaningful signal to the team (e.g. the "I intend to work on this" vs "I'm just observing for now")

This broader problem (iframes stealing focus from parent frames) is not unknown to the broader browser software community; the [`focus-without-user-activation` permissions policy](https://github.com/w3c/webappsec-permissions-policy/blob/main/policies/focus-without-user-activation.md) looks like it would possibly work for us?  Could you try experimenting with that approach to see if it looks fruitful?

ebroder:
Oh just kidding, `document.featurePolicy` is a Chrome-ism, so that's no good

zarvox:
I think I could be happy with a polling approach with either `requestAnimationFrame` or a timer setup that avoids doing work while `document.visibilityState === "hidden"` (probably using https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event ) as long as we avoid polling indefinitely (so perhaps if we notice that Sheets will only steal focus up to N times, or if we have a way to detect that the Sheet has fully loaded and will no longer be aggressively stealing focus again, or a reasonable timeout later, automatically dismiss the focus guard).  I think that would address my concerns around power usage, and also avoid adding a click before people can reliably interact with the sheet.  I'm curious if we can better characterize the Sheets behavior -- does it steal focus once?  several times?  Does that sound like a feasible approach?

I could also be happy with waiting for `focus-without-user-activation` to become more reliably available, though I'm a little unsure if the implementation will wind up doing exactly what we want.

zarvox:
I did some experimentation with this, logging what actions we appear to wind up taking.  On macOS:

* In Chrome: Sheets claims focus twice, and we grab it back
* In Firefox: Sheets claims focus once, and we grab it back
* In Safari: Safari refuses to allow Sheets to steal focus, so this isn't a problem to begin with

ebroder:
Yeah, I'm happy with this approach. I'll update - I think the most resilient implementation stops listening for focus changes (a) after we've observed any focus steals (b) we've observed no more steals for X period of time.

ebroder:
OK. After playing with that implementation, I ended up going a slightly different direction:

- I stuck with rAF. When I used `setInterval`, it didn't react fast enough. (My test method: click on the chat input immediately after the page loads and start mashing the keyboard - maybe not entirely fair, but I *can* type pretty fast).
- I found the auto-clear logic annoyingly inconsistent, so I ended up encoding the expected behavior of different browsers so that they *should* clear immediately after Sheets finishes doing its thing. I also added a tab visibility handler to check focus - if (a) we are still focus-guarding (b) we become visible (c) Sheets stole our focus, steal it back and assume it's not going to do it again
- There's some time-based fallback logic in case any of this leaves the focus guard up for too long
- And the user can always click through

PR #2781, opened 2026-02-12
Cost optimization: ARM/Graviton, spot instances, haproxy consolidation

### LINKED ISSUE #2117: Optimize AWS costs, focused on lower-traffic operation

ebroder:
## Summary

Fixes #2117.

- **ARM/Graviton support**: Add a Lambda-backed custom resource that detects CPU architecture from `AppInstanceType` and threads it through AMI lookup, Docker apt source, and CloudWatch agent URL. Changing to a Graviton instance type (e.g. `t4g.micro`) automatically picks up `arm64` everywhere.
- **Spot instance support**: Add `UseSpotInstances` parameter (default `No`) that switches the ASG to use spot instances via `MixedInstancesPolicy`.
- **Consolidate haproxy and nginx**: Replace the separate haproxy (TCP multiplexer) + nginx (HTTP reverse proxy) with a single haproxy config using `switch-mode http`. Eliminates the nginx container entirely. Proxy-level asset caching is intentionally not carried over — haproxy's in-memory cache doesn't work well with Meteor's `Vary: User-Agent` on JS bundles, and browser caching (`cache-control: public, max-age=31536000` on content-hashed URLs) handles repeat visits.

## Testing

Deployed to prod (`deathandmayhem.com`) in SingleInstance mode and verified:

| Test | Method | Result |
|------|--------|--------|
| HTTPS proxying | `curl https://deathandmayhem.com/` | 200, content served |
| HTTP→HTTPS redirect | `curl http://deathandmayhem.com/` | 301 to https:// |
| Healthcheck | `curl /healthcheck` | 200 OK |
| gzip compression | `curl --compressed` | `content-encoding: gzip` |
| HSTS header | Response headers | `strict-transport-security: max-age=31536000` |
| WebSocket/DDP | WebSocket upgrade over HTTP/1.1 | 101 Switching Protocols |
| SockJS fallback | POST to `/sockjs/000/test/xhr` | 200 |
| STUN/TURN routing | `turnutils_uclient -S -t -p 443` | TLS handshake succeeds, reaches coturn |
| 503 error page | Stopped app container on instance | Custom "Temporarily Down" page served |
| CloudFormation validation | `aws cloudformation validate-template` | Passes |

Not externally verifiable: httpoxy mitigation (Proxy header stripping), X-Forwarded-For propagation.

### DIFF
--- a/scripts/run_jolly_roger.sh
+++ b/scripts/run_jolly_roger.sh
@@ -10,10 +10,12 @@ if [ -z "${HTTP_FORWARDED_COUNT+set}" ]; then
 fi
 
 if [ -z "${CLUSTER_WORKERS_COUNT+set}" ]; then
-    # If we have less than 500M of memory, we don't have enough to run more
-    # than 1 worker
+    # Each Node.js worker process has ~150MB of fixed overhead (V8 engine,
+    # compiled code, native modules) on top of the ~130MB JS heap. Only
+    # enable clustering on instances with enough memory to absorb the cost
+    # of multiple processes (main + workers + mediasoup workers per process).
     MEMORY_KB="$(awk '$1=="MemTotal:" {print $2}' /proc/meminfo)"
-    if [ "$MEMORY_KB" -gt 512000 ]; then
+    if [ "$MEMORY_KB" -gt 1536000 ]; then
         export CLUSTER_WORKERS_COUNT=auto
     fi
 fi


ebroder:
Also now includes configuration to proactively replace spot instances using ["capacity rebalancing"](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-capacity-rebalancing.html), which will spawn new instances of AWS assesses an "elevated risk" of interruption.

In single-instance mode, we previously attached the EIP when the instance was first launched. Enabling capacity rebalancing meant that we would take down JR unnecessarily whenever AWS made this assessment until the new instance came up, so instead we move the EIP attachment into the cloud-init script. When an instance finishes booting, if we're in single-instance mode, it steals the EIP from whoever was previously holding it.

ebroder:
(As per usual, I deployed this to prod to validate that it works)

zarvox (approved):
As usual, I only pretend to actually understand cloudformation, but it all sounds plausible to me.

PR #2782, opened 2026-02-12
Warn users when third-party cookie blocking causes anonymous animals

### LINKED ISSUE #1987: Nudge users with 3rd-party cookies blocked to allow them, if reasonable
User identities are not revealed in iframed sheets even when the containing folder is shared to the user when third-party cookies are blocked by the browser.  This results in more anonymous animals than ideal.

This gets tripped up by Safari in its default configuration as well as by some ad-blockers or privacy extensions for other browsers.

It would be nice if we could detect these circumstances (user has a linked Google identity but 3rd-party cookies are blocked) and nudge the user to take actions that would allow other hunters to be able to tell who a particular cursor is.

ebroder:
## Summary

- Detects when browsers block third-party cookies (Firefox TCP, Safari ITP) that cause users to appear as "Anonymous Animals" in embedded Google Sheets, even when they've linked their Google account
- Shows a dismissible toast notification with browser-specific instructions for disabling tracking protection
- Uses the existing Apps Script web app infrastructure: a new `doGet` handler serves a cookie detection page in a hidden iframe, which runs a three-step check (`hasStorageAccess()`, cookie set/read test, `requestStorageAccess()` without gesture) and posts the result back via `postMessage`

## Test plan

- [ ] In Firefox (with Total Cookie Protection on), verify the "Third-party cookies blocked" toast appears with Firefox-specific instructions
- [ ] In Safari (with ITP on), verify the toast appears with Safari-specific instructions
- [ ] In Chrome, verify the toast does NOT appear
- [ ] Dismiss the notification — verify it does not reappear during SPA navigation
- [ ] Reload the page — verify the check re-runs (so it picks up settings changes)
- [ ] Disable tracking protection in Firefox/Safari, reload — verify the toast does not appear
- [ ] Verify the toast only appears for users with a linked Google account
- [ ] Verify admin users still see the "Update Google Script" notification when the script is out of date

Closes #1987


ebroder:
(I have already gone through these tests but let Claude go ahead and enumerate the things it suggested trying)

imports/client/components/NotificationCenter.tsx
```diff
@@ -1092,9 +1174,21 @@ const NotificationCenter = () => {
...
 
   return (
-    <StyledToastContainer position="bottom-end" className="p-3 position-fixed">
-      {messages}
-    </StyledToastContainer>
+    <>
+      {cookieCheckEndpointUrl && cookieCheckResult === undefined && (
+        <iframe
+          src={cookieCheckEndpointUrl}
+          style={{ display: "none" }}
```
zarvox: I'm aware of some interventions that browsers take to limit resource usage of `iframe`s with `display: none` (things like never triggering requestAnimationFrame, suspending JS, or delaying loading the contents).  I'm guessing that none of those were an issue for you as you tested this out, which is great (I don't have great ideas for better places to put this anyway), just wanted to flag that this is a known area where browsers may some day quietly decide that they're not going to load the `src` and so we might quietly never see a `postMessage`'d result.

zarvox (approved):

PR #2783, opened 2026-02-12
Implement hunt purging

zarvox:
Some teams like to set up a new Hunt and invite users and let them poke around at the Jolly Roger UI to familiarize themselves with features, but then would like to wipe any data generated in that hunt.

This change adds:

* a new async job purgeHunt, which deletes nearly everything within a Hunt, including Google documents and files on S3 associated with that hunt, but not the Hunt object itself (nor its membership, roles, or InvitationCodes).
* indexes to collections that the purgeHunt job will query
* a new TypedMethod purgeHunt which queues an instance of that job
* a new button & modal on the HuntListPage for admins to be able to call the purgeHunt method

The general strategy is: delete Google documents from Google, delete S3 files from S3, and then for every collection that has a foreign key on Hunt, except InvitationCodes, remove everything from that collection.

To avoid the risk of new collections with hunt foreign keys being added to the codebase but not to the list of collections to purge in purgeHunt, we exhaustively list collections and (at server startup, once all Models should have been imported) verify that we have classified all collections as "to be purged" or "explicitly ignored".

In the fullness of time, it would be nice to only need to classify collections that have a hunt foreignKey, but 1) the code to do that would be a little annoying since it'd have to walk all the schemas, and 2) that code would be tied to zod 3, which we're trying to upgrade away from.  So for now we'll just make ourselves classify every Model.

### DIFF
--- a/imports/server/jobs/purgeHunt.ts
+++ b/imports/server/jobs/purgeHunt.ts
@@ -1,0 +1,287 @@
+import { setTimeout } from "node:timers/promises";
+import { Meteor } from "meteor/meteor";
+import {
+  DeleteObjectsCommand,
+  paginateListObjectsV2,
+  S3Client,
+} from "@aws-sdk/client-s3";
+import { z } from "zod";
+import Flags from "../../Flags";
+import Logger from "../../Logger";
+import Announcements from "../../lib/models/Announcements";
+import APIKeys from "../../lib/models/APIKeys";
+import BlobMappings from "../../lib/models/BlobMappings";
+import BookmarkNotifications from "../../lib/models/BookmarkNotifications";
+import Bookmarks from "../../lib/models/Bookmarks";
+import ChatMessages from "../../lib/models/ChatMessages";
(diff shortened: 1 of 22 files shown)


imports/server/jobs/purgeHunt.ts
```diff
@@ -0,0 +1,275 @@
...
+      let itemsCompleted = 0;
+
+      // Delete any Documents from the Google Drive
+      if (await Flags.activeAsync("disable.google")) {
+        Logger.info(
+          "Skipping Document deletion because Google is disabled by feature flag",
+          { huntId: args.huntId },
+        );
+      } else {
+        const docs = await Documents.find({ hunt: args.huntId }).fetchAsync();
```
ebroder: We don't soft-delete documents currently, but if we did this would miss them; should probably be `findAllowingDeleted`.

Alternatively, we could plausibly look up the HuntFolder and just delete everything in the folder? I'm not entirely sure if that's better or worse. I guess probably worse since we don't know what's been stuffed in the folder.

PR #2784, opened 2026-02-13
Cache Let's Encrypt certs in S3 to avoid rate limits

ebroder:
In SingleInstance mode, every instance launch previously requested a new certificate from Let's Encrypt, which could hit their rate limit of 5 duplicate certs per domain per week during frequent instance churn (spot replacements, deployments, etc.).

Now the renew-certificate script first tries to restore a cached cert from a dedicated private S3 bucket. Certbot is only invoked when no cached cert exists or the existing one is within 30 days of expiry. After renewal, the new cert is uploaded back to S3 for future instances.

cc @jpd236 

Because I've been locked out while testing the single instance changes earlier today, I'm not testing this yet, so also not asking for a formal review/merge yet. I'll test it out this weekend once the rate limit resets.

### DIFF
--- a/cloudformation/jolly-roger.yaml
+++ b/cloudformation/jolly-roger.yaml
@@ -379,6 +379,15 @@ Resources:
             Action: ["s3:GetObject"]
             Resource: !Sub "arn:aws:s3:::${ImageUploadBucket}/*"
 
+  CertificateStorageBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      PublicAccessBlockConfiguration:
+        BlockPublicAcls: true
+        BlockPublicPolicy: true
+        IgnorePublicAcls: true
+        RestrictPublicBuckets: true
+
   AppInstanceRole:
     Type: AWS::IAM::Role
     Properties:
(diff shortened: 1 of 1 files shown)


cloudformation/jolly-roger.yaml
```diff
@@ -647,19 +665,34 @@ Resources:
...
-                    else
-                      email_flag="--register-unsafely-without-email"
+                    CERT_BUCKET="${CertBucket}"
+                    CERT_PEM="/etc/haproxy/certs/cert.pem"
+
+                    mkdir -p /etc/haproxy/certs
+
+                    need_certbot=true
+
+                    # Try to restore cached certificate from S3
```
jpd236: This script is also configured to run daily via cron; I can understand restoring the certificate for the first run, but do we want this to happen for every daily run, or should it only be written? I suppose it's just an unnecessary copy at worst.

cloudformation/jolly-roger.yaml
```diff
@@ -647,19 +665,34 @@ Resources:
...
+                    if aws s3 cp "s3://$CERT_BUCKET/cert.pem" "$CERT_PEM" 2>/dev/null && \
+                       openssl x509 -in "$CERT_PEM" -checkend 2592000 -noout 2>/dev/null; then
```
jpd236: I wonder if we should have this check or just run certbot and let it decide what needs renewal. I believe the command will skip renewals that it doesn't yet deem necessary.

jpd236 (approved):
looks functional to me; just a couple of behavioral nits to consider:

zarvox (approved):

PR #2806, opened 2026-02-15
Better third-party cookie blocking detection

ebroder:
Previously, the cookie detection page used a three-step client-side algorithm (hasStorageAccess, document.cookie, requestStorageAccess) to detect when browsers block third-party cookies. However, privacy extensions like Privacy Badger strip Cookie and Set-Cookie headers at the HTTP network level without affecting JavaScript cookie APIs, so the client-side checks reported cookies as working when they weren't.

Instead, the Apps Script doGet() handler now calls Session.getTemporaryActiveUserKey() server-side and embeds the result into the page. This API returns a non-empty key when Google can identify the user from the HTTP request - which requires authentication cookies to be present. If empty, something is preventing cookies from reaching the server, whether that's browser-level tracking protection (Firefox ETP, Safari ITP, Chrome settings) or an extension stripping HTTP headers. This catches all known blocking mechanisms with a single check.

Also adds a 10-second timeout on the postMessage response so that if the iframe is blocked entirely (e.g. by an extension or network issue), we still warn the user rather than waiting indefinitely. Updates the warning tips to mention both browser settings and extensions for all browsers.

Fixes #2782.

### DIFF
--- a/imports/client/components/NotificationCenter.tsx
+++ b/imports/client/components/NotificationCenter.tsx
@@ -784,6 +784,14 @@ const useCookieCheck = (
   useEffect(() => {
     if (!endpointUrl || result !== undefined) return undefined;
 
+    let timeout: ReturnType<typeof setTimeout> | undefined;
+
+    function resolve(ok: boolean) {
+      if (timeout) clearTimeout(timeout);
+      cookieCheckCache = ok;
+      setResult(ok);
+    }
+
     const onMessage = (event: MessageEvent) => {
       const data = event.data as any;
       if (
@@ -791,12 +799,21 @@ const useCookieCheck = (
         data.type === "jr-cookie-check" &&
         typeof data.ok === "boolean"
       ) {
-        cookieCheckCache = data.ok;
-        setResult(data.ok);
+        resolve(data.ok);
       }
     };
     window.addEventListener("message", onMessage);
-    return () => window.removeEventListener("message", onMessage);
+
+    // If the iframe never responds (e.g. an extension blocks the request
+    // entirely, or the page gets redirected to a login wall), treat it as
+    // cookies blocked — if the test iframe can't load, the Sheets embed
+    // won't work either.
+    timeout = setTimeout(() => resolve(false), 10000);
+
+    return () => {
+      window.removeEventListener("message", onMessage);
+      clearTimeout(timeout);
+    };
   }, [endpointUrl, result]);
 
   if (!endpointUrl || result !== undefined) return result;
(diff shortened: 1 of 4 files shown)


ebroder:
Oh and for the record, I did run through what I thought was a fairly thorough set of test cases:

  Chrome:
  - Removed all ad blocker extensions. Blocked third-party cookies at browser level. Result: Warning
  - Allowed third-party cookies. Result: No warning
  - Block scripts.google.com in Chrome Dev Tools. Result: Warning
  - Enabled Privacy Badger, uBlock Origin Lite, and Adblock Plus. Result: Warning
  - Whitelist site in Privacy Badger. Result: No warning (uBlock and Adblock Plus aren't blocking cookies)
  - Incognito (No extensions). Result: Warning [Slightly surprising but Incognito unconditionally blocks 3P cookies)

  Firefox:
  - Removed all ad blocker extensions and left default ETP setting. Result: Warning
  - Disabled ETP. Result: No warning
  - Enabled Privacy Badger, uBlock Origin, and Adblock Plus. Result: Warning
  - Whitelist site in Privacy Badger. Result: No warning (uBlock and Adblock Plus aren't blocking cookies)

  Safari:
  - Default settings. Result: Warning
  - Uncheck "Prevent cross-site tracking". Result: No warning

imports/client/components/NotificationCenter.tsx
```diff
@@ -784,19 +784,36 @@ const useCookieCheck = (
...
       }
     };
     window.addEventListener("message", onMessage);
-    return () => window.removeEventListener("message", onMessage);
+
+    // If the iframe never responds (e.g. an extension blocks the request
+    // entirely, or the page gets redirected to a login wall), treat it as
+    // cookies blocked — if the test iframe can't load, the Sheets embed
+    // won't work either.
+    timeout = setTimeout(() => resolve(false), 10000);
```
zarvox: Just to avoid odd network conditions making this pop up when it shouldn't: consider if we should make this fire only after the iframe emits a successful `onload` (or unsuccessful `onerror`) event?  Perhaps we should immediately fail if the iframe triggers an onerror, and fire this setTimeout only after seeing a successful `onload`?

zarvox (approved):
I think this is basically fine -- thanks for adding i18n for the messages!

I had one idea around monitoring the iframe's load/error state that might be worth adding; feel free to give that a shot or if you think it's not worth the effort, just merge.

imports/client/components/NotificationCenter.tsx
ebroder: Hmm. I think part of the point of the timeout is hedging against an ad blocker that blocks hidden iframes entirely, in which case I'm not sure any iframe hook would fire

zarvox: Yeah okay that sounds reasonable

PR #2807, opened 2026-02-15
Add a link to the README video on mobile

ebroder:
GitHub will display video embeds on mobile if the video is a user attachment, but it gets scrubbed on mobile, so add a link to the video so that mobile users can still see it.

(Quick followup to #2752 based on @jimsug's observations)


zarvox:
While we're here: for some reason, the existing video link on desktop currently 404s.  I wonder if GitHub thought that the upload never got recorded anywhere durable (like a comment) and GC'd it from storage?

zarvox:
I guess to be clear: the link that you are adding, which is the same link, also 404s

zarvox (approved):
That works great, thanks!

PR #2808, opened 2026-02-15
Unbundle and rework permissions

zarvox:
This PR:

* adds a new role, "hunt_owner", which is assigned to the user who creates a Hunt
* adds a flexible "custom permissions" field to Hunts which can 
* restructures permission checks in permission_stubs.ts to use either custom permissions from the Hunt object (if provided) or default permissions when doing permission checks
* migrates the existing `openInvites` option to the new custom permissions scheme 
* adds UI for editing custom permissions to `HuntEditPage`

This is still work in progress; there's a few things that I think would be good to do:

- [ ] Review frontend strings that mention "operator" as the source of privilege -- under the unbundled permissions model, you might still have a power without being an operator
- [ ] Review the ability of non-operators to still hide privileged UI
- [ ] Make it possible to create/revoke the `hunt_owner` role (but make sure you can't remove the permissions from yourself)


zarvox:
A couple screenshots:

The fixture hunt's default state:

Screenshot 2026-02-15 at 20 20 54

Default permissions for a new hunt:

Screenshot 2026-02-15 at 20 21 02

imports/lib/permission_stubs.ts
```diff
@@ -6,11 +6,90 @@ import MeteorUsers from "./models/MeteorUsers";
 import type { Selector } from "./models/Model";
 import type { User } from "./models/User";
 
-function isOperatorForHunt(
+// Explicitly defined roles
+type DefinedRole = "hunt_owner" | "operator";
```
ebroder: I sort of understand why you did this, but both of the roles are already hunt-scoped, so it's a little weird to only have the "hunt_" prefix in one of them.

zarvox: Hmmm, maaaaaaaybe?

I was intending `hunt_owner` to be delegable whereas "captain" feels like it's a little bit more singleton-y and also sets an expectation that it's the person who registered your team for MH?  I feel like in practice, you and I and maybe James would be `hunt_owner` and everybody else would be `operator`, so it feels slightly odd to me to call us captains, even if we do in some ways serve that function (contact, comms, etc.).

I feel like it's usually better not to borrow a term of art if it has a specific meaning that doesn't align with how you intend it to be used.  But also I could be missing the forest for the trees here?

"leader"?  IDK

imports/lib/models/Hunts.ts
```diff
@@ -12,22 +12,42 @@ export const SavedDiscordObjectFields = z.object({
...
+  inviteUsers: RequiredPermission, // Invite users one at a time
+  bulkInviteUsers: RequiredPermission, // Invite users in batch
+  manageOperators: RequiredPermission, // Promote or demote operators
+  manageInvitationLink: RequiredPermission, // Create, roll, or deactivate invitation link
+  editPuzzles: RequiredPermission, // Add or edit puzzles
+  deletePuzzles: RequiredPermission, // Destroy or undestroy puzzles
+  operateGuessQueue: RequiredPermission, // Update guess states after submission when guess queue is enabled
+  sendAnnouncements: RequiredPermission,
+  purgeHunt: RequiredPermission, // Purge all contents of hunt
+});
```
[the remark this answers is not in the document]
zarvox: I figured out how to make enums work as keys in our existing json schema generator so I'll roll with that.

imports/lib/permission_stubs.ts
```diff
@@ -1,16 +1,142 @@
...
 import type { User } from "./models/User";
+import type {
+  ConfigurableAction,
+  ConfiguredPermissionLevel,
+  DefinedRole,
+} from "./permissions";
 import type { Selector } from "./typedModel/Model";
 
-function isOperatorForHunt(
+type EffectivePermissionLevel = "admin" | ConfiguredPermissionLevel | "none";
```
ebroder, line 15: EffectivePermissionLevel is already defined in `./permissions`, right?

imports/client/components/GuessQueuePage.tsx
```diff
@@ -373,6 +395,10 @@ const GuessQueuePage = () => {
 
   useBreadcrumb({ title: pageTitle, path: `/hunts/${huntId}/guesses` });
 
+  const canOperateQueue = useTracker(() => {
+    return userHasPermissionForAction(Meteor.user(), hunt, "operateGuessQueue");
+  }, [hunt]);
+
```
ebroder, line 401: Do we need both this and `canEdit` below?

imports/client/components/HuntEditPage.tsx
```diff
@@ -250,6 +597,21 @@ const HuntEditPage = () => {
     hunt?.memberDiscordRole,
   );
 
+  const onSetCustomPermissionsEnabled = useCallback((enabled: boolean) => {
+    setCustomPermissions(enabled ? DEFAULT_PERMISSION_LEVELS : undefined);
```
ebroder, line 601: Hmm. This has the side-effect of destroying the old custom permission settings when you toggle it, which seems a little surprising.

imports/client/components/NotificationCenter.tsx
```diff
@@ -1052,6 +1041,22 @@ const NotificationCenter = () => {
     return roles;
   }, [chatNotifications]);
 
+  // Notify backend if we are an active operator, for the purposes of role @-mention support.
+  const activeOperatorHuntIds = useTracker(() => {
+    const allHunts = Hunts.find({}).fetch();
+    const operatorHunts = huntsUserIsOperatorFor(Meteor.user());
```
ebroder, line 1047: This is just `huntsUserMayOperateGuessQueueFor`, right?

imports/server/publications/pendingGuessesForSelf.ts
```diff
@@ -48,68 +61,129 @@ definePublication(pendingGuessesForSelf, {
...
+
+    // The last known state of the user's hunts/roles fields
+    const userResidue: Pick<Meteor.User, "hunts" | "roles"> = {
+      hunts: [],
+      roles: {},
+    };
+
+    // A cache of the last-known contents of all Hunts that the user is a member of.
+    let huntsCache: Map<string, HuntType> = new Map();
+    let huntsWatch: Meteor.LiveQueryHandle | undefined = undefined;
```
ebroder, line 100: It's maybe a minor simplification, but does it make sense to just blindly cache all hunts here and trust our narrowing later on based on user state? Seems like it would get rid of one round of re-jiggering the query state, and would have better deduplication at the Meteor observer level, although every subscription would have a copy of hunts in memory.

Maybe setup a shared module-level hunt cache with an observer that gets spun up at `Meteor.startup`?

ebroder (commented):
Went ahead and did another review pass here, although I assume you're still working on it. Thanks for suffering through my UI request - it does feel a lot more pleasant to me.

Overall this still feels pretty good to my eye, and I'd be excited to see it land.

imports/lib/permission_stubs.ts
```diff
@@ -61,133 +192,42 @@ export function queryOperatorsForHunt(
...
-  if (isAdmin(user)) {
-    return true;
-  }
-
-  if (isOperatorForHunt(user, hunt)) {
-    return true;
-  }
-
-  return false;
+  return userHasRoleForHunt(user, huntId, "hunt_owner");
```
ebroder, line 203: Oh hmm. I think this needs an `isAdmin` check, or some other tweak so that admins don't get locked out.

imports/server/migrations/54-backfill-hunt-owner-roles.ts
```diff
@@ -0,0 +1,19 @@
+import Hunts from "../../lib/models/Hunts";
+import MeteorUsers from "../../lib/models/MeteorUsers";
+import { addUserToRole } from "../../lib/permission_stubs";
+import Migrations from "./Migrations";
+
+Migrations.add({
+  version: 54,
+  name: "Backfill hunt_owner role on creators of Hunts",
+  async up() {
+    for await (const hunt of Hunts.find({})) {
```
ebroder, line 10: Possible that this migration (and migration 55) should be `findAllowingDeleted`? (I think we're generally bad about doing this in migrations /shrug)

imports/client/components/Puzzle.tsx
ebroder, line 280: This feels minor, but FWIW `editButtons` attempts to treat update and delete permissions as orthogonal, but conditionalizing on `showEdit` here ties them together.

imports/client/components/PuzzleListPage.tsx
ebroder, line 381: Similarly, this probably wants to be `canDestroy`

imports/server/publications/pendingGuessesForSelf.ts
```diff
@@ -48,68 +61,129 @@ definePublication(pendingGuessesForSelf, {
...
+            ...fields,
+          } as unknown as HuntType;
+          newHuntsCache.set(_id, newObj);
+          if ("customPermissions" in fields) {
+            // trigger refresh
+            refreshObservers(permissionedHunts());
+          }
+        },
+        removed: (_id) => {
+          newHuntsCache.delete(_id);
```
ebroder, line 136: Do we need a `refreshObservers` call here and in `added`? (Assuming this mechanism stays)

imports/client/components/HuntListPage.tsx
```diff
@@ -43,7 +43,7 @@ const Hunt = React.memo(({ hunt }: { hunt: HuntType }) => {
       // Because we delete by setting the deleted flag, you only need
       // update to "remove" something
       canDestroy: userMayUpdateHunt(Meteor.user(), hunt),
```
ebroder, line 45: Hmm. Uh, this comment is kind of a lie. Only admins can delete hunts (hard enforcement in the method implementation). I guess the hunt owner maybe should be able to?

(Also applies to the "Undelete" flow in `HuntApp`)

imports/lib/permission_stubs.ts
```diff
@@ -61,133 +192,42 @@ export function queryOperatorsForHunt(
...
+  if (!user) {
     return false;
   }
 
   if (isAdmin(user)) {
     return true;
   }
 
-  if (isOperatorForHunt(user, hunt)) {
+  if (userHasRoleForHunt(user, huntId, "operator")) {
```
ebroder, line 218: I think this has a similar bug to `userMayConfigureHunt` - owners that aren't also operators don't pass this check.

ebroder (commented):
I think there are a few cases where we have some blurriness between "does this user have this role" and "does this user have the permissions of this role", which makes me think we should reevaluate the API.

Maybe we could (e.g.) rename `userHasRoleForHunt` to, like, `userHasAssignedRoleForHunt` and introduce a new `userHasEffectiveRoleForHunt` that utilizes the `effectivePermissionCoversRequiredPermission` helpers for the new permissions logic?

PR #2810, opened 2026-02-17
Add user account merging

### LINKED ISSUE #296: Merging accounts

ebroder:
I've gone through this a couple of times, and as far as I can tell, I think it's ready to go. See #296 for a writeup of the design scheme and some explanation for some of the specific structural choices. The first two commits are supporting refactors and should be independently reviewable.

Users often end up with multiple accounts under different email addresses. This adds both a self-service and admin-initiated flow for merging two users together. (In the self-service case, we use time-limited tokens to authenticate the merge)

Merging users requires careful attention to concurrency, idempotency, and failure resilience. It requires us to be especially careful about fields with unique indexes (like email addresses) — we have to remove them from the source user before we can add them to the target, and a poorly-timed failure could cause us to lose that data. We use a new MergeOperation record to manage all of these concerns:

* A unique index prevents multiple merges from operating on the same user concurrently. Creating the MergeOperation marks the start of the critical section of the merge; marking it as complete indicates the end. (And, to be clear, that critical section is...most of the merge)
* The snapshot field allows us to stash information with unique constraints *before* removing it from the source. If we find that the information is already gone from the source user, we should still have it.

Since the internal failures are the interesting things to test, this also introduces Sinon.js as a stubbing library so that we can induce a failure when the merge is mostly complete and confirm that it resumes correctly. (Normally I wouldn't bother introducing a new library just for this, but I think there are potentially interesting tests to be written around our interactions with external services.)

Fixes #296.


ebroder:
I will note that I have opted not to take any action to make these database queries performant, on the assumption that:

* These will be rare
* Our data is not *insignificant* but it is also definitely not Big Data
* If we add the indexes, we're paying permanent overhead costs for them

One argument in favor of more indexes is that if you're merging a user away, that user probably has very limited activity, so indexes would likely be *extremely* selective.

imports/client/components/MergeConfirmPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+      tokenResult.displayName && tokenResult.email
+        ? `${tokenResult.displayName} (${tokenResult.email})`
+        : (tokenResult.displayName ?? tokenResult.email ?? sourceUserId);
+
+    return (
+      <>
+        <p>
+          {t(
+            "merge.confirm",
+            "You are about to merge the account {{description}} into your current account. The other account will be deactivated and its data will be transferred to yours.",
```
zarvox: If we're going to expose this to end users, I think I would prefer that we be a little more explanatory and add just a little bit more friction.  I'm imagining wanting:

* A visual indicator of which account is going to be destroyed and which account is going to consume it; ideally with some graphical depiction (avatar, display name, email, google account, discord, and which Hunts each account is a member of).
* Maybe a summary of the data that may be transferred from the source account to the target account (e.g. 3 Hunts (Mystery Hunt 2016, Mystery Hunt 2017, Mystery Hunt 2019), 1320 Chat messages, 34 Guesses, etc.)
* Some sort of "This action cannot be undone" warning along with requiring typing some sort of exact text (like purgeHunt does) to confirm (maybe "yes, merge <source_email> into <current_email>" before taking the irreversible action.

ebroder:
Alright. I think I'd rather get this merged than have it stuck on the debate around our auth/security posture, so I'll pull the self-service mode and we can move that discussion to an issue.

PR #2829, opened 2026-02-20
Bump @biomejs/biome from 2.3.15 to 2.4.3

ebroder:
Biome 2.4 promoted several nursery rules to stable groups, so this updates the config (via `biome migrate --write`) and fixes five inline suppression comments that still referenced `lint/nursery/noUnusedExpressions` (now `lint/suspicious/noUnusedExpressions`).

Two newly promoted rules are turned off:

- noUnresolvedImports: can't resolve `meteor/...` virtual packages or react-bootstrap's `exports`-field deep imports (same situation as the existing noUndeclaredDependencies exclusion)
- noImportCycles: we have a pre-existing cycle in the hooks registry that would need architectural refactoring to untangle

Also adds `lang="en"` to cookie-test.html for the newly enforced useHtmlLang rule on HTML files.

Supersedes #2822

PR #2830, opened 2026-02-21
Switch to Oxlint (Take II)

### LINKED ISSUE #2697: Investigate oxlint as a biome replacement
I know we just switched to biome, but I recently learned about [oxlint](https://oxc.rs/docs/guide/usage/linter.html), which seems to be the even newer kid on the Rust reimplementation block.

It has some features which seem appealing:

* It seems to have even stronger performance than biome. Their benchmarks show it as roughly 2x faster, although it's unclear if they're using type-aware testing.
* It takes a significantly different approach to type-aware linting, by using `tsgo` as its backend. They believe that the result has better fidelity than biome's approach, which is to implement their own type inference engine. That seems believable to me.
* It has support for custom JS plugins. It's not clear if that already includes type-aware custom plugins (like ours), but potentially that eventually gives us a landing pad to completely eliminate eslint.

One thing I'm not wild about is their configuration strategy. They include a fairly short default list of "plugins" (i.e. rule groups), and adding plugins turns on all rules that come with those plugins. So we'll need to do some work to come up with the list of rules we actually want enabled.

But net-net, it feels like there's a win here.

### LINKED ISSUE #2785: Discord daemon doesn't handle promise rejections in event handlers
Just pushed the latest version of JR, got the following error on startup:

```
RangeError Shard 0 not found 
    /built_app/bundle/programs/server/npm/node_modules/@discordjs/ws/src/strategies/sharding/SimpleShardingStrategy.ts:72:10 SimpleShardingStrategy.send
    /built_app/bundle/programs/server/npm/node_modules/@discordjs/ws/src/ws/WebSocketManager.ts:335:24 WebSocketManager.send
    /built_app/bundle/programs/server/npm/node_modules/discord.js/src/client/websocket/WebSocketShard.js:230:22 WebSocketShard.send
    /built_app/bundle/programs/server/npm/node_modules/discord.js/src/managers/GuildMemberManager.js:297:24 
    /built_app/bundle/programs/server/npm/node_modules/discord.js/src/managers/GuildMemberManager.js:241:12 GuildMemberManager._fetchMany
    /built_app/bundle/programs/server/npm/node_modules/discord.js/src/managers/GuildMemberManager.js:213:31 GuildMemberManager.fetch
    imports/server/daemons/discord.ts:203:35 
    /built_app/bundle/programs/server/npm/node_modules/@swc/helpers/cjs/_async_to_generator.cjs:5:28 asyncGeneratorStep
    /built_app/bundle/programs/server/npm/node_modules/@swc/helpers/cjs/_async_to_generator.cjs:22:17 _next
    node:internal/process/task_queues:105:5 processTicksAndRejections
```

ebroder:
I took another stab at this because Biome makes me a little sad (e.g. not being able to understand Meteor's imports in #2829). Instead of trying to eliminate Biome entirely, I decided to split the labor. Biome continues to handle formatting (and import organization, which it treats as a separate problem but IMO is part of formatting), while Oxlint takes over linting.

This time around, it seems like the Oxlint branch is actually slightly faster. Here's Oxlint:

```
meteor npm run lint  42.56s user 6.08s system 413% cpu 11.758 total
```

vs Biome:

```
meteor npm run lint  33.79s user 6.25s system 311% cpu 12.870 total
```

although it mostly comes out in the wash. The real win here is the improvement in linting. Oxlint catches errors that Biome doesn't (e.g. unhandled promises, which I believe fixes #2785), and also has some cleanup, modernization, and stylistic rules that I find pleasing (e.g. `array[array.length - 1]` => `array.at(-1)`, or detecting unnecessary casts and non-null assertions)

I've read through the complete set of changes here, and I believe them to be solid and mostly rote. I think the only one I found that seemed particularly interesting (and worth extra testing):

* In `imports/client/components/DocumentDisplay.tsx`, changing the focus guard from a `div` to a `button`. I did test this locally and the rendering and behavior seem to be unchanged.

Finally, while I mostly find Oxlint to be a more thorough linter than Biome, there are a few things that Biome can detect that Oxlint doesn't, which we regress on here:

* Biome's [useExhaustiveDependencies](https://biomejs.dev/linter/rules/use-exhaustive-dependencies/) is stricter than Oxlint's [react/exhaustive-deps](https://oxc.rs/docs/guide/usage/linter/rules/react/exhaustive-deps) (or ESLint's [react-hooks/exhaustive-deps](https://react.dev/reference/eslint-plugin-react-hooks/lints/exhaustive-deps), which Oxlint is reimplementing). Biome detects *extra* deps, while Oxlint only detects *missing* deps.
* Oxlint has no equivalent for Biome's [useUniqueElementIds](https://biomejs.dev/linter/rules/use-unique-element-ids/)

This branch will conflict lightly with #2808, but the significant conflicts are all in  `imports/client/components/HuntEditPage.tsx` (reconciling the changes to FontAwesome icons and the suppression comments).

Fixes #2697.

### DIFF
--- a/docs/google-drive.md
+++ b/docs/google-drive.md
@@ -35,7 +35,7 @@ files:
   - imports/server/setup.ts
   - private/google-script/cookie-test.html
   - private/google-script/main.js
-updated: 2026-02-20T00:00:00Z
+updated: 2026-02-21T04:45:00Z
 ---
 
 # Google Drive Integration
--- a/imports/client/components/ChatMessage.tsx
+++ b/imports/client/components/ChatMessage.tsx
@@ -1,4 +1,3 @@
-// biome-ignore-all lint/suspicious/noArrayIndexKey: migrated from eslint
 import { decodeHTML } from "entities";
 import type { Token, Tokens } from "marked";
 import { marked } from "marked";
--- a/imports/client/components/FancyEditor.tsx
+++ b/imports/client/components/FancyEditor.tsx
@@ -169,7 +169,6 @@ const EditableMentionRenderer = ({
       name = element.roleId;
       break;
     default:
-      // biome-ignore lint/suspicious/noUnusedExpressions: exhaustive check
       element satisfies never;
   }
   const Elem = selected && focused ? SelectedMentionSpan : MentionSpan;
@@ -375,7 +374,6 @@ const MatchCandidate = ({
         </MatchCandidateRow>
       );
     default:
-      // biome-ignore lint/suspicious/noUnusedExpressions: exhaustive check
       mention satisfies never;
       return null;
   }
@@ -933,7 +931,6 @@ const FancyEditor = ({
                 break;
               }
               default:
-                // biome-ignore lint/suspicious/noUnusedExpressions: exhaustive check
                 mention satisfies never;
             }
             setCompletionAnchorRange(undefined);
@@ -1023,7 +1020,6 @@ const FancyEditor = ({
           insertRoleMention(editor, m.roleId);
           break;
         default:
-          // biome-ignore lint/suspicious/noUnusedExpressions: exhaustive check
           m satisfies never;
       }
       setCompletionAnchorRange(undefined);
--- a/imports/client/components/PuzzleAnswer.tsx
+++ b/imports/client/components/PuzzleAnswer.tsx
@@ -86,7 +86,6 @@ const PuzzleAnswer = React.memo(
         },
       );
       formattedAnswer = segments.map((segment, i) => (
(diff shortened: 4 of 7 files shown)


zarvox (commented):
Okay, with the exception of the last commit, all this looks good.

The last commit throws away a lot of comments recording intent which I'd prefer to keep around, especially because some of them are doing non-obvious things.  Should the linter someday gain the ability to lint for extra effect deps, I fear a well-meaning developer (or oblivious AI agent) will turn on the lint, see a bunch of violations, and then regress a bunch of subtle behaviors around scroll sticking to the bottom of the viewport on resize and similar because there's nothing there to remind them not to.

PR #2938, opened 2026-04-10
Take updates

ebroder:
Apparently it's "Javascript updates break our project" week.

This takes an assortment of updates that are compatible with our project with some changes:

* styled-components: Moved `@types/stylis` from dependencies to dev dependencies, but continues referencing it in its type declarations, so we need to add it to *our* dev dependencies.
* sinon: Introduced Typescript bugs in their new type declarations, so we need to pin `@sinonjs/fake-timers` until https://github.com/sinonjs/fake-timers/pull/559 lands in a release (which, in fairness, could happen pretty quickly)
* i18next: Dropped their default behavior of outputting an ad for their service via `console.log` and dropped the config parameter we were using to suppress it.
* oxlint: Made it an error to reference a non-existing rule in oxlintrc.json

oxlint additionally promoted some new rules from the nursery, enabling them by default, which required some code fixes:

* no-unnecessary-type-conversion: Identifies use of `!!` on already-boolean values
* no-unnecessary-type-arguments: Identifies explicit type parameters when Typescript can already infer the same thing (this mostly affected `useState`)
* await-thenable: Not a new rule, but was updated to detect when `Promise.race` receives a value that's not a promise.

The last of those detected a real bug where we were prematurely `await`ing a `setTimeout` in `gdriveActivityManager`.


imports/client/hooks/useCallState.ts
```diff
@@ -802,7 +802,7 @@ const useCallState = ({
   // A map from track ID to state for that track, including subscription handle,
   // mediasoup Producer, and a call-exactly-once callback function, all of
   // which need careful lifecycle handling and explicit cleanup.
-  const producerMapRef = useRef<Map<string, ProducerState>>(new Map());
```
zarvox, line 805: Having read through the preceding diff, I'm sold on eliding types in most places when it's just a string or boolean or enum or undefined, but I do find them at least slightly useful as an easy reference for things where the types are less obvious, as well as to make it easier to jump to the value type's definition.  This is a place where I think when reading I find the types helpful, even if *technically* they could be elided.

To babble for a bit about why: I think this is especially true of things like maps or hashes where the shape of the value type is not immediately obvious from its initialization, and while the typechecker can correctly infer the type, to do so it must reference both the initial declaration also its usage sites.  Requiring the reader to do that same context-gathering-through-examining-usage to understand what this object is adds cognitive burden, whereas if that information is in the declaration line, the reader can work that out in fewer passes through the code.  On the flipside, with primitives, the initial value argument communicates enough information about the type in a single line, so there's no need to collect that context from other lines in the file to understand what the shape of the data is, so eliding the types there seems harmless.

Anyway, if there's a good way to reify that sort of policy into the lint rule, that'd be lovely, but if that's not straightforward, then maybe we should just disable the lint for these cases since there aren't very many at all (there's really just this one and `consumerMapRef` below).

zarvox (approved):
Two spots I'd like to keep technically-unneeded type annotations (and some discussion on why), but otherwise LGTM, thanks!

imports/client/hooks/useCallState.ts
ebroder: Yeah I definitely missed this when I was looking through the diff, and I think the issue is actually more significant, because we've regressed the type here to `Map<any, any>`

I think the right way to phrase this is just `const producerMapRef = useRef(new Map<string, ProducerState>());`. I'll do another pass for these, and see if there's also any other stricter linting we could use to catch this.

zarvox: Thinking out loud: did any of the enum-valued types regress to `string`?  That would also be potentially a loss for autocomplete or typo detection

ebroder: No, those seem to get inferred correctly:

Screenshot 2026-04-10 at 4 41 21 PM

ebroder: (FWIW I'm also going to pursue this as a bug in Oxlint - I think it's an incorrect behavior that it flags this, since the type parameter *was* refining the type)

ebroder: Aha - looks like https://github.com/oxc-project/oxc/issues/20933

zarvox (approved):
Affirming LGTM

PR #2939, opened 2026-04-11
Update to meteor 3.4.1-beta.1

zarvox:
This is basically picking up where https://github.com/deathandmayhem/jolly-roger/pull/2678 left off.  At that point, we were blocked on upstream fixing their type errors some months ago.

This appears to pass lints and runs fine in dev mode locally; I'm interested in letting our presubmit tests determine if most actual functionality all still works as expected once bundled.

I don't have a strong opinion on whether we should actually try to merge this before an official 3.4.1 release; I just wanted to verify that this beta release did not wind up still blocked on a different problem.


ebroder:
FWIW I don't know that I feel in a rush to upgrade to 3.4, although these changes all seem fine.

I think this is OK given that the tests seem to be passing, but one thing I'll call out is that I originally structured them with a unified client/server entrypoint because the test files introduce some new server-side hole punch methods that the client-side tests exercise, and I think I believed that the client test runtime wouldn't get a separate server that had that code loaded. Seems like that all seems to be working fine, though.

zarvox:
My best understanding of tests is:

* `meteor test` without `--full-app` (which we do not use) loads files matching `*.tests?.*` or `*.specs?.*`, and then the specified testModule, all with Meteor.isTest set to true and Meteor.isAppTest set to false, and then the test driver.
* `meteor test --full-app` eagerly loads files matching `*.tests?.*` or `*.specs?.*`, the mainModule, and then the specified testModule, with Meteor.isAppTest set to true, and then the test driver

We currently use only the latter setup (running the full app, plus some additional code from our testModule, then the test driver).  It appears to be supported (now) to define separate client and server entrypoints for `testModule`, which results in the above loading behavior for each (and then the test driver running all the server tests sequentially, and then all the client tests sequentially).  If we wanted to, we could split our test code up by lib/client/server rather than using `Meteor.isClient` or `Meteor.isServer` blocks, which might allow us to do less casting around imports and more "just load what makes sense".  (Many of our test files wrap all the tests in an isClient or isServer block anyway, so it might make sense to only import that file in the appropriate context in the first place.)  I might not go as far as making whole /client and /server directories for tests, but I might consider splitting test files that are relevant in both scopes out to `*-client.ts` and `*-server.ts` and just importing the appropriate one for the entrypoint.  We'd probably be able to get rid of a bunch of the dynamic imports and casts and just write code that looks like the rest of our code.

-----

There appears to be one last loose end which is that when we run `meteor build`, Meteor is invoking `rspack` which in turn is [using a shebang with env and node](https://github.com/web-infra-dev/rspack/blob/main/packages/rspack-cli/bin/rspack.js), and since we've been not putting a `node` on `$PATH` (instead expecting meteor to use the one in its bundle), that's now failing.  Do we want to explicitly put the one from the meteor dev bundle on PATH for that `meteor build` invocation?  We can determine it with something like `dirname $(meteor node -e "console.log(process.execPath);")`.  Or we could add a stub `/usr/local/bin/node` like

```bash
#!/bin/bash
exec meteor node "$@"
```
?  Or patching the installed rspack cli's first line to be invoked through `meteor node`?

I think I lean towards the "invoke meteor build with a modified PATH" route, since it should be fast, correct, and avoid modifying the filesystem

ebroder:
Yeah that seems fine, no objections. I know that Meteor upstream has been a bit sluggish to move on issues lately, but should we file this one with them? (Or at least see if they would consider it to be a bug?)

zarvox:
Hmmm, probably?  I can understand it not being a thing they've tested much since I imagine most users have a node and an npm already, but it does seem to be the case that they still intentionally ship their own bundle as part of the meteor tool to achieve predictable behavior, so this invocation leaking out of that tested bundle/environment to the broader system environment seems like it might be considered a bug.

On the bright side, even if they don't care to fix it, it's something we can work around ourselves.

With one mildly ugly hack in the Dockerfile to work around Meteor printing a long warning when you invoke it as uid=0, this branch now passes CI.

I think I'm inclined to try to land this now?  Theoretically switching to rspack should [unblock the react-router upgrade](https://github.com/deathandmayhem/jolly-roger/pull/2561#issuecomment-3624121266).

ebroder (approved):
I mean you know I love an upgrade

PR #3000, opened 2026-05-03
Work around potential issue with multilogin and Apps Script URL.

jpd236:
In some multi-login scenarios, the published Google Apps Script app URL redirects to a bogus URL that 404s, breaking third-party cookie detection. Adding "/a/~/" to the front of the URL works around the issue.

### DIFF
--- a/docs/google-drive.md
+++ b/docs/google-drive.md
@@ -35,7 +35,7 @@ files:
   - imports/server/setup.ts
   - private/google-script/cookie-test.html
   - private/google-script/main.js
-updated: 2026-04-14T07:38:00Z
+updated: 2026-05-03T04:05:35Z
 ---
 
 # Google Drive Integration
--- a/imports/server/setup.ts
+++ b/imports/server/setup.ts
@@ -99,6 +99,33 @@ Meteor.publish("enabledChatImage", async function () {
   this.ready();
 });
 
+/**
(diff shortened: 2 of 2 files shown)


ebroder (approved):

ebroder:
@jpd236 I'm running into weird behavior here with my Google Workspace account, where I'm getting redirected to `/a/ebroder.net/a/~/macros/[...]`, which breaks.

Do you remember why we did it this way instead of using an `?authuser=${Meteor.user().googleAccount}` query param like we do for the embed?

jpd236:
Reproduced with a workspace account here too. But I don't think that URL supports such an authuser parameter. At least when I try it, it doesn't seem to fix the original issue (redirecting me to `/u/2/macros/[...]`, leaving the authuser parameter untouched). So it will work if the workspace account is the only account, but in a multilogin scenario it may fail per the original bug. The idea to use /a/~ was from https://groups.google.com/g/google-apps-script-community/c/1xrlqlkGEQ0/m/P1_SsfJbAwAJ and I don't know how they figured it out.

I don't think doing something different based on the email's domain would work either. Maybe the only option is to try both paths and see which one works?

ebroder:
Oh huh, yeah, try both does seem like a good alternative. I'll work on something.

PR #3068, opened 2026-06-07
Fix PR builds from forks (and also fix required status checks)

ebroder:
Two issues, detailed more in comments and the commit messages:

* PRs from forks were failing to build due to fallout from #2836 - the builds were trying to populate a cache that they didn't have write access to
* The "merge" job was configured as a required status check with the intention of blocking merges if the build failed, but actually it didn't do that

(Does this actually work? My best guess is yes, but we may have to wait until we see it in action to know for sure)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -139,17 +139,35 @@ jobs:
           if-no-files-found: error
           retention-days: 1
 
+  # The merge check is a gate on the rest of CI passing. For PRs and non-main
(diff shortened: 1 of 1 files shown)


zarvox (approved):
I feel like I have never really grasped the GitHub Actions model but after reading https://docs.docker.com/reference/cli/docker/buildx/build/#cache-to and eventually coming to understand that the merge step implicitly only runs today if all previous steps succeeded, but we want it to run every time (and potentially fail) so the check can be blocking, so this seems probably correct.

PR #3082, opened 2026-06-14
Oxlint bump

ebroder:
The dependabot upgrade to 1.69 failed because Oxlint has added new checks to the `suspicious` and `correctness` categories that we include by default.

A few of them we suppress:

- no-underscore-dangle: We had this enabled in the old eslint days but with a long enough list of overrides that honestly we should have just disabled it then.
- control-has-associated-label: This implementation is just buggy and overly sensitive

The rest we fix:
 - no-useless-default-assignment flagged two cases where we had destructuring default assignments that were noops (because the values were never undefined)
 - consistent-function-scoping flagged a case where we could hoist a function to top-level because it wasn't capturing anything in scope
 - no-unnecessary-type-parameters, split out into a separate commit, flagged several cases where we had generic functions that weren't actually generic (i.e. were using generics mostly as a constraint on the parameter, in a way that you can just express directly). Most of this of course is hopefully going to be dead code if/when zod v4 ever lands


zarvox (approved):
Looks good.  Probably needs a rebase (I merged a bunch of dependabot updates first without looking to see if there were human changes, sorry)

PR #3144, opened 2026-07-15
Force node-gyp to use C++17 when building Meteor's bcrypt.

jpd236:
Fixes build error when bcrypt 5.0.1 is built locally as no arm64 prebuilt is available.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -120,12 +120,15 @@ EOF
 
 # Install server dependencies
 WORKDIR /built_app/bundle/programs/server
-
-# Force node-gyp to use C++17 for legacy Meteor binary dependencies
-ENV CXXFLAGS="-std=c++17"
-ENV CFLAGS="-std=c++17"
-
-RUN --mount=type=cache,target=/root/.npm meteor npm install --omit=dev
+RUN --mount=type=cache,target=/root/.npm <<'EOF'
+#!/bin/bash
+set -eux
+set -o pipefail
+# meteor prints junk about running as root to stdout, so capture this over stderr *sigh*
+NODE_PATH=$(cd /app && meteor node -e 'console.error(process.execPath);' 2>&1 1>/dev/null)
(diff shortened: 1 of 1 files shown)


ebroder:
I think this change is right in the specifics, but I think there's a more generalized fix that I'd prefer.

This is happening because invoking the `meteor` CLI outside of a project directory causes it to use the latest Meteor release that its catalog is aware of (since there's no `.meteor/release` file). That catalog is only updated with certain actions. For our purposes, `meteor list` is the only relevant one, which means that the catalog is built into our Docker cache and not updated later. With a clean build, the final `meteor npm install` runs on Meteor 3.5 instead of the current release.

The 3.5 release bumped the version of Node, npm, and node-gyp that it uses to one that (as you observed) defaults to C++20 and can't build bcrypt 5.0.1, and from there everything else falls out as you described. And as you suggested in Discord, all of this doesn't happen on our upstream because we have an old pre-3.5 Docker cache that doesn't have 3.5 in its catalog

However, I think the better fix here is twofold:
* Close the gap on which Meteor release we're using by changing that final `meteor npm install` invocation to use the same `PATH` override that we [already use in the production build](https://github.com/deathandmayhem/jolly-roger/blob/05a941419e2fe669078e81c874931b975ac7cf7f/Dockerfile#L116-L118)
* Try to get upstream Meteor to upgrade bcrypt to v6 (which is safe) before they release v3.5.1

If you want to take the Dockerfile update, I'll take the PR for upstream.

ebroder:
(Opened https://github.com/meteor/meteor/pull/14591)

ebroder:
Also one clarification: we can't just blindly copy the lines from the build step; we need to make sure we set the right CWD when we extract the path

jpd236:
I think this is done as you meant it, PTAL. (It works but I don't know if the cache is now masking something on my fork as well).

ebroder (approved):

PR #3158, opened 2026-07-17
Fix Apps Script cookie check for Workspace accounts

ebroder:
The third-party cookie check embeds the published Apps Script web app in a hidden iframe. 2994ffd worked around a multilogin redirect bug by prefixing the path with `/a/~/`, but that prefix itself 404s for Google Workspace domain accounts (Google redirects it to `/a/<domain>/a/~/...`).

Since the server can't tell which URL form a given browser session needs, stop committing to one. `googleScriptCookieCheckUrls` now publishes the configured URL alongside both redirect variants, and the client loads each in its own hidden iframe and uses whichever reports back first. A frame only responds if it actually loaded, and cookie blocking is per-site, so the first answer is authoritative.

cc @jpd236 since this is a followup from #3000.

### DIFF
--- a/imports/client/components/NotificationCenter.tsx
+++ b/imports/client/components/NotificationCenter.tsx
@@ -781,24 +781,32 @@ const StyledToastContainer = styled(ToastContainer)`
 let cookieCheckCache: boolean | undefined;
 
 const useCookieCheck = (
-  endpointUrl: string | undefined,
-): boolean | undefined => {
+  urls: string[] | undefined,
+): { result: boolean | undefined; frames: React.ReactNode } => {
   // On mobile we show a link to the Google Doc rather than embedding it in an
   // iframe, so third-party cookie support doesn't matter.
   const isDesktop = useMediaQuery(desktopQuery);
   const [result, setResult] = useState(cookieCheckCache);
 
-  useEffect(() => {
-    if (!endpointUrl || !isDesktop || result !== undefined) return undefined;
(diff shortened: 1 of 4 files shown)


imports/client/components/NotificationCenter.tsx
```diff
@@ -807,23 +815,33 @@ const useCookieCheck = (
...
     window.addEventListener("message", onMessage);
 
-    // If the iframe never responds (e.g. an extension blocks the request
-    // entirely, or the page gets redirected to a login wall), treat it as
-    // cookies blocked — if the test iframe can't load, the Sheets embed
+    // If no frame ever responds (e.g. every candidate 404s, an extension blocks
+    // the requests, or a frame is redirected to a login wall), treat it as
+    // cookies blocked, since if the test iframes can't load, the Sheets embed
     // won't work either.
     timeout = setTimeout(() => resolve(false), 10000);
```
ebroder, line 822: Small note: because the `useEffect` is only depending on `hasUrls`, if the URL changes while this timeout has been scheduled, it won't get reset. That feels obscure enough that I don't really care, but calling it out just as a note.

jpd236:
I think we also need to try both URLs when uploading images through the app script, right?

ebroder:
No, because the client doesn't hit the script directly, it goes through the server invocation.

jpd236 (approved):

zarvox (approved):

Issue #601, opened 2022-01-16
For posterity, don't hard-delete voice call data


ebroder:
Right now, most of the mediasoup models are hard-deleted after the end of the call. This makes it difficult to collect post-hoc stats on how many calls there were.

We should see if we can switch to soft-deletes - at least for `Peer`. Note that this is made slightly tricky by the unique index on hunt/call/tab - we either need a partial index on `{deleted: false}` or we need something more like a `deletedAt`, so that each value of `deletedAt` is unique.

PR #3159, opened 2026-07-19
Record call presence, not just speech, in CallActivities

ebroder:
Previously `CallActivities` only recorded when someone crossed a volume threshold, so we couldn't see who was on a call but just not saying anything. Joining a call is useful as an explicit signal of user engagement on a puzzle, so we should capture it. Therefore, we add a `speaking` flag to `CallActivities`, migrated to `true` for all existing records. Going forward, we track call membership (with `speaking` set to `false`) when a user joins a call and with a new periodic sweep that creates new buckets as time passes. The existing audio observer becomes an upgrade from speaking=false to speaking=true, rather than the only thing that creates those records.

For now, sparklines only look at speakers, matching the current behavior, so none of this has user-visible impact, but it starts us down the road towards richer tracking of user engagement and improves the status quo re #601, even if it doesn't actually solve it.

(I did do some quick local testing and confirm that it appears to write out records with both speaking=false and speaking=true where you'd expect)

### DIFF
--- a/imports/server/mediasoup.ts
+++ b/imports/server/mediasoup.ts
@@ -40,10 +40,14 @@ import {
   cleanupDeadServer,
   registerPeriodicCleanupHook,
 } from "./garbage-collection";
-import ignoringDuplicateKeyErrors from "./ignoringDuplicateKeyErrors";
+import { isDuplicateKeyError } from "./ignoringDuplicateKeyErrors";
 import CallActivities from "./models/CallActivities";
 import onExit from "./onExit";
+import recordCallPresence from "./recordCallPresence";
 import serverId from "./serverId";
+
+// Clamp the sweep interval so we don't hammer the database too much in dev mode
+const presenceSweepInterval = Math.max(2500, ACTIVITY_GRANULARITY / 5);
 
 const mediaCodecs: types.RouterRtpCodecCapability[] = [
   {
@@ -248,6 +252,8 @@ class SFU {
 
   public monitorConnectAcksHandle?: Meteor.LiveQueryHandle;
 
+  public presenceSweepHandle?: number;
+
   private constructor(
     ips: [ListenIp, ...ListenIp[]],
     worker: types.Worker,
@@ -266,7 +272,7 @@ class SFU {
     this.onMonitorRouterCreated(this.monitorRouter);
   }
 
-  async setupDBWatches() {
+  async setup() {
     this.peersHandle = await Peers.find({}).observeChangesAsync({
       // Use this as an opportunity to cleanup data created as part of the
       // transport negotiation
@@ -420,6 +426,12 @@ class SFU {
         });
       },
     });
+
+    this.presenceSweepHandle = Meteor.setInterval(() => {
+      this.sweepCallPresence().catch((error) => {
+        Logger.error("mediasoup sweepCallPresence failed", { error });
+      });
+    }, presenceSweepInterval);
   }
 
   static async create(ips: [ListenIp, ...ListenIp[]]) {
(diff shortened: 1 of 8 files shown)


imports/server/mediasoup.ts
```diff
@@ -685,14 +723,23 @@ class SFU {
...
+          const selector = { call: observerAppData.call, user, ts };
+          try {
+            await CallActivities.upsertAsync(selector, {
+              $set: { speaking: true },
+              $setOnInsert: { hunt: observerAppData.hunt },
             });
-          });
+          } catch (e) {
+            if (!isDuplicateKeyError(e)) throw e;
+            await CallActivities.updateAsync(selector, {
```
zarvox, line 742: In what circumstance would an upsert receive a duplicate key error?

imports/server/mediasoup.ts
```diff
@@ -40,11 +40,14 @@ import {
...
   registerPeriodicCleanupHook,
 } from "./garbage-collection";
-import ignoringDuplicateKeyErrors from "./ignoringDuplicateKeyErrors";
+import { isDuplicateKeyError } from "./ignoringDuplicateKeyErrors";
 import CallActivities from "./models/CallActivities";
 import onExit from "./onExit";
+import recordCallPresence from "./recordCallPresence";
 import serverId from "./serverId";
 
+const presenceSweepInterval = ACTIVITY_GRANULARITY / 5;
```
zarvox: In dev mode, this is going to be a sweep/round of writes every second, which feels like it might be a little bit much.

imports/server/mediasoup.ts
```diff
@@ -504,6 +518,30 @@ class SFU {
     await Routers.removeAsync({ createdServer: serverId });
   }
 
+  async sweepCallPresence() {
```
zarvox, line 522: I'd think I'd kinda like it if we computed the (rounded) timestamp here once, rather than doing so in every call of `recordCallPresence`?  The presence of the peers was checked up here at this time; if the DB is slow to service the insert loop or tears it across an interval boundary, the data should probably still all agree about when the sample was taken.

imports/server/mediasoup.ts
ebroder: Hmm. I guess I'm open to relaxing it, but in virtually all cases it's just going to be a single query that returns no (or maybe one, or very rarely a handful of) results, right? It definitely wants to be at least ACTIVITY_GRANULARITY or the data is weird, and I think ideally you want it to be a little more frequent so that you don't accidentally skip over a bucket.

imports/server/mediasoup.ts
ebroder: As I understand it, upserts are basically a two-stage operation that isn't wholly atomic. It queries for a record, and depending on whether or not it finds one, runs either an update or an insert, so there's a race if an insert is happening in parallel. (That's my read of https://www.mongodb.com/docs/manual/reference/method/db.collection.update/#upsert-with-duplicate-values, at least, and I have some vague non-specific memories of slightly non-intuitive behavior around MongoDB + upserts + unique indexes)

imports/server/mediasoup.ts
ebroder: (I'd be OK with `ACTIVITY_GRANULARITY/2` or maybe `Math.max(ACTIVITY_GRANULARITY/5, 2500)`?)

zarvox: With the exception of the times I test all the different browsers interoperating, probably?  It's probably fine

imports/server/mediasoup.ts
zarvox: Ah, okay, sounds reasonable then.

imports/server/mediasoup.ts
zarvox: `Math.max(ACTIVITY_GRANULARITY/5, 2500)` sounds good to me

zarvox (approved):

PR #3160, opened 2026-07-23
Prepare for a hopeful eventual Meteor 3.5 upgrade

ebroder:
I was curious if we were actually going to be able to upgrade to Meteor 3.5.1 when it came out and the answer was...mostly. I've filed meteor/meteor#14608, which is a showstopper bug for us, but otherwise Jolly Roger seems to mostly work, although the tests fail.

Why do the tests fail? Because as of 3.5, Meteor defaults to using MongoDB change streams (yay!) but our tests reach into the internals of the oplog driver (boo!).

So here is an anticipatory fix that makes the tests backend-agnostic by using a mechanism that Meteor methods already depend on. It should work against both 3.4 and 3.5. (And for bonus points, might work against the polling backend if for some reason anyone was manually opting into running tests with that? Although I haven't tested and don't especially care - if you want to break your MongoDB you can keep both pieces)

Here's the change, for reference:

> Decouple reactivity tests from the oplog observe driver
> 
> Previously these tests fenced on `_oplogHandle.waitUntilCaughtUp()`, which (obviously) depends on observers being driven by the oplog. With Meteor 3.5, the default switches to change streams; the oplog endpoints still exist but never fire.
> 
> Instead, wrap test writes in a DDP write fence, which resolves once every interested observer has processed them. That's the same guarantee used by Meteor methods (writes that are part of a method are pushed before the method itself returns), so we can rely on all drivers to support it.
> 
> The publishJoinedQuery tests also leaked their subscriptions until the end of the suite (`after()` inside `it()` registers a suite-level hook), letting a later identical query seed itself from an earlier test's stale multiplexer cache; they now stop in `afterEach`. Their convergence loop counted `setImmediate` turns, which bounds nothing under a driver with real network latency; it now polls a wall-clock deadline.
> 
> The full suite passes under both drivers (3.4.1 oplog, release-3.5.1 change streams).

### DIFF
--- a/tests/lib/waitForAssertion.ts
+++ b/tests/lib/waitForAssertion.ts
@@ -1,0 +1,21 @@
+import { setImmediate } from "node:timers/promises";
+
+// Polls an assertion until it stops throwing. Once the wall-clock deadline
+// passes, a final attempt runs outside the try, so the assertion's own error
+// is what propagates. The default deadline stays below mocha's 2000ms test
+// timeout so that error, not mocha's generic timeout, is what reports.
+export default async function waitForAssertion(
+  assertion: () => void,
+  timeoutMs = 1500,
+): Promise<void> {
+  const deadline = Date.now() + timeoutMs;
+  while (Date.now() < deadline) {
+    try {
+      assertion();
+      return;
+    } catch {
+      await setImmediate();
+    }
+  }
+  assertion();
+}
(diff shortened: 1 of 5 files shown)


tests/lib/waitForAssertion.ts
```diff
@@ -0,0 +1,21 @@
...
+  assertion: () => void,
+  timeoutMs = 1500,
+): Promise<void> {
+  const deadline = Date.now() + timeoutMs;
+  while (Date.now() < deadline) {
+    try {
+      assertion();
+      return;
+    } catch {
+      await setImmediate();
```
zarvox, line 17: I was initially worried about this busy-looping, but I guess if its guarded by a 1500ms timeout it's probably fine

zarvox (approved):

PR #3229, opened 2026-08-23
Create a bookmarklet to add puzzles from puzzle pages

jpd236:
This is modeled off of #2392, but as a bookmarklet rather than a browser extension for a simpler installation, security, and maintenance story.

The bookmarklet is linked from the advanced section of the user's profile page, next to API keys. It launches a popup modal of a standalone Add Puzzle form, sharing most logic with the existing form. A hunt dropdown is added, and the most recently-selected hunt is saved in local storage. The current page's title and URL are prepopulated in the form, and we apply some rough heuristics to strip out redundant information from the title (which can be updated for new hunts as they arise).

In addition, for both the bookmarklet popup as well as the regular Add Puzzle form, we add easy links to add any of the ten most recently used tags for the selected hunt.

See #237

### DIFF
--- a/imports/client/components/AddPuzzlePage.tsx
+++ b/imports/client/components/AddPuzzlePage.tsx
@@ -1,8 +1,13 @@
 import { Meteor } from "meteor/meteor";
 import { useTracker } from "meteor/react-meteor-data";
-import { useCallback, useMemo, useRef } from "react";
+import { useCallback, useMemo } from "react";
 import Alert from "react-bootstrap/Alert";
+import Col from "react-bootstrap/Col";
 import Container from "react-bootstrap/Container";
+import FormGroup from "react-bootstrap/FormGroup";
+import FormLabel from "react-bootstrap/FormLabel";
+import FormSelect from "react-bootstrap/FormSelect";
+import Row from "react-bootstrap/Row";
 import { useTranslation } from "react-i18next";
 import { useNavigate, useSearchParams } from "react-router";
 import inferPuzzleTitle from "../../lib/inferPuzzleTitle";
@@ -17,10 +22,7 @@ import { useAddPuzzleLastSelectedHuntId } from "../hooks/persisted-state";
 import useDocumentTitle from "../hooks/useDocumentTitle";
 import useTypedSubscribe from "../hooks/useTypedSubscribe";
 import Loading from "./Loading";
-import type {
-  PuzzleModalFormHandle,
-  PuzzleModalFormSubmitPayload,
-} from "./PuzzleModalForm";
+import type { PuzzleModalFormSubmitPayload } from "./PuzzleModalForm";
 import PuzzleModalForm from "./PuzzleModalForm";
 
 /**
@@ -83,8 +85,6 @@ const AddPuzzlePage = () => {
     }
     return Tags.find({ hunt: selectedHuntId }).fetch();
   }, [selectedHuntId, loadingTags]);
-
-  const addModalRef = useRef<PuzzleModalFormHandle>(null);
 
   const isPopup = Boolean(window.opener);
 
@@ -166,17 +166,38 @@ const AddPuzzlePage = () => {
   return (
     <Container className="p-3">
       <PuzzleModalForm
-        ref={addModalRef}
         huntId={selectedHuntId}
         tags={huntTags}
         initialTitle={inferredTitle}
         initialUrl={urlParam}
-        hunts={writableHunts}
-        onHuntChange={onHuntChange}
         onSubmit={onSubmit}
         onHide={onHide}
         inline
-      />
+      >
+        {writableHunts.length > 1 && (
+          <FormGroup
+            as={Row}
+            className="mb-3"
+            controlId="add-puzzle-hunt-select"
+          >
+            <FormLabel column xs={3}>
+              {t("puzzle.edit.hunt", "Hunt")}
+            </FormLabel>
+            <Col xs={9}>
+              <FormSelect
+                value={selectedHuntId}
+                onChange={(e) => onHuntChange(e.currentTarget.value)}
+              >
+                {writableHunts.map((h) => (
+                  <option key={h._id} value={h._id}>
+                    {h.name}
+                  </option>
+                ))}
+              </FormSelect>
+            </Col>
+          </FormGroup>
+        )}
+      </PuzzleModalForm>
     </Container>
   );
 };
--- a/imports/client/components/AddPuzzlePage.tsx
+++ b/imports/client/components/AddPuzzlePage.tsx
@@ -1,0 +1,184 @@
+import { Meteor } from "meteor/meteor";
+import { useTracker } from "meteor/react-meteor-data";
+import { useCallback, useMemo, useRef } from "react";
+import Alert from "react-bootstrap/Alert";
+import Container from "react-bootstrap/Container";
+import { useTranslation } from "react-i18next";
+import { useNavigate, useSearchParams } from "react-router";
+import inferPuzzleTitle from "../../lib/inferPuzzleTitle";
+import Hunts from "../../lib/models/Hunts";
+import Tags from "../../lib/models/Tags";
+import { userMayWritePuzzlesForHunt } from "../../lib/permission_stubs";
+import huntsAll from "../../lib/publications/huntsAll";
+import tagsForHunt from "../../lib/publications/tagsForHunt";
+import createPuzzle from "../../methods/createPuzzle";
+import { useBreadcrumb } from "../hooks/breadcrumb";
+import { useAddPuzzleLastSelectedHuntId } from "../hooks/persisted-state";
+import useDocumentTitle from "../hooks/useDocumentTitle";
+import useTypedSubscribe from "../hooks/useTypedSubscribe";
+import Loading from "./Loading";
+import type {
+  PuzzleModalFormHandle,
+  PuzzleModalFormSubmitPayload,
+} from "./PuzzleModalForm";
+import PuzzleModalForm from "./PuzzleModalForm";
+
+/**
+ * Standalone page for adding a puzzle to a hunt, designed for use with
+ * the bookmarklet. Allows selecting the target hunt and prepopulates
+ * the title and URL from query parameters.
+ */
+const AddPuzzlePage = () => {
+  const { t } = useTranslation();
+  const navigate = useNavigate();
+  const [searchParams] = useSearchParams();
+
+  const urlParam = searchParams.get("url") ?? undefined;
+  const titleParam = searchParams.get("title") ?? undefined;
+  const inferredTitle = useMemo(
+    () => inferPuzzleTitle(titleParam, urlParam),
+    [titleParam, urlParam],
+  );
+
+  const [lastSelectedHuntId, setLastSelectedHuntId] =
+    useAddPuzzleLastSelectedHuntId();
+
+  const huntsLoading = useTypedSubscribe(huntsAll);
+  const loadingHunts = huntsLoading();
+
+  const user = useTracker(() => Meteor.user(), []);
+  const writableHunts = useTracker(() => {
+    if (loadingHunts) {
+      return [];
+    }
+    const allHunts = Hunts.find({}, { sort: { createdAt: -1 } }).fetch();
+    return allHunts.filter((h) => userMayWritePuzzlesForHunt(user, h));
+  }, [loadingHunts, user]);
+
+  const selectedHuntId = useMemo(() => {
+    if (
+      lastSelectedHuntId &&
+      writableHunts.some((h) => h._id === lastSelectedHuntId)
+    ) {
+      return lastSelectedHuntId;
+    }
+    return writableHunts[0]?._id;
+  }, [lastSelectedHuntId, writableHunts]);
+
+  const onHuntChange = useCallback(
+    (newHuntId: string) => {
+      setLastSelectedHuntId(newHuntId);
+    },
+    [setLastSelectedHuntId],
+  );
+
+  const tagsLoading = useTypedSubscribe(tagsForHunt, {
+    huntId: selectedHuntId ?? "",
+  });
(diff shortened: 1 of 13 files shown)


imports/client/components/OwnProfilePage.tsx
```diff
@@ -380,6 +382,71 @@ const APIKeysSection = ({ apiKeys }: { apiKeys?: APIKeyType[] }) => {
   );
 };
 
+const BookmarkletSection = () => {
+  const { t } = useTranslation();
+  const origin = window.location.origin;
```
ebroder: Nit: this almost certainly works fine but I think technically wants to be `Meteor.absoluteUrl("/hunts/addpuzzle")` or whatever

imports/client/components/OwnProfilePage.tsx
```diff
@@ -380,6 +382,71 @@ const APIKeysSection = ({ apiKeys }: { apiKeys?: APIKeyType[] }) => {
...
+      <p>
+        {t(
+          "profile.bookmarklet.help",
+          "Drag this button to your bookmarks bar. When viewing a puzzle on an external site, click the bookmarklet to quickly add it to Jolly Roger with the title and URL prepopulated.",
+        )}
+      </p>
+      <div>
+        {/* The href is set via ref to bypass React's javascript: URL blocking.
+            We keep href="#" here so the <a> is valid for a11y linting. */}
+        {/* oxlint-disable-next-line jsx-a11y/anchor-is-valid -- must be an <a> with href for bookmarklet drag-to-bookmarks-bar; onClick only prevents in-page navigation */}
```
ebroder, line 430: Does dropping the `href` cause other linting to fail beyond this one?

imports/client/components/OwnProfilePage.tsx
```diff
@@ -380,6 +382,71 @@ const APIKeysSection = ({ apiKeys }: { apiKeys?: APIKeyType[] }) => {
...
+  const bookmarkletHref = useMemo(() => {
+    const code = `
+      (function () {
+        const title = encodeURIComponent(document.title);
+        const url = encodeURIComponent(window.location.href);
+        const target = "${origin}/hunts/addpuzzle?url=" + url + "&title=" + title;
+        window.open(target, "jr_addpuzzle", "width=550,height=700,scrollbars=yes");
+      })();
+    `;
+    return `javascript:${code.replaceAll(/\s+/g, " ").trim()}`;
```
ebroder: I think this can just be `encodeURIComponent(code)`? (And would be a little less fragile too)

imports/client/components/OwnProfilePage.tsx
```diff
@@ -380,6 +382,71 @@ const APIKeysSection = ({ apiKeys }: { apiKeys?: APIKeyType[] }) => {
...
+      <div>
+        {/* The href is set via ref to bypass React's javascript: URL blocking.
+            We keep href="#" here so the <a> is valid for a11y linting. */}
+        {/* oxlint-disable-next-line jsx-a11y/anchor-is-valid -- must be an <a> with href for bookmarklet drag-to-bookmarks-bar; onClick only prevents in-page navigation */}
+        <a
+          ref={setLinkRef}
+          href="#"
+          className="btn btn-outline-primary"
+          onClick={preventClick}
+          onKeyDown={preventClick}
```
ebroder: Small note: this will also sink the tab key

imports/client/components/AddPuzzlePage.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const loadingTags = Boolean(selectedHuntId && tagsLoading());
+
+  const huntTags = useTracker(() => {
+    if (!selectedHuntId || loadingTags) {
+      return [];
+    }
+    return Tags.find({ hunt: selectedHuntId }).fetch();
+  }, [selectedHuntId, loadingTags]);
+
+  const addModalRef = useRef<PuzzleModalFormHandle>(null);
```
ebroder: I don't think this actually gets used?

imports/lib/inferPuzzleTitle.ts
```diff
@@ -0,0 +1,47 @@
...
+    return trimmedTitle.slice(trimmedTitle.indexOf(" | ") + 3).trim();
+  }
+
+  if (
+    hostname.includes("puzzlehunt.azurewebsites.net") &&
+    trimmedTitle.includes(" - ")
+  ) {
+    // Microsoft Puzzle Server: e.g. "Puzzle Title - Microsoft Puzzle Server" -> "Puzzle Title"
+    return trimmedTitle.slice(0, trimmedTitle.lastIndexOf(" - ")).trim();
+  }
```
ebroder, line 46: I was curious if it made sense to try and generalize the trimming logic so that we could handle (e.g.) MITMH servers that had similar dash- or colon- or pipe-separated `<title>` tags. Turns out
we don't really need to bother:

| Year | Hunt / running team | Puzzle URL shape | `<title>` format |
|---|---|---|---|
| 2017 | Monsters et Manus (Setec Astronomy) | `/2017/puzzle/<slug>.html` | `<Puzzle Name>` |
| 2018 | Operation: Head Hunters (Death & Mayhem) | `/2018/full/puzzle/<slug>.html` | `<Puzzle Name>` |
| 2019 | Holiday Forest (Setec Astronomy) | `/2019/puzzle/<slug>.html` | `<Puzzle Name>` |
| 2020 | Penny Park (Left Out) | `/2020/puzzle/<slug>` | `<Puzzle Name>` |
| 2021 | Yew Labs / ⊥IW (Galactic Trendsetters) | `/2021/puzzle/<slug>/` | `<Puzzle Name> — MIT Mystery Hunt 2021` (em dash) |
| 2022 | Bookspace (Palindrome) | `/2022/puzzle/<slug>/` | `<Puzzle Name> \| MIT Mystery Hunt 2022` (pipe) |
| 2023 | Museum of Interesting Things (teammate) | `/2023/interestingthings.museum/puzzles/<slug>` | `<Puzzle Name>` |
| 2024 | Mythstory Hunt (TTBNL) | `/2024/mythstoryhunt.world/puzzles/<slug>` | `<Puzzle Name>` |
| 2025 | The Case of the Shadow Diamond (Death & Mayhem) | `/2025/hunt/puzzles/<slug>` | `<Puzzle Name>` |
| 2026 | Puzzmon (Cardinality) | `/2026/hunt/puzzles/<slug>` | `<Puzzle Name>` |

imports/lib/inferPuzzleTitle.ts
```diff
@@ -0,0 +1,47 @@
...
+    parsedUrl = new URL(pageUrl);
+  } catch {
+    return trimmedTitle;
+  }
+
+  const hostname = parsedUrl.hostname.toLowerCase();
+  const pathname = parsedUrl.pathname.toLowerCase();
+
+  if (
+    hostname.includes("pandamagazine.com") &&
```
ebroder: Teeeechnically, I think this should be `hostname.endsWith`? (Same with below)

imports/client/components/PuzzleModalForm.tsx
```diff
@@ -242,7 +272,8 @@ const PuzzleModalForm = ({
           setConsiderCompletedWithNoAnswerDirty(false);
           setConfirmingDuplicateUrl(false);
           setAllowDuplicateUrls(false);
-          callback();
+          addPuzzleHuntRecentTags(tags);
```
ebroder: Because this gets reused for both add and edit, I think that editing a puzzle will cause those tags to get pushed into the recent tags list. Arguably harmless but I think undercuts the goal of the feature a bit

imports/client/components/AddPuzzlePage.tsx
```diff
@@ -0,0 +1,184 @@
...
+            "You do not have permission to add puzzles to any hunts. Please join a hunt or contact an operator.",
+          )}
+        </Alert>
+      </Container>
+    );
+  }
+
+  if (!selectedHuntId || loadingTags) {
+    return <Loading />;
+  }
```
ebroder, line 164: This feels like an extremely Claude finding, but it points out that changing the selected hunt will cause the form to get unmounted while `loadingTags`, which means that any user edits to the form will get dropped and replaced.

It proposes pushing loading state down to specifically the tag selection field, which doesn't feel especially practical, and I'm not sure I have an alternative, so this may just be one to live with.

imports/client/components/PuzzleModalForm.tsx
```diff
@@ -74,8 +71,30 @@ const PuzzleModalForm = ({
     callback: (error?: Error) => void,
   ) => void;
   showOnMount?: boolean;
+  /** When true, renders as a full-page form instead of inside a modal dialog. */
+  inline?: boolean;
+  initialTitle?: string;
+  initialUrl?: string;
+  hunts?: HuntType[];
+  onHuntChange?: (huntId: string) => void;
```
ebroder: This maybe conflicts with my complaint about the form unmounting/remounting, but should the hunt form fields push up to `AddPuzzlePage` instead of living here? It feels a little awkward to blow up the interface like this.

ebroder (commented):
OK. Overall, I do prefer this approach over installing an extension. I didn't have time to finish reviewing `PuzzleModalForm` (which I realize is a big chunk of the PR, sorry), but figured I'd go ahead and fire off some low-level thoughts from reviewing the rest as well as a few high-level thoughts:

- Putting the bookmarklet on the profile page (near the bottom) feels pretty undiscoverable. Maybe that's fine - this is a power user feature and operators are power users and maybe we just treat this as a training problem, or something we solve in our Bright New Tailwind Future (🙄 😬 ) - but I'm curious if there's somewhere else where people will find it without it getting in the way on an ongoing basis. Maybe we track if the user has ever used it and until they do we show it on the default operator view?
- The overlap between the `initialTitle` / `initialUrl` params and the `puzzle` param feels kind of messy. I think my best proposal is to factor into a `defaults?: {title?: string, url?: string [...]}` + a `mode: "add" | "edit"`, but consider this preliminary feedback.

This weekend is going to be busy but I'll try to pick up and finish out the review some time next week.

imports/client/components/OwnProfilePage.tsx
jpd236: jsx-a11y/anchor-is-valid still fails (because it has both the `href` and `onClick`), and jsx-a11y/no-static-element-interactions also fails because having event handlers (`onClick` and `onKeyDown`) requires an ARIA `role`, which I suppose we could add. Let me know if you prefer this a different way.

imports/client/components/OwnProfilePage.tsx
jpd236: I think this can be removed without adverse effect, since "Enter" gets suppressed by the onClick handler already.

imports/client/components/AddPuzzlePage.tsx
jpd236: Yes, not in this case. Made the parameter optional in PuzzleModalForm and removed it here.

In an earlier iteration it was popping up as a modal inside the popup for more consistency with the other add pages, but I think the full-page view looks better enough to justify the difference.

imports/lib/inferPuzzleTitle.ts
jpd236: Yeah, it seems common enough to just use the puzzle name that my preference was to keep it simple and run the risk of having to make a day-of fix for a novel format vs. trying to handle things on unknown domains and risk having an overzealous trim. Particularly for dashes - it seems plausible that a handful of puzzles might have legitimate ones in a title.

At least the nice thing about the bookmarklet vs. the extension (at least as I had it) is that we can update this list server-side without needing the bookmarklet JS to change.

imports/lib/inferPuzzleTitle.ts
jpd236: adjusted to `hostname === "[hostname]" || hostname.endsWith(".[hostname]")` to cover junk at the beginning/end.

imports/client/components/PuzzleModalForm.tsx
jpd236: Good point! Adjusted to only invoke this for new puzzles.

imports/client/components/AddPuzzlePage.tsx
jpd236: To pit AI vs. AI, Gemini proposed just removing the "loadingTags" check. Since the tags are only used for autocomplete suggestions, the form still works; you might just not have tag autocomplete working for a bit while the suggestions are loaded.

Alternatively, it suggested using `isLoading` on the `<Creatable>` which shows a loading spinner inside the tag input box.

LMK if you prefer either approach. I agree it's unlikely to matter much in practice since we don't expect the hunt to be changed at all once a hunt begins, and even when you do change it it's unlikely you'd need to do it after entering things in other fields.

imports/client/components/PuzzleModalForm.tsx
jpd236: I agree that's a bit cleaner - done.

I don't think it interferes with the suggestion about unmounting/remounting; the approaches proposed there seem like they'd work even with this new organization. I did a quick test of the proposal to push the loading state to the `<Creatable>` and it seemed to work as expected. (But I left that out here until you confirm how you'd like it to work).

imports/client/components/AddPuzzlePage.tsx
jpd236: I guess one minor concern would be that if you've picked tags under one hunt, should those selections carry over to another hunt? It works, in that it will just create new tags if they don't already exist. But maybe the pre-selected tags from one hunt aren't present in another, and you would make different selections if the right hunt had been selected?

I'm fine either way.

jpd236:
No rush to take a closer look - thanks for the first pass!

Re: discoverability, personally I'd lean towards "advanced feature, tell the people who need to know and don't clutter things otherwise". I'm not opposed to making it more prominent for operators as long as we make whatever the reminder is dismissable in case someone doesn't want it at all. But either way, I think we could land this as incremental progress, and then add another entry point later if we want?

Ack re: the params feedback - will hold off on that for now.

